PR #521, opened 2014-07-22
Add a link to the instructor dash in the course staff debug info

wedaly:
In anticipation of using the analytics infrastructure to put ORA2 data CSV files in the instructor dash, this PR adds a link to the instructor dash to the ORA2 staff debug information.  This will allow instructors to more easily find where to download this information.

@cpennington Is it safe to use course IDs in the instructor dash URL like this?
```
courses/{{ COURSE ID }}/instructor#view-analytics
```
Right now, I'm getting the course ID from the XBlock scope IDs and using `to_deprecated_string()`, but this could change in the future depending on what we need to do to support split mongo.

@stephensanchez Please review.


cpennington:
very belated comment, but yes, it's safe to use `unicode(course_key)` in a url (although url-escaping is probably advised)

mulby:
Note that this has been moved to T&L, see this ticket:

https://openedx.atlassian.net/browse/TNL-1402

PR #640, opened 2014-11-27
Allow file upload backends other than S3

regisb:
**Context**: We do not use Amazon S3 for file storage, but we would like to support file uploads in open assessment. For instance, at [France Université Numérique](https://www.france-universite-numerique-mooc.fr/) we store static files in GlusterFS.

**Proposed solution**: Depending on the value of a setting, uploaded files are sent to one of two backends (s3 or local filesystem). Both backends expose the same API for retrieving upload and download urls. Files uploaded to the local filesystem are stored in:

    /root/bucket/prefix/filename.jpg/
        content
        metadata.json

where 'root', 'bucket' and 'prefix' are user-defined settings. Expiry of upload/download urls is handled via django cache keys.

**Note for the future**: in this pull request we decided to design an ora2-specific solution for file upload. We udnerstand that file upload is an important topic across the Edx codebase. In the future, we might want to extract this feature and place it in an external app to be used in openedx.


regisb:
By "green light" I meant an "OK" on the general ideas behind my proposed implementation :) The rest is work on improving documentation, variable names, code layout, etc. I'll post a new message on the original mailing list thread to notify the original commentators and see if they are interested in reviewing this PR.

singingwolfboy:
@regisb Nobody is objecting, and a few people have responded on the mailing list fairly positively. I think that you can consider that to be a green light. :)

regisb:
Awesome! So in my opinion this PR is ready to be merged.

sarina:
@regisb it will need to go through code review!!! We don't just merge things :) but, it is ready for review, then?

sarina:
@regisb if it is ready for review, please update the description (and title, if necessary) to give as much context to reviewers (and also to remove text such as "wip" or "do not merge")

regisb:
@sarina Right! Description was updated and PR is thus ready for review.

regisb:
Do we have any visibility on an upcoming review date for this PR?

sarina:
@regisb : the review ticket is in the backlog for the ORA2 team to review. [On the ticket](https://openedx.atlassian.net/browse/OSPR-247), I have asked the product owner to estimate when the team will get a chance to look at it. Please understand that our teams work in 2 week sprints, and have competing priorities between a lot of pull requests and other feature work they must attend to, so it can take a bit of time. Hopefully the product owner will respond with an estimate shortly.

For your reference, there is a conflict in your branch. You can choose to rebase on master now, or when the team picks up review - either approach is fine. Just giving you the heads up it will need to be done eventually.

regisb:
No worries :) I just rebased this PR.

doctoryes:
@regisb  I'm diving in to review the design and implementation this week.

@jimabramson  Teaching & Learning is about to do user profile picture uploads, correct? Maybe someone on your time would like to review this PR as well?

openassessment/fileupload/backends/base.py
```diff
@@ -0,0 +1,119 @@
...
+
+class Settings(object):
+    """Store settings related to file upload
+
+    The following settings are used:
+
+        FILE_UPLOAD_STORAGE_BUCKET_NAME (str, required): name of the bucket
+        (AWS or local file directory) to which content will be uploaded.
+
+        FILE_STORAGE_SETTING (str, defaults to "submissions_attachment"): this
```
doctoryes: docstring doesn't match param below.

openassessment/fileupload/backends/base.py
```diff
@@ -0,0 +1,119 @@
...
+    @classmethod
+    def get_bucket_name(cls):
+        bucket_name = getattr(settings, "FILE_UPLOAD_STORAGE_BUCKET_NAME", None)
+        if not bucket_name:
+            raise FileUploadInternalError("No bucket name configured for FileUpload Service.")
+        return bucket_name
+
+    @classmethod
+    def get_prefix(cls):
+        return getattr(settings, "FILE_UPLOAD_STORAGE_PREFIX", "submissions_attachments")
```
doctoryes: Nit: Perhaps this default value would be best pulled out into a separate variable with a comment.

openassessment/fileupload/backends/base.py
```diff
@@ -0,0 +1,119 @@
...
+            A URL (str) to use for a one-time upload.
+
+        Raises:
+            FileUploadInternalError: Raised when an internal error occurs while
+                retrieving a one-time URL.
+            FileUploadRequestError: Raised when the request failed due to
+                request restrictions
+
+        """
+        raise NotImplementedError
```
doctoryes: Nit: Consider using the Python `abc` module to make these abstract methods, since every resulting Backend deriving from this class needs to implement them. It'd make the error happen upon instantiation instead of on calling the method.

sarina:
@doctoryes I'm not sure how profile picture upload in edx-platform impacts adding a different backend to the ORA2 repo

doctoryes:
@sarina The ultimate goal is to arrive at a separate, multiple-backend file storage service and API for edx-platform that can be used by ORA2 for assessment and Studio/LMS for course assets, profile pictures, and other binary files. I looped in T&L to look at this PR since they'll be developing a similar service for Studio/LMS - if these implementations are more unified from the start, it'll be easier to unify all of them later. No direct impact - just keeping the communication lines open.

openassessment/fileupload/views.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    dir_path = os.path.abspath(os.path.dirname(path))
+    if not dir_path.startswith(get_bucket_path()):
+        raise exceptions.FileUploadRequestError("Uploaded file name not allowed: '%s'" % path)
+    root_directory = get_root_directory_path()
+    if not os.path.exists(root_directory):
+        raise exceptions.FileUploadInternalError("File upload root directory does not exist: %s" % root_directory)
+    if not os.path.exists(dir_path):
+        os.makedirs(dir_path)
+    with open(path, 'w') as f:
```
doctoryes: If the file already exists, is it the intended behavior to over-write the previous file?

openassessment/fileupload/views.py
```diff
@@ -0,0 +1,151 @@
...
+        content (str): uploaded file content
+        metadata (dict): json-dumpable data
+    """
+    file_path = get_file_path(key)
+    metadata_path = get_metadata_path(key)
+    if metadata is None:
+        metadata = {}
+
+    safe_save(file_path, content)
+    safe_save(metadata_path, json.dumps(metadata))
```
doctoryes: If the first call succeeds and the second call fails, would you want to do any cleanup?

openassessment/fileupload/views.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    Uploading and download files to the local filesystem backend.
+    """
+    if isinstance(key, unicode):
+        key = key.encode("utf-8")
+    if request.method == "PUT":
+        if not is_upload_url_available(key):
+            raise Http404()
+        content, metadata = get_content_metadata(request)
+        save_to_file(key, content, metadata)
```
doctoryes: `save_to_file` can throw an exception. Perhaps you want to catch it and return something other than a 500 error?

sarina:
@regisb while you're responding to comments, please also perform a rebase to resolve conflicts

openassessment/fileupload/backends/base.py
regisb: Ok.

openassessment/fileupload/backends/base.py
regisb: Ok.

openassessment/fileupload/backends/base.py
regisb: Ok.

openassessment/fileupload/views.py
regisb: Yes: we assume that key collisions result in file overwriting.

openassessment/fileupload/views.py
regisb: Yes we should. I'm adding a test to test this behaviour.

openassessment/fileupload/views.py
regisb: Not 100% sure, but I don't think so. It seems to me that the calls to download_url and upload_url are in ajax and exception handling is performed on the client side: https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/static/js/src/oa_server.js#L515

regisb:
@sarina: thanks for pushing forward this PR! I updated it after rebasing. Also, I took the liberty to add an extra (tiny) commit to fix the Vagrantfile. Is that ok?

openassessment/fileupload/backends/base.py
```diff
@@ -0,0 +1,130 @@
...
+            raise FileUploadInternalError("No bucket name configured for FileUpload Service.")
+        return bucket_name
+
+    @classmethod
+    def get_prefix(cls):
+        """Return the prefix for stored files.
+
+        Defaults to 'submissions_attachments'.
+        """
+        default_file_upload_storage_prefix = "submissions_attachments"
```
doctoryes: Sorry - I wasn't clear. I meant consider moving this default up into `Settings` class scope as `DEFAULT_FILE_UPLOAD_STORAGE_PREFIX`. Then remove the references to `submissions_attachments` in the docstrings above. That way, other users can more easily change the default storage prefix and the tests can also use this default variable in `Settings`.

openassessment/fileupload/backends/base.py
doctoryes: Awesome, thanks!

openassessment/fileupload/backends/s3.py
```diff
@@ -0,0 +1,63 @@
...
+    def get_upload_url(self, key, content_type):
+        bucket_name, key_name = self._retrieve_parameters(key)
+        try:
+            conn = _connect_to_s3()
+            upload_url = conn.generate_url(
+                expires_in=self.UPLOAD_URL_TIMEOUT,
+                method='PUT',
+                bucket=bucket_name,
+                key=key_name,
+                headers={'Content-Length': '5242880', 'Content-Type': content_type}
```
doctoryes: I realize that this magic number was likely copied from the old implementation. But - can a comment be added as to why it's this value?

openassessment/fileupload/views.py
doctoryes: Ok - thanks for checking.

Vagrantfile
```diff
@@ -53,7 +53,7 @@ ora_mount_dir = "edx-ora2"
 
 if ENV['VAGRANT_MOUNT_BASE']
 
-  ora2_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + ora2_mount_dir
+  ora_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + ora_mount_dir
```
sarina: I assume this change it to match the variable name at L52? It may be better to just make this in a separate PR because it's a separate change, and will go in quickly.

sarina:
I don't think the build failed due to your fault, but restarting builds on Travis won't work because it restarts only the most recent build on the repo which may not be your branch. So perhaps push a new commit or rebase again to get a new build.

sarina:
wait nevermind! the restart did work. so we'll wait and see.

openassessment/fileupload/backends/filesystem.py
```diff
@@ -0,0 +1,108 @@
+
+from .base import BaseBackend
+from .. import exceptions
+
+from django.conf import settings
+import django.core.cache
+from django.core.urlresolvers import reverse
+
+
+class Backend(BaseBackend):
```
doctoryes: Since you added `get_backend()` that code will call to instantiate the correct backend, perhaps you should name this `LocalFileBackend` and the the S3 `Backend` to `S3Backend`.

openassessment/fileupload/views.py
```diff
@@ -0,0 +1,163 @@
+import hashlib
```
doctoryes: A general comment about this file: The views contained therein are supposed to support both backends encapsulated behind the `BaseBackend` interface. However, local filesystem backend code is contained in this view file - and it's unclear which code is relevant to which backend. Ideally the filesystem-specific methods should be moved behind the backend interface. At the very least, it should be clear which methods are relevant to which backend implementation - maybe in a separate file.

openassessment/fileupload/backends/base.py
regisb: Ok, I understand.

openassessment/fileupload/backends/s3.py
regisb: I honestly have no idea where it comes from :-/ The boto documentation is quite sparse on that point: http://boto.readthedocs.org/en/latest/ref/s3.html
I suspect "Content-Length" is an entirely optional header that was set to a value higher than the actual request length.

Vagrantfile
regisb: Alright, I created another PR: https://github.com/edx/edx-ora2/pull/681

openassessment/fileupload/backends/filesystem.py
regisb: IMHO that would be a bit redundant with the module name: "s3.py" and "filesystem.py". In fileupload/backends/__init__.py, the backends are refered to as "s3.Backend" and "filesystem.Backend".

openassessment/fileupload/views.py
regisb: Actually, the views listed here all refer to the filesystem backend. The S3 backend does not require any views. I agree that can create some confusion. I will move the views.py module to views_filesystem.py for clarity.

openassessment/fileupload/backends/s3.py
doctoryes: Unfortunate for us all - but fair enough.

openassessment/fileupload/backends/filesystem.py
doctoryes: OK. My personal preference is for the class to contain the differentiating name - but this is fine.

doctoryes:
:+1:  - contingent on a clean test run.

sarina:
@regisb looks like you have a test failure: `FAIL: test_image_and_text_submission (openassessment.xblock.test.test_leaderboard.TestLeaderboardRender)` - can you please address?

regisb:
@sarina: I'm afraid it's a flaky test :-/ I added a commit to help debug it, but it was not supposed to fix it.

sarina:
@symbolist - how does your team deal with a flaky test in ORA? Are you aware of this one (openassessment.xblock.test.test_leaderboard.TestLeaderboardRender)? Should @doctoryes merge this PR with or without the last commit here?

regisb:
Awesome, great job everyone!

symbolist:
@sarina Sorry for the late reply! Yes, this is the only flaky test in this repo. For now we just go to travis and restart the build.

sarina:
@mhoeber this has implications for Open edX - I'm not sure if we're doing documentation yet for open installations, but if/when we are, this should be mentioned.

mhoeber:
@sarina  yes we're trying to do more.  we should at latest add this for cyprus.  https://openedx.atlassian.net/browse/DOC-1776

sarina:
@mhoeber yikes - Cyprus is the country, cypress will be the name of the release. Hopefully that won't cause much confusion...

PR #641, opened 2014-12-02
Use student username instead of anonymous user id to get student info.

muzaffaryousaf:
Get student info via username instead of anonymous user id.

This depends on: https://github.com/edx/edx-platform/pull/6013


muzaffaryousaf:
Open up a https://github.com/edx/edx-ora2/pull/665 instead to resolve the code conflicts.

PR #649, opened 2014-12-18
Limit OpenAssessment BaseView scroll to current element. TNL-499

La0:
This change uses only children of current view when searching for a target to scroll on top.
This is useful when having several Open Assessment on the same pages (and thus with the same html ids).

https://openedx.atlassian.net/browse/TNL-499


La0:
Thanks for the udpate.
Is there any specific documentation about acceptance tests regarding the ora2 project ?

waheedahmed:
The best and easy way to write an acceptance test for given scenario is ```bok-choy``` test in edx-platform here ```edx-platform/common/test/acceptance/tests/```
For help see the documentation here http://bok-choy.readthedocs.org/en/latest/tutorial.html and example tests in given path.

mtyaka:
While your approach to limit the search to the current subtree looks correct, having multiple DOM elements with the same id in a single document is never a good idea. Would it make sense to replace the id with a class, and do a subtree search based on the class instead of the id?

La0:
@mtyaka Sure it would work. The element object in the JS scope is the full assessment, so every reference to ids can be replaced by references to classes.
Should i do this in this PR or make another one ?

mtyaka:
@La0 You can do it in this PR.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -38,7 +38,7 @@ OpenAssessment.BaseView.prototype = {
      */
     scrollToTop: function() {
         if ($.scrollTo instanceof Function) {
-            $(window).scrollTo($("#openassessment__steps"), 800, {offset:-50});
+            $(window).scrollTo($(this.element).find("#openassessment__steps"), 800, {offset:-50});
```
e-kolpakov: @La0 minor nit: jQuery allows specifying a [`context` parameter](http://api.jquery.com/jQuery/) to `$`, so this is equivalent to:

    $(window).scrollTo($("#openassessment__steps", this.element), 800, {offset:-50});

which is a bit cleaner.

test/acceptance/tests.py
```diff
@@ -169,6 +172,9 @@ def test_student_training(self):
 
             self.student_training_page.wait_for_page().wait_for_response().assess(options_selected)
 
+            # Check browser scrolled back to top only on first example
+            self.assertEqual(self.self_asmnt_page.is_on_top, example_num == 0)
```
e-kolpakov: @La0 this approach happen to work fine and test the fix, but it would be good to actually verify that correct "Learn to Assess Responses" block is shown. It's a nice to have though, so if it's too onvolving to implement - don't waste too much time on it.

e-kolpakov:
@La0 I've added a couple of notes as well, and PR needs rebasing.

La0:
I rebased my work and used @e-kolpakov recommendation about jQuery context.
But i won't be able to do some meaningful work until next week (enhancing the test & using some class instead of ids).

sarina:
@waheedahmed please review this pull request

waheedahmed:
:+1: lgtm.

sarina:
@waheedahmed this is an open source pull request, so please merge it or suggest someone else to review.

waheedahmed:
@symbolist Can you review this please?

sarina:
@La0 can you please rebase this pull request?

La0:
@sarina I just rebased it.

sarina:
@symbolist @waheedahmed can you please rerun the build on this branch? I don't know how to do it. The last build on this branch passed, before it was rebased.

waheedahmed:
@La0 Please squash your commits then i will rerun the build.

La0:
@waheedahmed I just squased them, the tests are running.

sarina:
@waheedahmed ready to be re-reviewed.

PR #651, opened 2014-12-23
Multiple prompts 1: Block initialization, xml import/export. 

symbolist:
https://openedx.atlassian.net/browse/TNL-708


openassessment/xblock/data_conversion.py
```diff
@@ -56,13 +57,42 @@ def convert_training_examples_list_to_dict(examples_list):
...
+
+    Returns:
+        list of dict
+    """
+
+    if prompt_or_serialized_prompts is None:
+        prompt_or_serialized_prompts = ''
+
+    try:
+        prompts = json.loads(prompt_or_serialized_prompts)
```
stephensanchez: Is this to say, if we get a single prompt instead of a list of prompts, json.loads will throw an exception, and we'll then assume it is one prompt?

I think I'd prefer to determine if we have one prompt, or a set of serialized prompts, before we hit this line, and if there isa  ValueError, it would only be a legitimate one. 

If there is a single prompt, the serialized prompt list given to this function would have the one element.

openassessment/xblock/openassessmentblock.py
```diff
@@ -422,6 +423,40 @@ def _(self):
...
+        parse and return it. Otherwise, assume it is a simple string prompt
+        and return it in a list of dict.
+
+        Returns:
+            list of dict
+        """
+        return create_prompts_list(self.prompt)
+
+    @prompts.setter
+    def prompts(self, value):
```
stephensanchez: I'm a little confused about what's being done here. Are we trying to dynamically update "prompt" when "prompts" is set? If so, I imagined "prompt" being a deprecated value, that we could read when "prompts" is not set, but that nothing would actually use "prompt" any longer.

openassessment/xblock/test/data/parse_rubric_xml.json
```diff
@@ -2,7 +2,6 @@
     "simple": {
         "xml": [
             "<rubric>",
-            "<prompt>Test prompt</prompt>",
```
stephensanchez: isn't this the rubric prompt?

openassessment/xblock/data_conversion.py
stephensanchez: Oh I see what's happening now. All the configuration is still in "prompt", but may be a serialized representation of multiple prompts.

Out of curiosity, why did you choose this approach versus a new "prompts" field on the XBlock, then only use "prompt" for deprecated blocks? I'd imagine it would save you some of the complexity around serialization.

openassessment/xblock/openassessmentblock.py
stephensanchez: Disregard this if you keep the single "prompt" field, this was before I realized how "prompts" was being evaluated.

openassessment/xblock/test/test_xml.py
```diff
@@ -317,6 +345,33 @@ def _dict_mutations(self, input_dict):
...
+        3) Replace list with unicode
+        4) Replace list with an integer
+
+        Args:
+            input_list (list): A JSON-serializable list to traverse.
+
+        Yields:
+            list
+        """
+        print "== Emptying list"
```
stephensanchez: Left over print statement?

openassessment/xblock/xml.py
```diff
@@ -368,13 +428,6 @@ def parse_rubric_xml(rubric_root):
     """
     rubric_dict = dict()
 
-    # Rubric prompt
-    prompt_el = rubric_root.find('prompt')
```
stephensanchez: Confused here. The rubric has multiple prompts too?

openassessment/xblock/openassessmentblock.py
symbolist: Actually we discussed the two approaches when I was starting work on this and you recommended this one. :)

We can go with the other one too, but this does look simpler in retrospect. If we do have two fields, we will have to add a property anyway which will first check `prompts` and if that is empty will check `prompt`. This way we have one less entity. What do you think?

stephensanchez: Ha! I would have done that. I'll leave it up to you. Looking at the implementation I could see advantages to cleaning it up by having two fields, but also having that deprecated field is a little unfortunate.

openassessment/xblock/test/data/parse_rubric_xml.json
symbolist: Actually in the xml the `prompt` is currently serialized inside the `rubric` element. I have moved it out to the top level. For example, here is a subset of the xml for the ORA problem in the demo course:
```
  <openassessment url_name="b24c33ea35954c7889e1d2944d3fe397" submission_start="2001-01-01T00:00:00+00:00" submission_due="2029-01-01T00:00:00+00:00" allow_file_upload="False" allow_latex="False">
    <title></title>
    <assessments>
      <assessment name="student-training"></assessment>
      <assessment name="peer-assessment" must_grade="1" must_be_graded_by="1" start="2001-01-01T00:00:00+00:00" due="2029-01-01T00:00:00+00:00"/>
      <assessment name="self-assessment" start="2001-01-01T00:00:00+00:00" due="2029-01-01T00:00:00+00:00"/>
    </assessments>
    <rubric>
      <prompt>    Describe your favorite meal, and write a persuasive argument to convince someone to try that meal. Support your position with facts and examples from your experience. 
</prompt>
      <criterion feedback="optional">
        <name>Ideas</name>
        <label>Ideas</label>
        <prompt>Determine if there is a unifying theme or main idea.</prompt>
      </criterion>
      <feedbackprompt>How does this meal compare with the meal you submitted?</feedbackprompt>
      <feedback_default_text>I noticed that this response...
</feedback_default_text>
    </rubric>
  </openassessment>
```

stephensanchez: Ah! That makes sense then. That clears up a bit of my confusion. Thanks.

openassessment/xblock/xml.py
symbolist: To add to what I wrote above, there is no thing as the rubric prompt. But the main text that is set onto the `prompt` field was being serialized inside the `rubric` element (which is why there is parsing logic here). I have moved it out to the root of the xml and this logic to the `_parse_prompts_xml ` function.

stephensanchez:
:+1:  Thanks for answering my questions. This part looks good.

PR #654, opened 2014-12-24
Multiple prompts 3: LMS

symbolist:


openassessment/xblock/test/data/submission_unavailable.xml
```diff
@@ -1,9 +1,13 @@
...
-    <prompt>
-        Given the state of the world today, what do you think should be done to
-        combat poverty? Please answer in a short essay of 200-300 words.
-    </prompt>
+    <prompts>
+        <prompt>
+            <description>Given the state of the world today, what do you think should be done to combat poverty?</description>
+        </prompt>
+        <prompt>
+            <description>Given the state of the world today, what do you think should be done to combat pollution?</description>
```
stephensanchez: Is it worth having different prompts and testing that the order is properly preserved?

stephensanchez:
:+1:  this part looks good.

stephensanchez:
While reading the next PR I thought of a few things regarding this PR:
* Doesn't student training need to be updated to support multiple parts as well (or did I miss this?)
* AI Grading, what is our story around that? Will EASE need to evaluate X parts of a submission?

openassessment/xblock/test/data/submission_unavailable.xml
symbolist: Actually, these two prompts are different; the first asks about poverty and the second about pollution. Sorry, couldn't think of an original question. ;-)

symbolist:
1. Yes, I mentioned in the email too that I'll be creating a separate PR for student training (studio and lms).
2. AI assessment does not need to work for problems with multiple prompts. But since there is one assessment for all the prompts anyway, I am going to change the AI step to concatenate the answers and assess that string, and so it will work for 1 or more prompts.
3. Oh, I missed that. Thanks for pointing it out.

So right now, I need to update the AI step, the training step and the leaderboard logic. I'll create a separate PR for these.

symbolist:
It looks like I missed updating a few tests here. Will look into that too.

stephensanchez:
Great, sounds like you have things covered!

openassessment/templates/openassessmentblock/oa_submission_answer.html
```diff
@@ -0,0 +1,17 @@
+{% spaceless %}
+{% load i18n %}
+<ol class="submission__answer__display__content">
+    {% for part in answer.parts %}
+        <li class="submission__answer__part">
+            <article class="submission__answer__part__prompt">
+                {{ part.prompt.description|linebreaks }}
```
frrrances: Add in a context header here (like line 58 in oa_response.html) for screenreaders.

openassessment/templates/openassessmentblock/oa_submission_answer.html
```diff
@@ -0,0 +1,17 @@
...
+            </article>
+            <div class="submission__answer__part__text">
+                <div class="submission__answer__part__text__value">
+                    {{ part.text|linebreaks }}
```
frrrances: Add in a context header here (like line 58 in oa_response.html), but the header should read "Peer Response:". I think it would help to make this header visible on the page as well to lead reviewers to the part(s) they should be assessing.

frrrances:
Overall, the code looks good. I made a couple of notes inline about retaining the headers to give some sense of what the content groups are on the page. 

For the UX side of things, I think it is a little unclear what all the boxes are on the page, for both the response and review screens. And for the visual styling on the review step, I think the weighting of what part of the answer to pay attention to is unclear... everything seems to be the same weight. I sketched a few ideas that could help with these. 

* Response step could have placeholder text (top example) or intro text (bottom example):

ora-reponse1and2


* Review step could use intro text: 

ora-review1

or reduce the focus on the prompt:

ora-review2

None of these are blockers, but I think they will reduce confusion if you are able to address them. :+1:

PR #658, opened 2014-12-31
Remove the submission from grading pool from staff debug panel.

muzaffaryousaf:
Remove the submission from grading pool.
https://openedx.atlassian.net/browse/TNL-900


muzaffaryousaf:
@explorerleslie @stephensanchez This PR includes the updated code for https://github.com/edx/edx-ora2/pull/658. I've implemented the new approach that we discussed. @stephensanchez Can you please review the code and let me know what do you think. 
Thanks.!

stephensanchez:
@muzaffaryousaf I'll check this out tomorrow morning, sorry for the delay.

openassessment/templates/openassessmentblock/peer/oa_peer_cancelled.html
```diff
@@ -0,0 +1,23 @@
...
+<li id="openassessment__peer-assessment" class="openassessment__steps__step step--peer-assessment">
+{% endblock %}
+
+{% block title %}
+<span class="step__status">
+  <span class="step__status__label">{% trans "This step's status" %}:</span>
+  <span class="step__status__value">
+    <i class="ico icon-warning-sign"></i>
+    <span class="copy">
+        {% trans "Cancelled" %}
```
stephensanchez: Is there someone from Doc to review the messaging for these new templates?

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
```diff
@@ -0,0 +1,52 @@
...
+    <div class="ui-toggle-visibility__content">
+        <div class="wrapper--step__content">
+
+            <div class="step__message message message--incomplete">
+                <h3 class="message__title">{% trans "Submission Cancelled" %}</h3>
+
+                <div class="message__content">
+                  <p>
+                    {% blocktrans with removed_datetime=workflow_cancellation.created_at|utc|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
+                        Your submission has been canceled by {{ removed_by_username }} on {{ removed_datetime }}
```
stephensanchez: Apparently "Cancelled" and "Canceled" are both appropriate spellings, go figure. But we should be consistent with one spelling.

Also, I think we may want to avoid having the "removed_by_username" in the message.

openassessment/xblock/test/test_submission.py
```diff
@@ -199,6 +199,29 @@ def test_open_submitted(self, xblock):
...
+        self._assert_path_and_context(
+            xblock, 'openassessmentblock/response/oa_response_cancelled.html',
+            {
+                'submission_due': dt.datetime(2999, 5, 6).replace(tzinfo=pytz.utc),
+                'student_submission': submission,
+                'allow_file_upload': False,
+                'has_peer': True,
+                'has_self': True,
+                'allow_latex': False,
+                'workflow_cancellation': None
```
stephensanchez: I'm confused, shouldn't this have a workflow_cancellation?

openassessment/xblock/staff_info_mixin.py
```diff
@@ -204,9 +214,13 @@ def render_student_info(self, data, suffix=''): # pylint: disable=W0613
...
-        student_id = data.params.get('student_id', '')
-        path, context = self.get_student_info_path_and_context(student_id)
-        return self.render_assessment(path, context)
+        try:
+            student_id = data.params.get('student_id', '')
+            path, context = self.get_student_info_path_and_context(student_id)
+            return self.render_assessment(path, context)
+
+        except PeerAssessmentInternalError as ex:
+            return self.render_error(self._(ex.message))
```
stephensanchez: Unfortunately, I don't think this will work. ugettext will parse the code pre-compiled to find the strings within the function, so you'd need the string in-line here, rather than ex.message.

If we're using this pattern elsewhere, we should probably fix it. (I may have done this at times before I knew the caveats of ugettext)

stephensanchez:
This is looking really clean, thanks a lot for revisiting this! Most of my comments are nit-picks.

One quick question. There's a "Leaderboard" feature, which shows the top X submissions at the bottom. We may want to make sure cancelled submissions don't show up there. I'm imagining a rude submission is cancelled early on, and then everyone sees it on the leaderboard. This may be worth a separate pull request if it requires additional work.

Other than that, if the comments are addressed, and commits are squashed, I'd say :+1:

openassessment/templates/openassessmentblock/peer/oa_peer_cancelled.html
muzaffaryousaf: @explorerleslie can you please refer someone for messaging review?
Thanks.

openassessment/xblock/test/test_submission.py
muzaffaryousaf: Well i'm checking based on the workflow status which is 'cancelled' and ignoring the further information but i think you are right it could be confusing. i will add the 'workflow_cancellation' information to make it more readable.

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
muzaffaryousaf: @explorerleslie what do you think about showing the username in above message ?

muzaffaryousaf:
@stephensanchez I think we need more work for top X submissions in case of cancelled submission.
We may need to restrict cancelled submission don't shown up there and also need to improve the user experience for cancelled submission.

@explorerleslie can we make other ticket for Top submissions functionality.
Thanks.

openassessment/assessment/migrations/0025_auto__add_field_peerworkflow_cancelled_at.py
```diff
@@ -0,0 +1,172 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
```
stephensanchez: One thought on this migration. There are many rows in the PeerWorkflow table in production. We should wrap this migration operation in a database lock to avoid having the migration fail. 

example https://github.com/edx/edx-ora2/blob/muzaffar/cancel-student-submission-tnl900/openassessment/assessment/migrations/0018_auto__add_field_assessmentpart_criterion.py#L17-17

muzaffaryousaf:
@stephensanchez I've added lock to the migration and did other changes as well. Please let me know if you have anything else that you want me to change/improve. 
Thanks.

explorerleslie:
I'd like the username to stay in the message presented to students. it gives some accountability (we've gotten feedback about the forums that, when you delete a post, it doesn't give any insight into who deleted it).

explorerleslie:
I've created https://openedx.atlassian.net/browse/TNL-1125 for hiding removed submissions from the leaderboard.

muzaffaryousaf:
@explorerleslie Just FYI ! 
This PR is ready except the Documentation.

muzaffaryousaf:
@stephensanchez I've looked the code for showing top X submissions. i was thinking we can hide the submissions that have 0 score because at submission api level that is the only thing we can check. Does this make sense to not showing the submissions with 0 score or is there any other way that i'm missing.
Please let me know what do you think. 
Thanks!

stephensanchez:
That seems like a reasonable assumption to me. Though a score of '0' could
theoretically be the top score early on, I'm sure people would not want to
broadcast that.

PR #662, opened 2015-01-20
Staff assessment.

symbolist:
@stephensanchez Instead of describing the design in text, I thought it would be clearer to do so in code. I have made changes to the `workflow` app and `student_view` to handle the peculiarities of an optional staff assessment. What are your thoughts?

Also included is the staff assessment api (but that is mostly a copy/paste of code from the other apis).

### DIFF
--- a/openassessment/assessment/api/staff.py
+++ b/openassessment/assessment/api/staff.py
@@ -1,0 +1,289 @@
+"""
+Public interface for staff grading, used by students/course staff.
+"""
+import logging
+from django.db import DatabaseError, IntegrityError, transaction
+from dogapi import dog_stats_api
+
+from openassessment.assessment.models import (
+    Assessment, AssessmentFeedback, AssessmentPart,
+    InvalidRubricSelection
+)
+from openassessment.assessment.serializers import (
+    AssessmentFeedbackSerializer, RubricSerializer,
+    full_assessment_dict, rubric_from_dict, serialize_assessments,
+    InvalidRubric
+)
+from openassessment.assessment.errors import (
+    StaffAssessmentRequestError, StaffAssessmentInternalError
+)
+from submissions import api as sub_api
+
+logger = logging.getLogger("openassessment.assessment.api.peer")
+
+
+STAFF_TYPE = "ST"
+
+
+def submitter_is_finished(submission_uuid, requirements):
+    """
+    Determine if the submitter has finished their requirements for staff
+    assessment. Always returns True.
+
+    Args:
+        submission_uuid (str): Not used.
+        requirements (dict): Not used.
+
+    Returns:
+        True
+
+    """
+    return True
+
+
+def assessment_is_finished(submission_uuid, requirements):
+    """
+    Determine if the assessment of the given submission is completed. This
+    checks to see if staff have completed the assessment.
+
+    Args:
+        submission_uuid (str): The UUID of the submission being graded.
+        requirements (dict): Any variables that may effect this state.
+
+    Returns:
+        True if the assessment has been completed for this submission.
+
+    """
+    required = requirements.get('staff', {}).get('required', False)
+    if required:
+        return bool(get_latest_assessment(submission_uuid))
+    return True
+
+
+def get_score(submission_uuid, requirements):
+    """
+    Generate a score based on a completed assessment for the given submission.
+    If no assessment has been completed for this submission, this will return
+    None.
+
+    Args:
+        submission_uuid (str): The UUID for the submission to get a score for.
+        requirements (dict): Not used.
+
+    Returns:
+        A dictionary with the points earned and points possible.
+
+    """
+    assessment = get_latest_assessment(submission_uuid)
+    if not assessment:
+        return None
+
+    return {
+        "points_earned": assessment["points_earned"],
+        "points_possible": assessment["points_possible"]
+    }
+
+
+def get_latest_assessment(submission_uuid):
+    """
+    Retrieve the latest staff assessment for a submission.
+
+    Args:
+        submission_uuid (str): The UUID of the submission being assessed.
+
+    Returns:
+        dict: The serialized assessment model
+        or None if no assessments are available
+
+    Raises:
+        StaffAssessmentInternalError
+
+    Example usage:
+
+    >>> get_latest_assessment('10df7db776686822e501b05f452dc1e4b9141fe5')
+    {
+        'points_earned': 6,
+        'points_possible': 12,
+        'scored_at': datetime.datetime(2014, 1, 29, 17, 14, 52, 649284 tzinfo=<UTC>),
+        'scorer': u"staff",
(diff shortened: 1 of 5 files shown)


openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,289 @@
...
+        requirements (dict): Any variables that may effect this state.
+
+    Returns:
+        True if the assessment has been completed for this submission.
+
+    """
+    required = requirements.get('staff', {}).get('required', False)
+    if required:
+        return bool(get_latest_assessment(submission_uuid))
+    return True
```
symbolist: If staff assessment is not required this step will complete immediately.

openassessment/workflow/models.py
```diff
@@ -239,7 +240,8 @@ def get_score(self, assessment_requirements, step_for_name):
                     else:
                         requirements = assessment_requirements.get(assessment_step_name, {})
                     score = get_score_func(self.submission_uuid, requirements)
-                    break
+                    if score:
+                        break
```
symbolist: `get_score` of existing apis can return `None` if the step is not complete. For staff assessment, `get_score` will also return `None` if the step is optional and staff have not accessed it. By making this change, the score of the next step will be used.

openassessment/workflow/models.py
```diff
@@ -557,6 +564,8 @@ def update(self, submission_uuid, assessment_requirements):
             step_changed = True
 
         # Has the step received a score?
+        # If staff assessment is optional we will mark assessment as complete immediately.
+        # But if staff comes and assesses later, this date will not be updated.
```
symbolist: Is this okay or should the `assessent_completed_at` date be updated if staff gives an assessment after the state is `done`?

openassessment/xblock/openassessmentblock.py
```diff
@@ -329,6 +335,10 @@ def _create_ui_models(self):
         """
         ui_models = [UI_MODELS["submission"]]
         for assessment in self.valid_assessments:
+            if assessment["name"] == "staff-assessment" and assessment["required"] == False:
+                # Check if staff have graded the assessment
+                # else
+                continue
```
symbolist: So if the staff assessment step is optional and staff have not given an assessment, we do not show the step in the UI.

openassessment/workflow/models.py
```diff
@@ -287,6 +287,17 @@ def update_from_assessments(self, assessment_requirements):
...
+        # If the status is done or cancelled, check if score has changed.
+        if self.status == self.STATUS.done:
+            if force_update_score:
+                new_score = self.get_score(assessment_requirements, step_for_name)
+                self.set_score(new_score)
+                self.save()
+                logger.info((
+                    u"Workflow for submission UUID {uuid} has updated score."
+                ).format(uuid=self.submission_uuid))
+            return
```
symbolist: When a staff assessment is given, we pass `force_update_score = True` to this method. If status is not `done` the normal flow continues but if status is `done`, the score is updated.

Is there a better way of updating the score after the workflow is `done`?

stephensanchez:
Sorry for the delay, Usman, I'll try to get to reviewing this in the afternoon!

openassessment/assessment/api/staff.py
stephensanchez: Seems logical.

openassessment/workflow/models.py
stephensanchez: This may not work. If you have a question with Peer + Self, and you complete your Peer section, then Self, then you should be in a state where you are waiting for assessments from your Peer. With this change, I think you may end up just giving them their "Self" score. 

I think you may need to change all the APIs to return a score, as well as whether the score should not be omitted.   This may be worth updating the get_score function to return another attribute in the dictionary, something like optional: True

openassessment/workflow/models.py
stephensanchez: I actually like the idea of a force option to the function. Though I wonder if it could be simplified, here, to just:
if self.status == self.STATUS.done and not force_update_score:
    return

If you did that, then you'd just go through the flow, per usual. That would more truly force the update of the workflow, and work the same way. If we ever wanted to use the 'force_update' override, it could be more repurposed.

openassessment/workflow/models.py
stephensanchez: Seems like a good idea to update the timestamp.

stephensanchez:
How do professors / staff use this? I don't see a way, as a professor, to open this section up and grade student submissions. Also, is the requirement that a staff member could grade the students one by one off a stack, or that they could specify the submission / student they want to grade?

openassessment/workflow/models.py
symbolist: The score is only set when for each workflow step `assessment_completed_at` has been set, which is only set when the corresponding api returns `True` for `assessment_is_finished`, and for peer api that will only happen if the peers have assessed the submission. So as far as I understand it, it should work.

However, I think your suggestion will make the code clearer. So I'll update the `get_score` function to return a tuple [done, score.]

openassessment/workflow/models.py
symbolist: That does seem like a more general solution but it will need a couple of tweaks towards the bottom of the function too (because the score is only set if the current status is `waiting`). Let me look into this.

symbolist:
The staff interface will most likely be a part of the staff info panel. And we will need both options; if staff grading is optional, staff can look up submissions for particular students and if it is mandatory, they can grade off a stack. @explorerleslie is working on a detailed acceptance criteria and UI design for the staff grading part, so will know soon.

openassessment/workflow/models.py
stephensanchez: I don't mean checking to see if a score is available, but whether you should wait for one module to return a score instead of moving onto the next one. 

So you can test the following scenario:
* Create an ORA2 problem with Peer Assessment, Self Assessment
* Go in as a student, submit an essay
* Grade some peers
* Complete your self assessment
At this point, you should be waiting for grades from your peers, and you should not have a score. I believe with your change, you will have a score, based on your self assessment.

stephensanchez:
Got it, so the staff interaction is separate. That works. Another option (and maybe less clunky for UX) is to just steal the "Over Grading" design, and when a staff member loads the ORA2 problem, allow them to start pulling and grading in the staff section. You'll still need to tweak it to allow them to select a specific essay, but might just be cleaner, overall.

openassessment/workflow/models.py
symbolist: Hmm.. I may be missing something but the score is only set if all the steps have `assessment_completed_at` set. And that will only happen for the peer assessment step if the submission has been graded by the peers. But I'll read through the code again in the morning.

https://github.com/edx/edx-ora2/blob/usman/tnl1162-staff-assessment/openassessment/workflow/models.py#L322

PR #665, opened 2015-01-27
Get student info via user service.

muzaffaryousaf:
This PR includes the code to get the Student Info using the username instead of course specific anonymous user id. It would get the anonymous_user_id using the xblock 'user' service. 

https://openedx.atlassian.net/browse/TNL-836


muzaffaryousaf:
@stephensanchez can you please review the code for this PR. 
Thanks.!

openassessment/xblock/openassessmentblock.py
```diff
@@ -200,22 +201,45 @@ class OpenAssessmentBlock(
...
 
         item_id = self._serialize_opaque_key(self.scope_ids.usage_id)
+
         # This is not the real way course_ids should work, but this is a
         # temporary expediency for LMS integration
         if hasattr(self, "xmodule_runtime"):
-            course_id = self._serialize_opaque_key(self.xmodule_runtime.course_id)  # pylint:disable=E1101
-            student_id = self.xmodule_runtime.anonymous_student_id  # pylint:disable=E1101
+            course_id = self.course_id  # pylint:disable=E1101
+            if anonymous_user_id:
```
stephensanchez: I'm unclear on what a anonymous user id is versus a anonymous student id.

openassessment/xblock/openassessmentblock.py
```diff
@@ -200,22 +201,45 @@ class OpenAssessmentBlock(
...
+        """
+        return self.runtime.service(self, 'user').get_anonymous_user_id(username, course_id)
+
+    def get_student_item_dict(self, anonymous_user_id=None):
         """Create a student_item_dict from our surrounding context.
 
         See also: submissions.api for details.
 
+        Args:
+            anonymous_user_id(str):
```
stephensanchez: Please fill this out

openassessment/xblock/staff_info_mixin.py
```diff
@@ -208,58 +208,59 @@ def render_student_info(self, data, suffix=''):  # pylint: disable=W0613
...
-                        submission['image_url'] = file_api.get_download_url(file_key)
-                    except file_api.FileUploadError:
-                        # Log the error, but do not prevent the rest of the student info
-                        # from being displayed.
-                        msg = (
-                            u"Could not retrieve image URL for staff debug page.  "
-                            u"The student ID is '{student_id}', and the file key is {file_key}"
-                        ).format(student_id=student_id, file_key=file_key)
-                        logger.exception(msg)
+        if student_username:
```
stephensanchez: Nitpick, but I think you could avoid some of this indentation. For example, line 242 does not have to be nested, since anonymous_user_id will be None if 238 is not true, and 247 could be brought out of the nested if statement, since submissions will be None if anonymous_user_id is None.

Not a big deal, but may read a little more cleanly.

openassessment/xblock/openassessmentblock.py
stephensanchez: I'm a little worried about what it means to pass in an anonymous user id versus pulling it from the modulestore. What would this fallback look like in production?

muzaffaryousaf: @stephensanchez Previously the anonymous_user_id was being set to "student_id" in **student_item**  in staff_info_mixin.py:L240 anyways. I've just shifted the logic to the **get_student_item_dict** method.

openassessment/xblock/staff_info_mixin.py
muzaffaryousaf: @stephensanchez All done. Please let me know if you have any other suggestions.
Thanks!!

stephensanchez:
:+1:

PR #667, opened 2015-01-27
Multiple prompts 4: Misc 

symbolist:
This PR makes the following changes:
- Staff info panel shows submissions in the new format (with prompts + answers).
- Leaderboard shows submissions in the new format.
- Student training step shows submissions in the new format.
- AI grader concatenates all of the parts of the answers and treats it as a single string.

### DIFF
--- a/openassessment/assessment/models/ai.py
+++ b/openassessment/assessment/models/ai.py
@@ -40,6 +40,35 @@ CLASSIFIERS_CACHE_IN_FILE = getattr(
 )
 
 
+def essay_text_from_submission(submission):
+    """
+    Retrieve the submission text.
+
+    Submissions are arbitrary JSON-blobs, which *should*
+    contain a single key, "answer", containing the essay
+    submission text.
+    If not, though, assume we've been given the essay text
+    directly (convenient for testing).
+    """
+    if isinstance(submission, dict):
+        if 'answer' in submission:
+            # Format used for answer in examples.
+            if isinstance(submission['answer'], unicode):
+                return submission['answer']
+            # Initially there was one prompt and submission had the structure
+            # {'answer': {'text': 'The text.'}}
+            elif 'text' in submission['answer']:
+                essay_text = submission['answer']['text']
+            # When multiple prompts were introduced the structure of submission become:
+            # {'answer': {'parts': [{'text': 'The text part 1.'}, {'text': 'The text part 2.'}]}}
+            # We concatenate these parts and let AI grader evaluate the total text.
+            else:
+                essay_text = u'\n'.join([part['text'] for part in submission['answer']['parts']])
+    else:
+        essay_text = unicode(submission)
+    return essay_text
+
+
 class IncompleteClassifierSet(Exception):
     """
     The classifier set is missing a classifier for a criterion in the rubric.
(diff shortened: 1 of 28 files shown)


openassessment/assessment/models/ai.py
```diff
@@ -40,6 +40,35 @@
...
+    """
+    Retrieve the submission text.
+
+    Submissions are arbitrary JSON-blobs, which *should*
+    contain a single key, "answer", containing the essay
+    submission text.
+    If not, though, assume we've been given the essay text
+    directly (convenient for testing).
+    """
+    if isinstance(submission, dict):
```
stephensanchez: It seems a little odd to have the logic of a dict vs. unicode just for testing convenience. Is this really necessary?

openassessment/xblock/data_conversion.py
```diff
@@ -57,6 +73,27 @@ def convert_training_examples_list_to_dict(examples_list):
...
+
+    Args:
+        assessments (list): list of assessments
+    Returns:
+        list of dict
+    """
+    for assessment in assessments:
+        if 'examples' in assessment:
+            for example in assessment['examples']:
+                if isinstance(example['answer'], unicode) or isinstance(example['answer'], str):
```
stephensanchez: wasn't this file in PR #668?

openassessment/xblock/leaderboard_mixin.py
```diff
@@ -72,13 +75,16 @@ def render_leaderboard_complete(self, student_item_dict):
         for score in scores:
             if 'file_key' in score['content']:
                 score['file'] = file_upload_api.get_download_url(score['content']['file_key'])
-            if 'text' in score['content']:
-                score['content'] = score['content']['text']
+            if 'text' in score['content'] or 'parts' in score['content']:
```
stephensanchez: One comment I'd make, given the need for backwards compatibility, is exploring the idea of a utility module for handling submissions that may or may not be multi-part, etc. Then we won't have to have special case back compat code throughout. 

Just a thought.

stephensanchez:
Generally looks good. Looks like this might have some commits from other PRs as well.  The number of places that know about the structure of a submission and backwards compatibility seems like it may be growing, so my only suggestion there is to see if there's some number of helper functions that could be created to separate that out. 

:+1:

openassessment/assessment/models/ai.py
symbolist: I just moved this code from below to a separate function. I get your point, but this will require updating a lot of test data.

openassessment/xblock/data_conversion.py
symbolist: Yes, this commit was needed in both.

openassessment/xblock/leaderboard_mixin.py
symbolist: Actually, most of the backwards compatibility support is in `data_conversion.py`. It looks like I missed this case.

symbolist:
@stephensanchez Thanks for the review. I agree with your suggestions. However, this story has been in development for too long and I think we should merge it now. But I'll keep this cleanup in mind when we do the work on staff assessment.

stephensanchez:
Thanks fine with me, go ahead with the merge!

PR #668, opened 2015-02-05
Multiple prompts 5: Training step examples editing

symbolist:
- For each example in the studio editor, the number of textareas matches the number of prompts.
- Parsing and serialization support for answers with multiple parts.

Some student training step tests are going to fail in this branch, but they are fixed in the other one.


openassessment/xblock/data_conversion.py
```diff
@@ -57,6 +73,27 @@ def convert_training_examples_list_to_dict(examples_list):
...
+
+    Args:
+        assessments (list): list of assessments
+    Returns:
+        list of dict
+    """
+    for assessment in assessments:
+        if 'examples' in assessment:
+            for example in assessment['examples']:
+                if isinstance(example['answer'], unicode) or isinstance(example['answer'], str):
```
stephensanchez: Just curious why you're doing the isinstance check here. Would there be a scenario where it would not be unicode or str?

stephensanchez:
:+1:

PR #671, opened 2015-02-12
LaTeX not rendering properly in peer assessment

mlkwaqas:
Latex was not working properly in peer assessment
as we were trying to render the Latex and Mathjax
rendering on string instead of element in the DOM.
Mathjax works when we apply it on the element which
is present in the DOM.

TNL-1350


mlkwaqas:
@adampalay @symbolist @zubair-arbi please have a look 
http://mlkwaqas.m.sandbox.edx.org/courses/RiceX/ELEC301x_/2015Q3/courseware/47d57323afb34f2cb559d658dd00f008/9aef0e13342646c1914086b6f49df296/

mlkwaqas:
working on tests failure now

openassessment/xblock/static/js/src/oa_server.js
```diff
@@ -78,11 +77,9 @@ if (typeof OpenAssessment.Server == "undefined" || !OpenAssessment.Server) {
             element: The element to modify.
         **/
         renderLatex: function(element) {
-            $('.allow--latex', element).each(
-                function() {
-                    MathJax.Hub.Queue(['Typeset', MathJax.Hub, this]);
-                }
-            );
+            if(element.hasClass('allow--latex')) {
```
zubair-arbi: Looking at the code I believe that the variable 'element' is not just a single element but an array of elements.
So instead of getting just first element by index ```element[0]``` you can write something like:
```javascript
element.filter(".allow--latex").each(function() {
    MathJax.Hub.Queue(['Typeset', MathJax.Hub, this]);
});
```

zubair-arbi:
LGTM :+1:

mlkwaqas:
@wedaly @stephensanchez please have a look

wedaly:
:+1:

PR #672, opened 2015-02-13
Unnecessary tabpanel to navigate activity steps

ahsan-ul-haq:
unnecessary tabpanel to navigate activity steps was used which was noth currently functioning
properly. It is currently considered reduntant so removed.

TNL-69


ahsan-ul-haq:
@cptvitamin could you please review whether this is how you want it to be?

ahsan-ul-haq:
@symbolist Could you please review it

symbolist:
If this is what is required. :+1:

cptvitamin:
Thank @ahsan-ul-haq.  This is exactly what needed to be done.  This navigation for screen reader users is unnecessary as the page has sufficient navigational aids already and its inclusion was actually confusing for keyboard only users.  Thanks for removing it :+1:   please merge.

PR #677, opened 2015-02-19
Ahsan/tnl 66 add aria expanded attributes

ahsan-ul-haq:
Panels did not include the  aria-expanded attribute to indicate the
current state of the panel

TNL-66


ahsan-ul-haq:
@cptvitamin @symbolist could you please review it

symbolist:
We will be merging some major changes to ORA in the next few days. So lets hold off on this for now.

PR #678, opened 2015-02-20
Ahsan/tnl 65 impossible to toggle the state

ahsan-ul-haq:
The peer and self assessment can be toggled using mouse but was
impossible to toggle using keyboard so tab index='0' added to solve
this issue

TNL-65


ahsan-ul-haq:
test this please

ahsan-ul-haq:
test me

PR #683, opened 2015-03-16
Update assessments format when serializing.

symbolist:


symbolist:
@muzaffaryousaf Can you give this a quick look?

waheedahmed:
lgtm :+1:

openassessment/xblock/xml.py
```diff
@@ -760,7 +761,7 @@ def serialize_examples_to_xml_str(assessment):
         A unicode string of the XML serialized examples.
 
     """
-    examples = assessment.get('examples', [])
+    examples = update_assessments_format([assessment])[0].get('examples', [])
```
muzaffaryousaf: @symbolist i can't understand why we are choosing 0th element. can you please explain?

muzaffaryousaf:
other than one small question.  :+1:

openassessment/xblock/xml.py
symbolist: `update_assessments_format` takes a list of assessments so we put the `assessment` in a list, and after the list has been updated we retrieve the `assessment` with `[0]`.

PR #686, opened 2015-03-25
Update NLTK to 2.0.5

singingwolfboy:
To work with https://github.com/edx/edx-platform/pull/7465


feanil:
Changelog?

singingwolfboy:
@feanil: There is no change in the code between 2.0.4 and 2.0.5. As you can see in this issue (https://github.com/nltk/nltk/issues/824), NLTK was having compatibility problems with setuptools, which caused the project developers to apply this patch (https://github.com/nltk/nltk/pull/827) to 2.0.4 in order to create 2.0.5. The only change is in setup.py. There is essentially zero risk to this change -- if NLTK 2.0.5 installs, then it will work just as well as as NLTK 2.0.4 (because it is the same code).

feanil:
:+1:   Cool makes sense.

benpatterson:
:+1:

PR #697, opened 2015-05-07
Update ease

singingwolfboy:
Make sure that we use the version of ease that has `pytz==2015.2` and `nltk==2.0.5`. See https://github.com/edx/edx-platform/pull/7465#issuecomment-99917399

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -27,4 +27,4 @@ South==1.0.1
 voluptuous==0.8.5
 
 # AI grading
-git+https://github.com/edx/ease.git@c6dee053eae6b3ac4fdf6be11fa7a9f8265540aa#egg=ease
+git+https://github.com/edx/ease.git@15ebafb570ae63e8e2c1438ede494e82a5528d17#egg=ease


requirements/base.txt
```diff
@@ -27,4 +27,4 @@ South==1.0.1
 voluptuous==0.8.5
 
 # AI grading
-git+https://github.com/edx/ease.git@c6dee053eae6b3ac4fdf6be11fa7a9f8265540aa#egg=ease
+git+https://github.com/edx/ease.git@15ebafb570ae63e8e2c1438ede494e82a5528d17#egg=ease
```
feanil: Since this doesn't use -e, we need to pin the requirement and on the ease side update the version number.

singingwolfboy: [The version for ease hasn't changed in a year](https://github.com/edx/ease/blame/master/setup.py#L37): is that a problem?

feanil:
Closing this since https://github.com/edx/edx-ora2/pull/699 supersedes this.

PR #702, opened 2015-05-28
Fix broken integration tests. TE-892

benpatterson:
@jzoldak @andy-armstrong

### DIFF
--- a/test/acceptance/pages.py
+++ b/test/acceptance/pages.py
@@ -144,7 +144,7 @@ class AssessmentPage(OpenAssessmentPage):
 
     @property
     def is_on_top(self):
-        # TODO: fix this. Behavior of being on top, as well as what the means, appear to be against tests
+        # TODO: On top behavior needs to be better defined. It is defined here more accurately as "near-top".
         # pos = self.browser.get_window_position()
         # return pos['y'] < 100
         # self.wait_for_element_visibility(".chapter.is-open", "Chapter heading is on visible", timeout=10)
--- a/test/acceptance/pages.py
+++ b/test/acceptance/pages.py
@@ -98,7 +98,7 @@ class SubmissionPage(OpenAssessmentPage):
         Returns:
             bool
         """
-        preview_latex_button_class = self.q(css="button#submission__preview").attrs('class')
+        preview_latex_button_class = self.q(css="button#submission__preview").attrs('class')[0]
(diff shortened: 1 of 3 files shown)


benpatterson:
I'm expecting another PR that will take care of using staging. This works on my sandbox for now.

test/acceptance/pages.py
```diff
@@ -144,8 +144,11 @@ def is_browser_on_page(self):
 
     @property
     def is_on_top(self):
-	pos = self.browser.get_window_position()
-	return pos['y'] < 100
+        # TODO: fix this. Behavior of being on top, as well as what the means, appear to be against tests
```
jzoldak: I'm not sure what this comment means...?

jzoldak:
@benpatterson it looks like travis ran twice but one failed? Not sure what's going on there. Also would it be possible to quiet those deprecation warnings? Not in the original scope, but it does make the output harder to mentally parse.

benpatterson:
@jzoldak the duplicate builds are being fixed as part of #703. Perhaps we can merge that and I can rebase?

I can take a look at the deprecation warnings. They seemed to be related to the xblock version but I did not look too closely.

jzoldak:
Sorry I wasn't clear - I meant to highlight the failure (not that the tests ran twice). Seems like that may be a concern?

benpatterson:
Oh, I see. Looks like we have a flaky test. I will see if I can reproduce, though, that does make the deprecation noise more of an issue...I'll likely look at that first.

benpatterson:
@jzoldak I spent some cycles looking at quieting the deprecation warning, but it has to do with the XBlock requirement. Even upgrading our version of XBlock to the latest does not quiet the warning, and I'm not sure where the implementation is. Looks like it's in the workbench, but that is not explicitly called out in the latest version of the Xblock. @cpennington any advice?

I think as far as this PR goes, we'd need some more fundamental adjustments to ora2 in order to quiet the deprecation warning. I don't see a way to silence it without using mocks.

jzoldak:
@benpatterson maybe we could quiet the XBlock logging [the way we do it for noisy loggers in the bok-choy repo](https://github.com/edx/bok-choy/blob/master/bok_choy/__init__.py)?

benpatterson:
Addressing the noisy logs in a separate PR ( #704 )

jzoldak:
:+1:

PR #701, opened 2015-05-22
Don't use sudo on Travis CI

singingwolfboy:

PR #703, opened 2015-05-28
Fix Travis builds

benpatterson:


benpatterson:
@jzoldak @feanil @andy-armstrong This should fix up our Travis builds on ora2.

CC @singingwolfboy

singingwolfboy:
@benpatterson: can you explain why this works? I thought that we weren't able to use sudo on Travis CI, due to their [new container-based infrastructure](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/). I started removing sudo in #701, but ran into some problems. I'm glad that this is working, but I'm also confused.

benpatterson:
Hey @singingwolfboy. Although Travis is now using docker, it is something we'd have to opt into. We're doing that on some repos, and here's a nice clean example: https://github.com/edx/edx-analytics-data-api/pull/69. 

I do think this .travis.yml could be refactored for the better, including some packages that may be available by default on images now. 

In terms of this particular PR, I believe the force-yes option on apt was causing a problem. That might be because there is some default (always-yes), or because -qq already covers it. I did not dig deeper; just followed the [Travis doc](http://docs.travis-ci.com/user/build-configuration/#Installing-Packages-Using-apt). I'm calling a larger refactoring out of scope--just want to have serviceable builds for other development right now.

jzoldak:
:+1:

feanil:
:+1:

PR #706, opened 2015-06-02
Update course with latex enabled.

benpatterson:
Latex was not enabled on the imported course; however we have a test that expects it
to be enabled.

Also update various names to match the contents of the tar.


benpatterson:
@symbolist @andy-armstrong @wedaly 

I believe there is still more Jenkins work after this, but we are close to getting the ora2 integration tests back online.

benpatterson:
@fredsmith @clytwynec could I get a look on this ditty?

wedaly:
:+1:

clytwynec:
:+1:  Looks good to me.

PR #707, opened 2015-06-30
Replacing focus on important elements

clrux:
This PR removes the styles that remove focus on links and buttons. It relates to [UX-2323](https://openedx.atlassian.net/browse/UX-2323) and addresses an accessibility issue in which there were no focus styles in the ORA component and place on the screen was lost when navigating through the assessment questions.

__To preview...__
* in your devstack (platform, not the orastack), open `edx-platform/requirements/edx/github.txt`
* copy the commit hash for this branch and replace the has on `line 41` (the ora2 requirement)
* restart your localhost LMS

_Example:_
```
-e git+https://github.com/edx/edx-ora2.git@079fae7e756aa64a3319019867bd60fe61420f10#egg=edx-ora2
```

You will need to create an ORA component to test with.

__Notes:__ I may have gone overboard with the focus outlines (bringing them back pretty much everywhere). If this is a problem, specifically in the LMS, let me know.

---

## Reviewers
- [x] @talbs (FED/Sass and ORA familiarity)


.gitignore
```diff
@@ -62,3 +62,5 @@ test/selenium/screenshots/*
 
 # logging
 logs/*.log*
+openassessment/xblock/static/css/openassessment-ltr.css.map
```
talbs: Hmm, I'd prefer these not to be added. Can you update your local Sass compiler not to create sourcemaps?

talbs: Or if you want to keep them, which may be a good idea, can you make this more general? Something like ``*.css.map`` makes sense.

talbs:
@clrux, also, can you make sure your Sass compiler compiles into compressed CSS format? The files you have here are uncompressed.

.gitignore
clrux: I don't recall adding this...

talbs: Hmm, they're green in this diff, which means they were added with this work or are based on work that is different from master. Feel free to remove them if they are not needed.

talbs:
@clrux, :+1: once you figure out:

* .css.map ignore files - either remove them from the listing or make them more generic
* compressing your checked-in CSS files.

talbs:
Sorry for the delay, :+1: !

PR #711, opened 2015-07-29
add bok choy accessibility tests

clytwynec:


clytwynec:
@benpatterson @jzoldak  These are the accessibility tests for ora2 (TE-962).  It looks like I'll need to upgrade phantomjs on Jenkins to get them working there, but wanted to tag you on this now so there's more time for code review.

scripts/jenkins-acceptance-tests.sh
```diff
@@ -37,3 +37,4 @@ pip install -r requirements/test-acceptance.txt
 
 cd test/acceptance
 python tests.py
+SELENIUM_BROWSER=phantomjs python accessibility.py
```
benpatterson: So this would mean all acceptance tests go with phantom. I don't have an objection but wanted to doublecheck...

benpatterson: Should probably give a heads up to Usman, who is currently on the hook for any failures in this pipeline.

benpatterson:
Given the challenges you faced, can anything be improved (e.g., documentation)? Even if there is just a link or reference in the README to acceptance tests (and README) being in a subdirectory (for example), that might make things a bit easier next time someone needs to jump in here.

The test code LGTM.

scripts/jenkins-acceptance-tests.sh
clytwynec: Only the accessibility ones would run under phantomjs.  (The line above this runs the others with the default firefox browser.)

benpatterson: Oh, right....sorry I didn't catch that. Cool.

clytwynec:
@benpatterson I just updated the READMEs too.

scripts/jenkins-acceptance-tests.sh
```diff
@@ -37,3 +37,10 @@ pip install -r requirements/test-acceptance.txt
 
 cd test/acceptance
 python tests.py
+
+# Unset SELENIUM_HOST so that bok-choy doesn't try to use saucelabs
+unset SELENIUM_HOST
+# AutoAuthPage times out in PhantomJS when using https, switch to use http
```
clytwynec: @benpatterson @jzoldak  I've update this and got a build working against the existing sandbox.  It looks like phantomjs handles ssl a bit differently than firefox, but I'm not sure how we have it set up.  (See http://phantomjs.org/troubleshooting.html)  For now, I've got it running the accessibility tests with http and the others with https.

jzoldak: @clytwynec do you have a link to a failing build?

clytwynec: @jzoldak  [here's a failing one](http://jenkins.edx.org:8080/view/All/job/ora2-acceptance-tests-copy/4/)

benpatterson: Interesting...I'm curious how old your devstack is... There were some TLS/SSL changes to buildouts...maybe 6 months ago?

clytwynec: I wasn't running in devstack

benpatterson: ah, gotcha....

benpatterson:
:+1:

PR #712, opened 2015-08-07
Updated Django REST Framework requirement

clintonb:
XCOM-524

@clytwynec @singingwolfboy

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -16,7 +16,7 @@ django>=1.4,<1.5
 django-celery==3.1.16
 django-extensions==1.2.5
 django-model-utils==1.4.0
-djangorestframework<2.4
+djangorestframework<3.0
 dogapi==1.2.1
 lazy==1.1
 loremipsum==1.0.5
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,7 @@ PACKAGES = [
     'openassessment.xblock'
 ]
 
+
 def is_requirement(line):
     """
     Return True if the requirement line is a package requirement;
@@ -27,6 +28,7 @@ def is_requirement(line):
         line.startswith('git+')
     )
 
+
 def load_requirements(*requirements_paths):
     """
     Load all requirements from the specified requirements files.
@@ -40,9 +42,10 @@ def load_requirements(*requirements_paths):
         )
     return list(requirements)
 
+
 setup(
     name='ora2',
-    version='0.0.1',
+    version='0.1.0',
     author='edX',
     url='http://github.com/edx/edx-ora2',
     description='edx-ora2',


clintonb:
FYI @feanil

singingwolfboy:
@clintonb if you want to actually upgrade DRF, [you'll need to make another pull request to edx-platform, as well](https://github.com/edx/edx-platform/blob/master/requirements/edx/base.txt#L36). We also have [a checklist for upgrading dependency libraries](https://openedx.atlassian.net/wiki/display/TE/Upgrading+Dependent+Libraries) that you should follow.

clintonb:
@singingwolfboy edx/edx-platform#9227

singingwolfboy:
@clintonb Is it possible for you to first merge *just* the DRF upgrade, and *then* merge the rest of the work in https://github.com/edx/edx-platform/pull/9227? That's a big PR.

clintonb:
I'd rather not deal with yet another PR at the moment. We can handle
edx-platform later. Any objections to merging this PR?

singingwolfboy:
@clintonb it doesn't appear to me that you've done everything in the [upgrading dependent libraries checklist](https://openedx.atlassian.net/wiki/display/TE/Upgrading+Dependent+Libraries), so yes, I have objections to merging this PR. I've done a lot of dependency upgrades for our projects, and I know that they can go really wrong really quickly.

clintonb:
Yeah...I'm too sleep-deprived to deal with that. I'll follow up later.

feanil:
Tests are failing and I agree with @singingwolfboy let's make sure we're not going break everything when this change lands.

feanil:
Please also update setup.py to up the semantic version.

clintonb:
@ormsbee can you walk me through what [NestedModelSerializer](https://github.com/edx/edx-ora2/blob/a4ce7bb00190d7baff60fc90fb613229565ca7ef/openassessment/assessment/serializers/base.py#L25) is doing? I believe that is the source of the test failures.

ormsbee:
Is this urgent? I'm on vacation this week...

clintonb:
Nope. Enjoy your vacation.

ormsbee:
@clintonb: In regards to `NestedModelSerializer`, I'm pretty sure that I mostly copied code from `ModelSerializer`, except that I made the part where it [linked serialization](https://github.com/tomchristie/django-rest-framework/blob/2.2.7/rest_framework/serializers.py#L747) recursive (i.e. extracted that bit of code into its own method and added the recursive call). That was back when we were on 2.2.x of DRF, which had a much simpler `save_object()` method.

That being said, the main reason the class exists in the first case was because DRF didn't seem to support deserialization more than one level deep (serialization was fine). It might be that the more recent versions of DRF have fixed this, and we can just use a straight up `ModelSerializer`. I'd try that first.

PR #715, opened 2015-08-21
Upgrade django-extensions to 1.5.5

wedaly:


wedaly:
@rlucioni please review.  Companion PR to https://github.com/edx/edx-submissions/pull/17

requirements/base.txt
```diff
@@ -6,7 +6,7 @@
 git+https://github.com/edx/XBlock.git@9c634481dfc85a17dcb3351ca232d7098a38e10e#egg=XBlock
 
 # edx-submissions
-git+https://github.com/edx/edx-submissions.git@8fb070d2a3087dd7656d27022e550d12e3b85ba3#egg=edx-submissions
+git+https://github.com/edx/edx-submissions.git@906c7b7a27e1a8684834334a4bf7fb50a3aede30#egg=edx-submissions
```
rlucioni: You may want to give this edx-submissions hash a semantic tag, à la `0.0.8`, and pin the egg with `==0.0.8`. DevOps pushing for this in edx-platform, and it's a common practice in several of our other repos.

rlucioni:
I would tag the changes from https://github.com/edx/edx-submissions/pull/17 as a release of edx-submissions, then use that tag here. However, as far as the version upgrade goes, :+1: .

wedaly:
@rlucioni Would it be okay to switch to tagged versions after the other dependencies are upgraded?  There are four or five more changes I'm going to need to make, and I'd prefer not to add more friction to the process.

rlucioni:
@wedaly, sure, that makes sense. I just wanted to call it out.

PR #716, opened 2015-08-21
Fix an intermittently failing leaderboard test

wedaly:
Saw this test fail a few times locally and in Travis.  The test is comparing two URLs to a fake S3 bucket, each of which contains expiration and signature parameters.  Since these parameters are time-dependent, the test fails when too much time elapses between generating the actual and expected URLs.

@clytwynec would you be able to review this?

### DIFF
--- a/openassessment/xblock/test/test_leaderboard.py
+++ b/openassessment/xblock/test/test_leaderboard.py
@@ -3,18 +3,21 @@
 Tests for leaderboard handlers in Open Assessment XBlock.
 """
 import json
-import mock
-from django.core.cache import cache
 from random import randint
-from submissions import api as sub_api
-from .base import XBlockHandlerTransactionTestCase, scenario
-from django.test.utils import override_settings
-from moto import mock_s3
+from urlparse import urlparse
+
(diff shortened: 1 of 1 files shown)


openassessment/xblock/test/test_leaderboard.py
```diff
@@ -285,3 +292,15 @@ def _assert_leaderboard_visible(self, xblock, is_visible):
         fragment = self.runtime.render(xblock, "student_view")
         has_leaderboard = 'openassessment__leaderboard' in fragment.body_html()
         self.assertEqual(has_leaderboard, is_visible)
+
+    def _clean_score_filenames(self, scores):
+        """
+        Remove querystring parameters from the file name of the score.
+        """
+        from urlparse import urlparse
```
clytwynec: Why import this here instead of at the top of the file?

wedaly: I just forgot to move it up :)  Updated (and standardized the import order)

clytwynec:
:+1:

PR #718, opened 2015-08-27
Upgrade djangorestframework to 3.2.3

wedaly:
Depends on https://github.com/edx/edx-submissions/pull/20

Opening this to run the tests -- I won't merge until edx-platform is ready for the upgrade.

### DIFF
--- a/openassessment/assessment/test/test_peer.py
+++ b/openassessment/assessment/test/test_peer.py
@@ -152,13 +152,14 @@ STEP_REQUIREMENTS = {
     }
 }
 
+
 @ddt
 class TestPeerApi(CacheResetTest):
     """
     Tests for the peer assessment API functions.
     """
 
-    CREATE_ASSESSMENT_NUM_QUERIES = 58
+    CREATE_ASSESSMENT_NUM_QUERIES = 53
 
     def test_create_assessment_points(self):
         self._create_student_and_submission("Tim", "Tim's answer")
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -6,7 +6,7 @@
 git+https://github.com/edx/XBlock.git@9c634481dfc85a17dcb3351ca232d7098a38e10e#egg=XBlock
 
 # edx-submissions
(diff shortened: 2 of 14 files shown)


andy-armstrong:
:+1: Looks good to me.

openassessment/assessment/test/test_peer.py
```diff
@@ -152,13 +152,14 @@
...
 
+
 @ddt
 class TestPeerApi(CacheResetTest):
     """
     Tests for the peer assessment API functions.
     """
 
-    CREATE_ASSESSMENT_NUM_QUERIES = 58
+    CREATE_ASSESSMENT_NUM_QUERIES = 53
```
andy-armstrong: Do you happen to know why the count went down?

requirements/base.txt
```diff
@@ -6,7 +6,7 @@
 git+https://github.com/edx/XBlock.git@9c634481dfc85a17dcb3351ca232d7098a38e10e#egg=XBlock
 
 # edx-submissions
-git+https://github.com/edx/edx-submissions.git@9538ee8a971d04dc1cb05e88f6aa0c36b224455c#egg=edx-submissions
+git+https://github.com/edx/edx-submissions.git@34d6987c6a3b3bd10b3c61c6abdddcb239fd74af#egg=edx-submissions==0.1.0
```
andy-armstrong: Do you generally use hashes rather than release tags?

openassessment/assessment/test/test_peer.py
wedaly: Yep -- we switched from using a custom "nested serializer" to having a single `create()` method on the parent rubric.  The code in the `create()` method needed to make fewer queries to get the same job done, because it could bulk-create the nested models.

requirements/base.txt
wedaly: I'm going to create a new tag when my PR to edx-submissions gets merged, then update this accordingly.

PR #720, opened 2015-09-09
Do not load fonts from googleapis.com since they are already included in edx-platform.

symbolist:
@marcotuts Is there any reason we should keep this?

1. The import is causing issues with the newer version of django-pipeline.
2. Since we host OpenSans with the static assets and that is included in the edx-platform css, this causes an unnecessary request to fetch the fonts again from fonts.googleapis.com.


marcotuts:
@symbolist Thanks for the ping here. I might actually shoot this back over to @talbs or @e0d since I believe they are more familiar with the font-resources & pipeline issues. If we need a story to track this I'm happy to make one but unfortunately I can't provide a review myself.

wedaly:
I think we needed this when doing styling work on ORA2 in workbench.  :+1: on this change -- it's more important to improve performance than to make local development easier.

PR #721, opened 2015-09-14
Django1.8 upgrade

symbolist:
https://openedx.atlassian.net/browse/TNL-2818


openassessment/assessment/api/student_training.py
```diff
@@ -454,7 +454,7 @@ def assess_training_example(submission_uuid, options_selected, update_workflow=T
             raise StudentTrainingRequestError(msg)
 
         # Check the student's scores against the staff's scores.
-        corrections = item.check(options_selected)
+        corrections = item.check_options(options_selected)
```
symbolist: Django added the method `check` on Models so we have to rename ours.

openassessment/tests/data/db_fixtures/feedback_on_assessment.json
```diff
@@ -820,23 +820,5 @@
...
-    }
-  },
-  {
-    "pk": 2,
-    "model": "submissions.scoresummary",
-    "fields": {
-      "highest": 2,
-      "student_item": 1,
-      "latest": 2
-    }
```
symbolist: Signals are now being called during loading of fixtures so these models are automatically created because of this signal handler: https://github.com/edx/edx-submissions/blob/master/submissions/models.py#L233-L271

openassessment/workflow/test/test_api.py
```diff
@@ -361,6 +361,13 @@ def test_unable_to_load_api(self):
         with self.assertRaises(AssessmentWorkflowInternalError):
             workflow_api.create_workflow(submission['uuid'], ['self'], ON_INIT_PARAMS)
```
symbolist: Since we are using `transaction.atomic`, the changes the models in this method call get reverted. Previously, an `AssessmentWorkflow` was getting created. We need to do it explicitly now.

openassessment/xblock/test/test_grade.py
```diff
@@ -327,7 +327,7 @@ def test_submit_feedback_invalid_options(self, xblock):
             'feedback_text': u'I disliked my assessment',
             'feedback_options': u'should be a list!',
         })
-        resp = self.request(xblock, 'submit_feedback', payload, response_format='json')
+        resp = self.request(xblock, 'submit_feedback', payload, response_format='json', use_runtime=False)
```
symbolist: The runtime does a `save()` on the XBlock after the request is complete which messes up the transaction management. See http://stackoverflow.com/a/23326971.

openassessment/fileupload/tests/test_api.py
```diff
@@ -268,7 +268,7 @@ def test_download_url_with_unauthorized_key(self):
         self.assertEqual(404, download_response.status_code)
 
     def test_upload_download_with_accented_key(self):
-        self.set_key("noël.jpg")
+        self.set_key(u"noël.jpg")
         upload_url = self.backend.get_upload_url(self.key, self.content_type)
```
symbolist: We use a unicode key so this should be unicode too.

https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/submission_mixin.py#L243-L254

openassessment/fileupload/backends/base.py
```diff
@@ -123,7 +123,7 @@ def _get_key_name(self, key):
         Returns:
             A key name (str) to use constructing URLs.
         """
-        return "{prefix}/{key}".format(
+        return u"{prefix}/{key}".format(
```
symbolist: I don't know why this was using a byte string instead of a unicode string.

openassessment/fileupload/views_filesystem.py
```diff
@@ -17,8 +17,6 @@ def filesystem_storage(request, key):
     """
     Uploading and download files to the local filesystem backend.
     """
-    if isinstance(key, unicode):
-        key = key.encode("utf-8")
```
symbolist: Once we make the above two changes, we need to make this change too, to get `test_upload_download_with_accented_key()` to pass.

openassessment/fileupload/tests/test_api.py
symbolist: Why this test no longer passes is because of this change in Django 1.5: https://code.djangoproject.com/ticket/19221#comment:7

symbolist:
@muzaffaryousaf @wedaly Can you be the reviewers for this? I am particularly interested for feedback on the unicode related changes.

openassessment/fileupload/backends/base.py
wedaly: That looks like it might have been a mistake.  I'm surprised this didn't cause failures in the current test suite, because I'm pretty sure we test with keys that have non-ASCII values.

wedaly: Ah, I see that test right below this file.

openassessment/fileupload/tests/test_api.py
wedaly: This change makes sense to get the test passing, but I wonder if this is really something we want to support.  Do `boto` and S3 even allow non-ASCII unicode keys?

openassessment/fileupload/views_filesystem.py
wedaly: Hmm, I'm a bit worried about this.  What happens further downstream when S3 receives a key with non-ASCII characters?

PR #728, opened 2015-10-06
(TNL-3436) Removing unnecessary padding from textarea elements

marjev:
JIRA ticket: https://openedx.atlassian.net/browse/TNL-3436

Fixing the issue no.2 reported in the ticket mentioned above. There was a padding style rule applied which caused the issue.

@cahrens, could you please review this, since you commented on JIRA about this, and perhaps suggest the 2nd reviewer? Or just suggest who should be the best person to review this?


edx-webhook:
Thanks for the pull request, @marjev! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-ora2&number=728

cahrens:
Thanks for fixing this, @marjev. I'm not yet familiar with how the ORA CSS is organized... but nobody really is right now due to folks that have moved on! This change looks reasonable, though I don't know why that padding was ever there (will removing it break other things?).

What are the the other 2 files that are being checked in? Compiled versions of the CSS that have to be manually checked in after each update?

FYI, @andy-armstrong

marjev:
@cahrens, I checked the edit modal in studio and it seems to me that nothing got broken, but I'll recheck just in case. Better safe than sorry.

Yes, other two files are the compiled versions of the css.

cahrens:
:+1:

cahrens:
Note though that in order for this fix to be visible, ORA will have to be released. I'm not sure when that will next happen.

marjev:
@cahrens Is there someone I can talk to about ORA release?

cahrens:
We do the ORA releases for edx. In what environment are you experiencing this bug? It was originally reported by Stanford, and it is my understanding that they are running their own branch.

Also, do you need me to merge the PR for you?

marjev:
I'm experiencing this bug on devstack environment.

If 1 thumbsup will suffice for this to be merged, please merge this. Thanks!

PR #729, opened 2015-10-15
Fix jQuery Ajax contentType issue. When sending an ajax request, jque…

mushtaqak:
…ry default contenttype is application/x-www-form-urlencoded; charset=UTF-8. So here sending ?? in the ajax request, it gets encoded in strange manner. So add content-Type to application/json; charset=UTF-8 when sending a json ajax request.

Mentioned in [ORA2 breaks when encountering multiple question marks](https://openedx.atlassian.net/browse/TNL-3437)

**Issue**: `$.ajax` method's default content-type is `application/x-www-form-urlencoded; charset=UTF-8` , which also encodes `?` when sending a ajax request.

**Solution** : explicitly specify `Content-Type` to `application/json; charset=UTF-8` when sending a json request.

@waheedahmed @muzaffaryousaf @symbolist Please review.

@adampalay FYI

### DIFF
--- a/openassessment/xblock/static/js/spec/oa_server.js
+++ b/openassessment/xblock/static/js/spec/oa_server.js
@@ -11,6 +11,8 @@ describe("OpenAssessment.Server", function() {
 
     var server = null;
 
+    var jsonContentType = "application/json; charset=utf-8";
+
     /**
     Stub AJAX requests.
 
@@ -126,7 +128,8 @@ describe("OpenAssessment.Server", function() {
         expect($.ajax).toHaveBeenCalledWith({
             url: '/submit',
             type: "POST",
-            data: JSON.stringify({submission: "This is only a test"})
+            data: JSON.stringify({submission: "This is only a test"}),
+            contentType : jsonContentType
         });
     });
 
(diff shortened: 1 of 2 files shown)


symbolist:
@mushtaqak There are some test failures. Can you look at them?

mushtaqak:
@symbolist Fixed the failing tests. Can you please look at it now ?

@cahrens FYI, this is the PR that handles ora-2 problem when encountering question marks.

cahrens:
Thanks @mushtaqak. :+1:

openassessment/xblock/static/js/spec/oa_server.js
```diff
@@ -126,7 +126,8 @@ describe("OpenAssessment.Server", function() {
         expect($.ajax).toHaveBeenCalledWith({
             url: '/submit',
             type: "POST",
-            data: JSON.stringify({submission: "This is only a test"})
+            data: JSON.stringify({submission: "This is only a test"}),
+            contentType : 'application/json; charset=utf-8'
```
waheedahmed: nit: can you make this a variable like you used in other file?

waheedahmed:
Just a little nit otherwise :+1:

symbolist:
:+1:

PR #732, opened 2015-10-21
[WIP] Fix RemovedInDjango19Warning warnings

muhammad-ammar:
TNL-3450

### DIFF
--- a/openassessment/assessment/models/ai.py
+++ b/openassessment/assessment/models/ai.py
@@ -5,8 +5,9 @@ from uuid import uuid4
 import json
 import logging
 from django.conf import settings
+from django.core import signals
 from django.core.files.base import ContentFile
-from django.core.cache import cache, get_cache
+from django.core.cache import cache, _create_cache
 from django.db import models, transaction, DatabaseError
 from django.utils.timezone import now
 from django_extensions.db.fields import UUIDField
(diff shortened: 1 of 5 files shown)


openassessment/assessment/models/ai.py
```diff
@@ -25,18 +25,12 @@
...
+    caches['cache_in_mem']
 )
 
 CLASSIFIERS_CACHE_IN_FILE = getattr(
     settings, 'ORA2_CLASSIFIERS_CACHE_IN_FILE',
-    get_cache(
-        'django.core.cache.backends.filebased.FileBasedCache',
-        LOCATION='/tmp/ora2_classifier_cache'
-    )
+    caches['cache_in_file']
```
muhammad-ammar: The above changes are working in ora2 but not working in edx-platform because `ora2` settings will not be available there due to which we will get `InvalidCacheBackendError: Could not find config for 'cache_in_mem' in settings.CACHES`. So we have two options:
1. leave the old code as is
2. use [_create_cache](https://github.com/django/django/blob/1.8.4/django/core/cache/__init__.py#L65) function

The issue with 2nd option is that thread safety is not guaranteed but [get_cache in 1.4](https://github.com/django/django/blob/1.4.22/django/core/cache/__init__.py#L141) is already thread unsafe and working fine

@nedbat @symbolist @doctoryes @alawibaba @symbolist @macdiesel Suggestions are welcome. Thanks!

symbolist: @muhammad-ammar If `_create_cache` allows us to retain the existing behaviour, lets use it.

PR #733, opened 2015-10-21
Make improvements to ORA's development environment

andy-armstrong:
@dianakhuang and I have made a number of improvements to support an ORA development environment:
 - added support for using ORA in devstack
 - made it possible to debug ORA and its tests in PyCharm
 - made the XBlock use the raw JavaScript source files when in debug mode (versus the bundled JS)
 - added a .jshintrc file (I haven't added the code to use it yet)
 - documented the process: https://openedx.atlassian.net/wiki/display/TNL/ORA+FAQ

Please review: @dianakhuang @cahrens 

FYI: @efischer19 @symbolist @muhammad-ammar @muzaffaryousaf

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -4,6 +4,7 @@
 set -e
 
 cd `dirname $BASH_SOURCE` && cd ..
+export DJANGO_SETTINGS_MODULE=${DJANGO_SETTINGS_MODULE:-"settings.test_with_coverage"}
 ./scripts/test-python.sh $1
 ./scripts/test-js.sh
 ./scripts/build-docs.sh
--- a/settings/test_with_coverage.py
+++ b/settings/test_with_coverage.py
@@ -1,0 +1,15 @@
+"""
+Django settings for running tests with coverage.
+"""
+
+# Inherit from the test settings
+from .test import *     # pylint:disable=W0614,W0401
+
+# Configure nose so that tests are run with coverage
+NOSE_ARGS = [
+    "-a !acceptance",
+    '--with-coverage',
+    '--cover-package=' + ",".join(TEST_APPS),
+    '--cover-branches',
+    '--cover-erase',
+    ]
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,146 @@
+// --------------------------------------------------------------------
+// JSHint Configuration
+// --------------------------------------------------------------------
+//
+// http://www.jshint.com/
+// http://jshint.com/docs/options/
+{
+    // == Enforcing Options ===============================================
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "edx-ora2",
-  "version": "0.0.1",
+  "version": "0.2.0",
```
cahrens: Why this version change?

setup.py
```diff
@@ -42,7 +42,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='ora2',
-    version='0.0.1',
+    version='0.2.0',
```
cahrens: Why this version change?

settings/test_with_coverage.py
```diff
@@ -0,0 +1,15 @@
+"""
+Django settings for running tests with coverage.
```
cahrens: Why do we want to separate test and test_with_coverage?

package.json
andy-armstrong: I described this in the PR comment. It seemed broken that this had never been bumped. However, this came from the feature branch, so I could move it back as that's where the new functionality is coming.

settings/test_with_coverage.py
andy-armstrong: The issue is that all tests were run with coverage, even when running a single test in the debugger. This change makes the behavior more like edx-platform, where we only run coverage when doing everything at once, i.e. on Travis.

setup.py
andy-armstrong: To match the version in the package. I can revert this one too.

package.json
cahrens: I think it is odd to bump it here, since there is no change in user functionality.

cahrens:
:+1: after you back out the version number change

andy-armstrong:
@cahrens I've removed the version number changes, squashed and pushed up again.

scripts/test.sh
```diff
@@ -4,6 +4,7 @@
 set -e
 
 cd `dirname $BASH_SOURCE` && cd ..
+export DJANGO_SETTINGS_MODULE=${DJANGO_SETTINGS_MODULE:-settings.test_with_coverage}
```
dianakhuang: Should `settings.test_with_coverage` be in quotes? `settings.test` is.

dianakhuang:
:+1: otherwise

symbolist:
@andy-armstrong Can the README be pointed to the instructions on the wiki? I don't think the orastack or workbench are of much use for doing any work on this repo.

andy-armstrong:
@symbolist That's a good point. I'll clean up the documentation.

PR #741, opened 2015-10-26
Run JSHint as part of the build

andy-armstrong:
I decided to resurrect this change but do it in a simper fashion. I've relaxed a couple of the settings that would be too disruptive to change, and then addressed the remaining ones. I think this is a good compromise as it means we can catch real JS issues going forward.

Please review: @cahrens @dianakhuang 
FYI: @symbolist @muhammad-ammar @muzaffaryousaf


cahrens:
:+1:

dianakhuang:
:+1: from me as well.

PR #742, opened 2015-10-27
Minor documentation and user-string fixes.

dianakhuang:
@catong this is the PR we discussed.


catong:
Thanks @dianakhuang :+1:

PR #743, opened 2015-10-28
Add make commands for acceptance and a11y testing

andy-armstrong:
I've added two new commands to the Makefile to simplify testing:

    make test-acceptance
    make test-a11y

@clytwynec @benpatterson Please review.

### DIFF
--- a/scripts/test-acceptance.sh
+++ b/scripts/test-acceptance.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+############################################################
+#
+#   test-acceptance.sh
+#
+#   Runs acceptance tests from a specified file against
+#   an edX platform sandbox.
+#
+#   Note: this script is invoked from the ORA Makefile and
+#   should not normally be used directly.
+#
+#   Usage:
+#
+#       ./test-acceptance.sh {test_file}
+#
+############################################################
+
+cd `dirname $BASH_SOURCE` && cd ..
+
+if [ -z "$BASE_URL" ]; then
+    echo "Error: BASE_URL must be set to point to your sandbox"
+    exit 1
+fi
+
+mkdir -p test/logs
+cd test/logs
+export SELENIUM_BROWSER=phantomjs
+
+test_name="${1:-acceptance}"
+
+echo "Running acceptance tests from ${test_name}.py against the sandbox..."
+
+python ../acceptance/${test_name}.py
--- a/Makefile
+++ b/Makefile
@@ -64,3 +64,9 @@ test-js: render-templates
 
 test-js-debug: render-templates
 	./scripts/js-debugger.sh
+
+test-acceptance:
+	./scripts/test-acceptance.sh tests
+
+test-a11y:
+	./scripts/test-acceptance.sh accessibility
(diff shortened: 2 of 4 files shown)


andy-armstrong:
@clytwynec I realized that a11y testing in edx-platform now runs within the regular Bok Choy tests, rather than standalone. Should we be doing the same thing for ORA? FYI @cahrens

scripts/test-a11y.sh
```diff
@@ -0,0 +1,15 @@
...
+    echo "Error: BASE_URL must be set to point to your sandbox"
+    exit 1
+fi
+
+mkdir -p test/logs
+cd test/logs
+export SELENIUM_BROWSER=phantomjs
+
+echo "Running accessibility tests against the sandbox..."
+python ../acceptance/accessibility.py
```
clytwynec: Since the only thing different about this file and test-acceptance.sh is the directory, can we make it one script and have directory as an arg?

clytwynec:
@andy-armstrong  the a11y tests in platform are written in the same place but they get run separately, so  I think it makes sense to keep it separate here too.  We'd have to change the way the tests are run here to be able to write them in the same file and still run them with separate commands.

scripts/test-a11y.sh
benpatterson: I was wondering the same. I could also see where someone might want to run both, one right after another.

andy-armstrong:
@clytwynec I've followed your suggestion to use just a single script, thanks.

@benpatterson You can do the following to run both. Does this seem reasonable?

    make test-acceptance test-a11y

Eventually we should have a directory full of tests and use ``nose`` to discover them, I suppose.

clytwynec:
Should the [script that runs in the jenkins job](https://github.com/edx/edx-ora2/blob/master/scripts/jenkins-acceptance-tests.sh#L38-L46) use the make commands too?

andy-armstrong:
@clytwynec Another great idea, thanks. Done.

scripts/test-acceptance.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
```
benpatterson: Not a must, but a brief description of what this script does (and what it does with args) would be nice.

benpatterson:
I had a last-minute nit but LGTM.

scripts/test-acceptance.sh
andy-armstrong: Good point. Done.

andy-armstrong:
@clytwynec @benpatterson Once more with feeling... :-) Please review when you get a chance.

clytwynec:
Awesome! :+1:  Thanks for adding this.

scripts/test-acceptance.sh
```diff
@@ -0,0 +1,34 @@
...
+############################################################
+#
+#   test-acceptance.sh
+#
+#   Runs acceptance tests from a specified file against
+#   an edX platform sandbox.
+#
+#   Note: this script is invoked from the ORA Makefile and
```
benpatterson: Solid. Thank you!

benpatterson:
Thanks for taking this up and making things much more straightforward.

:+1:

PR #747, opened 2015-11-02
Implement ORA staff toolbar

andy-armstrong:
## [TNL-3466](https://openedx.atlassian.net/browse/TNL-3466)

Splits the old staff info panel into a toolbar with two options ("Staff Tools" and "Staff Info").

Note: full UX, i18n and accessibility reviews will be performed once the full staff grading feature is complete.

### Sandbox
- [x] http://andy-armstrong.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/4dbd99f691064633a9be0daadb805ce6/

### Testing
- [x] Unit, integration, acceptance tests as appropriate

### Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [x] Code review: @cahrens 
- [x] Code review: @dianakhuang 
- [x] UI strings/error msgs review: @catong
- [x] Product review: @explorerleslie 

FYI: Tag anyone who might be interested in this PR here.

### Post-review
- [x] Squash commits

PR #750, opened 2015-11-03
Fix acceptance tests, and minor doc tweaks

andy-armstrong:
This commit fixes a few issues from my previous ORA PR (https://github.com/edx/edx-ora2/pull/747):
* Updated some doc strings based on feedback from @catong 
* Fixed the acceptance tests which I broke when I introduced DDT (I was running the wrong tests)
* Fixed the Jenkins build to have a DDT dependency
* Renamed the environment variable to ORA_SANDBOX_URL so as to be more explicit

@muzaffaryousaf @cahrens @catong please review.


openassessment/templates/openassessmentblock/staff_area/staff_area.html
```diff
@@ -103,7 +103,7 @@ <h2 class="staff-info__title ui-staff__title">
                 </div>
 
                 <div class="staff-info__status ui-staff__content__section">
-                    <table class="staff-info__status__table" summary="{% trans "Where are your learners currently in this problem" %}">
```
cahrens: Was this string supposed to move someplace else?

test/acceptance/auto_auth.py
```diff
@@ -2,12 +2,13 @@
...
 
+import os
 import re
 import urllib
+
 from bok_choy.page_object import PageObject
 
-import os
-BASE_URL = os.environ.get('BASE_URL')
+ORA_SANDBOX_URL = os.environ.get('ORA_SANDBOX_URL', failobj=os.environ.get('BASE_URL'))
```
cahrens: Is this necessary even with the backward compatibility code in test-acceptance.sh?

test/acceptance/pages.py
```diff
@@ -1,11 +1,13 @@
...
 """
+
+import os
+
 from bok_choy.page_object import PageObject
 from bok_choy.promise import EmptyPromise
 
-import os
-BASE_URL = os.environ.get('BASE_URL')
+ORA_SANDBOX_URL = os.environ.get('ORA_SANDBOX_URL', failobj=os.environ.get('BASE_URL'))
```
cahrens: Same question.

cahrens:
Just a few questions, otherwise :+1:

openassessment/templates/openassessmentblock/staff_area/staff_area.html
andy-armstrong: Reviewers: @catong thought this summary was confusing, and in our investigations it appears that the attribute has been deprecated anyway:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table

catong:
:+1:

muzaffaryousaf:
:+1:  Thanks for the import fix.

openassessment/templates/openassessmentblock/staff_area/staff_area.html
andy-armstrong: @cahrens You beat me to my reviewer comment... see above.

test/acceptance/auto_auth.py
andy-armstrong: I was nervous that it wouldn't work but thinking about it some more it seems that it should. I'll take it out and then only add it back in if the tests still fail.

PR #754, opened 2015-11-05
update django-extenions

muhammad-ammar:


requirements/base.txt
```diff
@@ -12,9 +12,9 @@ git+https://github.com/edx/edx-submissions.git@muzaffar/upgrage-django-1.8#egg=e
 boto>=2.32.1,<3.0.0
 celery==3.1.18
 defusedxml==0.4.1
-django>=1.8,<1.9
+django==1.8.5
```
nedbat: I don't think we should pin django versions in libraries if we can avoid it.

muhammad-ammar: @nedbat Please see the details about the issues at https://openedx.atlassian.net/browse/TNL-3762

PR #755, opened 2015-11-05
Staff score recording

efischer19:
Getting a PR up for collaboration as we work through the backend of staff grading.

FYI @dianakhuang - this is very rough in its current state, but will likely be closely related to what you're working on.


dianakhuang:
@efischer19 i would recommend pointing this at the ora-staff-grading branch

efischer19:
:+1: thanks!

PR #758, opened 2015-11-05
Staff Grading API

dianakhuang:
FYI @andy-armstrong @efischer19


openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,102 @@
...
+            return {'success': False, 'msg': self._(u"Missing options_selected key in request")}
+
+        if 'overall_feedback' not in data:
+            return {'success': False, 'msg': self._('Must provide overall feedback in the assessment')}
+
+        if 'criterion_feedback' not in data:
+            return {'success': False, 'msg': self._('Must provide feedback for criteria in the assessment')}
+
+        if self.submission_uuid is None:
+            return {'success': False, 'msg': self._(u"You must submit a response before you can perform a self-assessment.")}
```
efischer19: Is this check needed for a staff assessment?

dianakhuang: I believe this check is needed, but you're right that I should probably change it to take an arbitrary submission id.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,348 @@
...
+        "secret": "yes",
+        u"ⓢⓐⓕⓔ": "yes",
+        "giveup": "eager",
+        "singing": "no",
+    }
+}
+
+# Answers are against RUBRIC_DICT -- this is worth 14 points
+# Feedback text is one character over the limit.
+LONG_FEEDBACK_TEXT = u"是" * Assessment.MAX_FEEDBACK_SIZE + "."
```
efischer19: This, and several other things in this file seem to be unneeded remnants of what I copy-pasted over to get things started. I'll create a new PR to remove them.

efischer19: https://github.com/edx/edx-ora2/pull/771

openassessment/workflow/models.py
```diff
@@ -335,12 +355,61 @@ def _get_steps(self):
...
+                    assessment_completed_at=now(),
+                )
+            )
         return steps
 
+    def set_staff_score(self, score, is_override=False, reason=None):
+        """
+        Set a staff score for the workflow.
+
+        Allows for staff scores to bet set on a submission, with annotations to provide an audit trail if needed.
```
efischer19: Doc typo: s/bet/be

openassessment/xblock/openassessmentblock.py
```diff
@@ -447,6 +456,10 @@ def _create_ui_models(self):
         """
         ui_models = [UI_MODELS["submission"]]
         for assessment in self.valid_assessments:
+            if assessment["name"] == "staff-assessment" and assessment["required"] == False:
+                # Check if staff have graded the assessment
```
efischer19: Is this an outstanding `TODO`?

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,112 @@
...
+            assessment = staff_api.create_assessment(
+                data['submission_uuid'],
+                self.get_student_item_dict()["student_id"],
+                data['options_selected'],
+                clean_criterion_feedback(self.rubric_criteria, data['criterion_feedback']),
+                data['overall_feedback'],
+                create_rubric_dict(self.prompts, self.rubric_criteria_with_labels)
+            )
+            self.publish_assessment_event("openassessmentblock.staff_assessment", assessment)
+            workflow_api.update_from_assessments(assessment["submission_uuid"], {}, force_update_score=True)
```
efischer19: Just to document the assumption being made here: if every staff assessment calls `update_from_assessments(..., force_update_score=True)`, then the last staff score will be the only score that is used.

So we will never be able to combine a staff score with a peer score to get any sort of composite, the staff score will always win.

efischer19:
Done with my first pass, just a few minor things.

I'm going to work on increasing test coverage for `openassessment/workflow/models.py` in https://github.com/edx/edx-ora2/pull/772, but I don't feel that needs to hold up landing this PR to unblock front-end work.

openassessment/assessment/api/ai.py
```diff
@@ -67,7 +67,8 @@ def get_score(submission_uuid, requirements):
         requirements (dict): Not used.
 
     Returns:
-        A dictionary with the points earned and points possible.
+        A dictionary with the points earned, points possible, and
+        contributing_assessments information.
```
andy-armstrong: I think this should mention ``staff_id`` even though it is always ``None`` in this case. It is returned as part of the dictionary.

openassessment/assessment/api/self.py
```diff
@@ -70,8 +70,8 @@ def get_score(submission_uuid, requirements):
         submission_uuid (str): The unique identifier for the submission
         requirements (dict): Not used.
     Returns:
-        A dict of points earned and points possible for the given submission.
-        Returns None if no score can be determined yet.
+        A dictionary with the points earned, points possible, and
+        contributing_assessments information.
```
andy-armstrong: Nit: ``staff_id`` as well...

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+)
+from openassessment.assessment.serializers import (
+    AssessmentFeedbackSerializer, RubricSerializer,
+    full_assessment_dict, rubric_from_dict, serialize_assessments,
+    InvalidRubric
+)
+from openassessment.assessment.errors import (
+    StaffAssessmentRequestError, StaffAssessmentInternalError
+)
+from submissions import api as sub_api
```
andy-armstrong: Nit: ``sub_api`` is confusing to read so I'd go with ``submissions_api``.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+
+logger = logging.getLogger("openassessment.assessment.api.staff")
+
+
```
andy-armstrong: Nit: extra blank line

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+    Args:
+        submission_uuid (str): The UUID of the submission being graded.
+        requirements (dict): Any variables that may effect this state.
+
+    Returns:
+        True if the assessment has been completed for this submission.
+
+    """
+    if requirements and requirements.get('staff', {}).get('required', False):
```
andy-armstrong: This logic is a little subtle and I'm not sure I fully understand it. 

I believe this is saying that there are two cases:
1. If the problem requires staff grading, then the assessment is only true if an assessment exists, which must by definition be staff graded. I'm not sure if there is a risk that a score was provided by some other means before this problem was made to require staff grading. Maybe it would be better to confirm that the latest assessment is of type staff assessment.
2. If the problem does not require staff grading, then the submission is always considered finished. This seems counter-intuitive to me, because if the problem doesn't require staff grading then doesn't it need to wait for some other kind of assessment? Is it the case that then this code wouldn't be invoked, because it isn't a staff grading workflow, but would instead go to peer or something else? If that's the case, maybe it would be better to throw an exception than to always return True.

andy-armstrong: I'm wondering if this actually is only determining whether the staff assessment step is finished for the submission, not that all assessments are finished.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+        submission_uuid (str): The UUID for the submission to get a score for.
+        requirements (dict): Not used.
+
+    Returns:
+        A dictionary with the points earned, points possible,
+        contributing_assessments, and staff_id information.
+
+    """
+    assessment = get_latest_assessment(submission_uuid)
+    if not assessment:
```
andy-armstrong: What happens if the latest assessment isn't a staff score? It seems odd to me that the staff assessment workflow could return a non-staff score. Is this intended to return the final score, or just the staff score?

openassessment/assessment/api/staff.py
andy-armstrong: I now see that ``get_latest_assessment`` returns only the latest staff assessment. This is confusing, so I recommend renaming it to ``get_latest_staff_assessment``.

openassessment/assessment/api/staff.py
andy-armstrong: As just noted, this is only confusing because ``get_latest_assessment`` doesn't return the latest assessment. Please rename it to ``get_latest_staff_assessment``.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+
+    Args:
+        submission_uuid (str): The UUID of the submission being assessed.
+
+    Returns:
+        dict: The serialized assessment model
+        or None if no assessments are available
+
+    Raises:
+        StaffAssessmentInternalError
```
andy-armstrong: This needs a comment explaining when this error will be raised. It seems a little odd that it will raise an internal error for a non-existent submission id separate from true internal errors. Maybe it would be worth having a SubmissionNotFoundError.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+        raise StaffAssessmentInternalError(msg)
+
+    if len(assessments) > 0:
+        return full_assessment_dict(assessments[0])
+    else:
+        return None
+
+
+def get_assessment_scores_by_criteria(submission_uuid):
+    """Get the score for each rubric criterion
```
andy-armstrong: Nit: "get the *staff* score for each rubric criterion", right?

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+            information from the scores, an error is raised.
+    """
+    try:
+        assessments = list(
+            Assessment.objects.filter(
+                score_type=STAFF_TYPE, submission_uuid=submission_uuid
+            )[:1]
+        )
+        scores = Assessment.scores_by_criterion(assessments)
+        return Assessment.get_median_score_dict(scores)
```
andy-armstrong: Why does this need to get a median score, when the most recent staff score always wins? Is this just a shortcut because we know that only one score will be returned for each assessment? If not, it seems wrong to get the median score from two staff assessments. It would be good to either use a less confusingly named method, or have a comment to explain the assumptions being made.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+    scored_at=None
+):
+    """Creates an assessment on the given submission.
+
+    Assessments are created based on feedback associated with a particular
+    rubric.
+
+    Assumes that the user creating the assessment has the permissions to do so.
+
+    Args:
```
andy-armstrong: This is missing several arguments: ``submission_uuid``, ``rubric_dict`` and ``scored_at``.

openassessment/assessment/api/staff.py
andy-armstrong: I see that other methods return ``StaffAssessmentRequestError`` for non-existent submissions, so it would be good to be consistent.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+            options_selected,
+            criterion_feedback,
+            overall_feedback,
+            rubric_dict,
+            scored_at
+        )
+        return full_assessment_dict(assessment)
+
+    except InvalidRubric:
+        msg = u"Rubric definition was not valid"
```
andy-armstrong: Nit: I'm not sure why this is stored in a variable, but since it is I'd prefer it be called ``error_message`` like it is below.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+        logger.exception(msg)
+        raise StaffAssessmentRequestError(msg)
+    except InvalidRubricSelection:
+        msg = u"Invalid options selected in the rubric"
```
andy-armstrong: Nit: I'm not sure why this is stored in a variable, but since it is I'd prefer it be called ``error_message`` like it is below.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,292 @@
...
+        overall_feedback,
+        rubric_dict,
+        scored_at
+):
+    """
+    Internal function for atomic assessment creation. Creates a staff assessment
+    in a single transaction.
+
+    Args:
+        rubric_dict (dict): The rubric model associated with this assessment
```
andy-armstrong: Nit: these arguments are in the wrong order.

openassessment/assessment/errors/staff.py
```diff
@@ -0,0 +1,33 @@
...
+    """Error indicating insufficient or incorrect parameters in the request.
+
+    Raised when the request does not contain enough information, or incorrect
+    information which does not allow the request to be processed.
+
+    """
+    pass
+
+
+class StaffAssessmentInternalError(StaffAssessmentError):
```
andy-armstrong: It seems a little unnecessary to me to have a unique error class for this. I see that the code already does this, but I think having a single class of ``AssessmentInternalError`` would be simpler. Does the code ever need to distinguish between different kinds of internal errors?

openassessment/assessment/errors/staff.py
```diff
@@ -0,0 +1,33 @@
+"""
+Errors for the staff assessment api.
+"""
+
+
+class StaffAssessmentError(Exception):
```
andy-armstrong: As with the request error class, I wonder if it would be better to have a generic ``AssessmentError`` class that this subclasses. In most situations, the client wants to catch all assessment errors, I would think.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,348 @@
...
+    student_id="Tim",
+    course_id="Demo_Course",
+    item_id="item_one",
+    item_type="Peer_Submission",
+)
+
+ANSWER_ONE = u"this is my answer!"
+
+# Possible points: 14
+RUBRIC_DICT = {
```
andy-armstrong: This looks like it is exactly the same as the one from test_peer.py, so I'd reuse it rather than duplicating it again. Maybe  some of these common pieces should be moved into a helper class.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,348 @@
...
+            tim_sub["uuid"],
+            "Dumbledore",
+            ASSESSMENT_DICT_PASS['options_selected'], dict(), "",
+            RUBRIC_DICT,
+        )
+        # Be sure to update the workflow!
+        workflow_api.update_from_assessments(tim_sub["uuid"], self.STEP_REQUIREMENTS, force_update_score=True)
+
+        # Verify both assessment and workflow report correct score
+        self.assertEqual(staff_assessment["points_earned"], 14)
```
andy-armstrong: It would be good to have scores that weren't the maximum to be sure that they are being derived correctly.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,348 @@
...
+        )
+
+        # Ensure points are calculated properly
+        self.assertEqual(assessment["points_earned"], 6)
+        self.assertEqual(assessment["points_possible"], 14)
+
+        #ensure submission is marked as finished
+        self.assertTrue(staff_api.assessment_is_finished(tim_sub["uuid"], self.STEP_REQUIREMENTS))
+
+    def test_create_assessment_overrides(self):
```
andy-armstrong: I'd like to see a test with a double staff override, to verify that the latest override beats out all others. The subsequent override should provide different scores to make sure that it is a true override, and that the staff scores don't get aggregated together in any way.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,348 @@
...
+        )
+        # Be sure to update the workflow!
+        workflow_api.update_from_assessments(tim_sub["uuid"], self.STEP_REQUIREMENTS, force_update_score=True)
+
+        # Verify both assessment and workflow report correct score
+        self.assertEqual(staff_assessment["points_earned"], 14)
+        workflow = workflow_api.get_workflow_for_submission(tim_sub["uuid"], self.STEP_REQUIREMENTS)
+        self.assertEqual(workflow["score"]["points_earned"], 14)
+
+    def test_create_assessment_does_not_block(self):
```
andy-armstrong: I don't understand exactly what this test is trying to do. It would be good to have a BDD-style docstring explaining the different aspects that it is looking for.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,348 @@
...
+            ASSESSMENT_DICT_PASS['options_selected'], dict(), "",
+            RUBRIC_DICT,
+        )
+        workflow_api.update_from_assessments(tim_sub["uuid"], self.STEP_REQUIREMENTS_WITH_STAFF)
+
+        # Verify assesment made, score updated, and no longer wating
+        self.assertEqual(staff_assessment["points_earned"], 14)
+        self.assertTrue(staff_api.assessment_is_finished(tim_sub["uuid"], self.STEP_REQUIREMENTS_WITH_STAFF))
+
+    def test_invalid_rubric_exceptions(self):
```
andy-armstrong: I'm thinking it would be good to make this test use ``DDT`` so that it can validate a variety of different invalid options. Currently it is passing a whole set of invalid options at once, so it is hard to be sure that each different option is being validated. It could be that only one of them is failing, but the test still sees that as correct behavior.

openassessment/assessment/test/test_staff.py
andy-armstrong: It probably would make sense to have a test that verifies the override behavior for each workflow type, i.e. can it override peer, self, AI etc. I wonder if you could do this with DDT where you pass in a function to create an initial assessment, and then override that.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,348 @@
...
+class TestStaffOverwrite(CacheResetTest):
+    """
+    Tests for staff assessments made as overrides, when none is required to exist.
+    """
+
+    STEPS = ['self']
+    STEP_REQUIREMENTS = {}
+    STEP_REQUIREMENTS_WITH_STAFF = {'staff': {'required': True}}
+
+    def test_create_assessment_points(self):
```
andy-armstrong: I'd use DDT to try out a variety of override values. In particular, it would be good to verify that staff can override to the minimum value (is that always 0?), the maximum value, and an in-between value.

openassessment/workflow/api.py
```diff
@@ -182,8 +182,10 @@ def get_workflow_for_submission(submission_uuid, assessment_requirements):
     return update_from_assessments(submission_uuid, assessment_requirements)
 
 
-def update_from_assessments(submission_uuid, assessment_requirements):
-    """Update our workflow status based on the status of peer and self assessments.
+def update_from_assessments(submission_uuid, assessment_requirements, force_update_score=False):
```
andy-armstrong: I'm wondering whether this ``force_update_score`` is really necessary. Couldn't it be derived to be true if a staff score is found? It seems less brittle to me to keep that logic within this function, than to remember to pass the correct value in every place that handles staff scoring. I could also imagine that for some reason the staff score didn't get applied correctly, so a subsequent update should notice it and pick it up.

openassessment/workflow/models.py
```diff
@@ -335,12 +355,61 @@ def _get_steps(self):
             # If no steps exist for this AssessmentWorkflow, assume
             # peer -> self for backwards compatibility
             self.steps.add(
-                AssessmentWorkflowStep(name=self.STATUS.peer, order_num=0),
-                AssessmentWorkflowStep(name=self.STATUS.self, order_num=1)
+                AssessmentWorkflowStep(name=self.STATUS.peer, order_num=1),
```
andy-armstrong: These order numbers seem brittle. What happens if the ``ORA2_ASSESSMENT_SCORE_PRIORITY`` setting changes the order? It seems that this should probably be based upon ``ASSESSMENT_SCORE_PRIORITY``. Do you know if there are tests that cover this logic?

openassessment/workflow/models.py
```diff
@@ -49,12 +50,14 @@
 # We then use that score as the student's overall score.
 # This Django setting is a list of assessment steps (defined in `settings.ORA2_ASSESSMENTS`)
 # in descending priority order.
-DEFAULT_ASSESSMENT_SCORE_PRIORITY = ['peer', 'self', 'ai']
+DEFAULT_ASSESSMENT_SCORE_PRIORITY = ['staff', 'peer', 'self', 'ai']
 ASSESSMENT_SCORE_PRIORITY = getattr(
```
andy-armstrong: What happens for installations that have already set this setting? I'm wondering if the code should always add 'staff' to the front of the list, regardless of what has been configured, because our UI assumes that staff grades will trump all others.

andy-armstrong: Maybe we should just drop this setting since it seems to open up a lot of potential bugs... it isn't clear to me that all of our code really obeys this setting, and there aren't any tests for what happens when it is set.

openassessment/workflow/models.py
andy-armstrong: Typo: "Allows for staff scores to bet set".

openassessment/workflow/models.py
```diff
@@ -335,12 +355,61 @@ def _get_steps(self):
...
+            self.steps.add(
+                AssessmentWorkflowStep(
+                    name=self.STATUS.staff,
+                    order_num=0,
+                    assessment_completed_at=now(),
+                )
+            )
         return steps
 
+    def set_staff_score(self, score, is_override=False, reason=None):
```
andy-armstrong: Why do we need special logic for setting staff scores? I would have thought it would be enough to submit a staff assessment and then the workflow would just know how to set the appropriate score.

openassessment/workflow/models.py
```diff
@@ -353,11 +422,27 @@ def set_score(self, score):
                 'points_possible'.
 
         """
-        sub_api.set_score(
-            self.submission_uuid,
-            score["points_earned"],
-            score["points_possible"]
-        )
+        if not self.staff_score_exists():
```
andy-armstrong: Why does this check have to be done? Won't the workflow ensure that the staff score will beat out any other assigned scores?

openassessment/xblock/openassessmentblock.py
```diff
@@ -66,6 +67,12 @@
         "navigation_text": "Your assessment of your response",
         "title": "Assess Your Response"
     },
+    "staff-assessment": {
+        "name": "staff-assessment",
+        "class_id": "openassessment__staff-assessment",
```
andy-armstrong: I'd been hoping to remove these special ids, but I hadn't noticed before that the Python code had some knowledge of them too.

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,112 @@
...
+    This mixin is for all staff-assessment related endpoints.
+    """
+
+    @XBlock.json_handler
+    @require_course_staff("STUDENT_INFO")
+    def staff_assess(self, data, suffix=''):
+        """
+        Create a staff assessment from a staff submission.
+        """
+        if 'options_selected' not in data:
```
andy-armstrong: Can we share some of this validation logic with the other endpoints? They all seem to be doing the same thing (except that some of them don't require a submission_uuid.

openassessment/xblock/test/test_staff.py
```diff
@@ -0,0 +1,126 @@
...
+
+        for key in self.ASSESSMENT:
+            assessment_copy = copy.copy(self.ASSESSMENT)
+            del assessment_copy[key]
+            resp = self.request(xblock, 'staff_assess', json.dumps(assessment_copy), response_format='json')
+            self.assertFalse(resp['success'])
+            self.assertIn('msg', resp)
+
+    @scenario('data/self_assessment_scenario.xml', user_id='bob')
+    def test_staff_assess_assessment_error(self, xblock):
```
andy-armstrong: Nit: ``test_staff_assess_assessment_error`` is hard to read. Given that all these tests are inside the ``TestStaffAssessment`` class, I think they could lose the redundant ``staff_access`` portion of their names.

andy-armstrong:
I'm finished with my first pass of this code, and it looks fantastic! There are a lot of subtleties that I'm not sure I fully comprehend so I look forward to discussing my comments.

openassessment/assessment/api/ai.py
efischer19: I'll update this for all 4 apis, good catch

openassessment/assessment/api/staff.py
efischer19: So what happens with waiting is that `status_details` (link: https://github.com/edx/edx-ora2/blob/diana/staff-grading-api/openassessment/workflow/models.py#L186) generates a dictionary of each applicable step and whether its done or not. That dictionary determines whether the entire submission is done, this particular method is a subset of that that answers "is the staff assessment step finished?".

By marking non-required staff grading as always complete, we avoid any potential issue where the workflow would have to wait on it, and can immediately proceed to the next step in order.

openassessment/assessment/test/test_staff.py
efischer19: Yeah, that would be much cleaner. I'll get working on that.

@dianakhuang - I'll handle all the feedback on the tests, since I wrote them initially.

openassessment/workflow/models.py
efischer19: We're currently in the workflow model, this is how we ensure that no other scores can be recorded after a staff score has been recorded.

openassessment/workflow/models.py
efischer19: This is the workflow knowing how to set the appropriate score - it needs special logic to add the annotation parameters when calling into `edx-submissions`

efischer19:
@robrap @scottrish Feel free to look over this PR as well. The JIRA ticket is https://openedx.atlassian.net/browse/TNL-3463

openassessment/xblock/openassessmentblock.py
dianakhuang: I think this is. It looks like we probably want to render it if the staff assessment exists.

openassessment/assessment/api/staff.py
dianakhuang: I just reused the original code, which I think was just a shortcut instead of doing the 'right' thing, which is to just return a single score dict. It is enforcing only using one score to ensure that's the one that 'wins' when calculating the median score dict.

openassessment/assessment/errors/staff.py
dianakhuang: The staff assessment mixin handles them slightly differently.

openassessment/assessment/api/staff.py
dianakhuang: It seems like this is the same logic that self-assessment uses. If we change it here, we should change it there too.

openassessment/assessment/api/staff.py
dianakhuang: If no submissions or assessments are found, it will return `None` and not error out. That error is caught for issues with contacting the database in the first place. I'll update the comment to make that clearer.

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,119 @@
...
+from xblock.core import XBlock
+
+from openassessment.assessment.api import staff as staff_api
+from openassessment.workflow import api as workflow_api
+from openassessment.assessment.errors import (
+    StaffAssessmentRequestError, StaffAssessmentInternalError
+)
+
+from .data_conversion import create_rubric_dict
+from .resolve_dates import DISTANT_FUTURE
```
cahrens: DISTANT_FITURE is unused import

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,119 @@
...
+            submission_uuid = data.get("submission_uuid")
+            path, context = self.self_path_and_context(submission_uuid)
+        except:
+            msg = u"Could not retrieve staff assessment for submission {}".format(self.submission_uuid)
+            logger.exception(msg)
+            return self.render_error(self._(u"An unexpected error occurred."))
+        else:
+            return self.render_assessment(path, context)
+
+    def staff_path_and_context(self, submission_uuid):
```
cahrens: submission_uuid is unused

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        Retrieve the correct template path and template context for the handler to render.
+
+        Args:
+            submission_uuid (str) -
+        """
+        #TODO: add in the workflow for staff grading instead of assuming it's allowed.
```
cahrens: When will this TODO be addressed?

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,119 @@
...
+        submission = submission_api.get_submission(self.submission_uuid)
```
cahrens: Where does submission_api come from (there is no import)?

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,119 @@
...
+
+        context = {'allow_latex': self.allow_latex}
+        context["rubric_criteria"] = self.rubric_criteria_with_labels
+        context["estimated_time"] = "20 minutes"  # TODO: Need to configure this.
```
cahrens: Also curious when this TODO will be addressed.

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @XBlock.handler
+    @require_course_staff("STUDENT_INFO")
+    def render_staff_assessment(self, data, suffix=''):
+        """
+        Render the staff assessment for the given student.
+        """
+        try:
+            submission_uuid = data.get("submission_uuid")
+            path, context = self.self_path_and_context(submission_uuid)
```
cahrens: submission_uuid is ignored in the method. How does self.submission_uuid get populated?

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,119 @@
...
+                data['overall_feedback'],
+                create_rubric_dict(self.prompts, self.rubric_criteria_with_labels)
+            )
+            self.publish_assessment_event("openassessmentblock.staff_assessment", assessment)
+            workflow_api.update_from_assessments(assessment["submission_uuid"], {}, force_update_score=True)
+
+        except StaffAssessmentRequestError:
+            logger.warning(
+                u"An error occurred while submitting a staff assessment "
+                u"for the submission {}".format(self.submission_uuid),
```
cahrens: Again, I'm confused about self.submission_uuid (vs. the other ways it is retrieved, for instance via assessment["submission_uuid"] on line 62).

openassessment/xblock/staff_assessment_mixin.py
dianakhuang: The best way to understand `self.submission_uuid` is to understand that it's the `submission_uuid` of the current xblock, which is matched up with the submission of the current user who's viewing the xblock. The reason why we want to fetch a separate `submission_uuid` in the staff mixin methods is because the staff member is acting on any student's submission, not just their own.

openassessment/xblock/staff_assessment_mixin.py
dianakhuang: Good catch. This error message is using the incorrect submission uuid.

openassessment/workflow/api.py
efischer19: @explorerleslie - I just want to clarify the product expectations here. Is it true that anytime a staff score is present, that will always be used as the overriding score?

explorerleslie: Yes, that's true.

efischer19:
@andy-armstrong @dianakhuang I'm done with my first set of review responses. Diana, let's touch base in the morning about how to proceed from here, if you're all set I think we're ready for another round of review.

openassessment/workflow/models.py
efischer19: @andy-armstrong - I now understand what you mean here a little better I think. It's not strictly necessary for this check to be here, the workflow `get_score()` function will still be working correctly due to the staff API coming first in priority order.

However, by keeping this check and using `reset_score` in `set_staff_score` above, we ensure that the score reported by the submissions API remains correct in the event that a different consumer asks for the current score on a given submission.

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,296 @@
...
+    """
+    Determine if the staff assessment step of the given submission is completed.
+    This checks to see if staff have completed the assessment.
+
+    Args:
+        submission_uuid (str): The UUID of the submission being graded.
+        requirements (dict): Any variables that may effect this state.
+
+    Returns:
+        True if a staff assessment has been completed for this submission.
```
robrap: Minor addition: " or if not required."

openassessment/assessment/api/staff.py
andy-armstrong: @dianakhuang I don't see a change to this docstring yet. Do you have it in your changes?

openassessment/assessment/api/staff.py
```diff
@@ -0,0 +1,299 @@
...
+        logger.exception(msg)
+        raise StaffAssessmentInternalError(msg)
+
+    if len(assessments) > 0:
+        return full_assessment_dict(assessments[0])
+    else:
+        return None
+
+
+def get_assessment_scores_by_criteria(submission_uuid):
```
andy-armstrong: Nit: rename to ``get_staff_scores_by_certeria``.

openassessment/assessment/api/staff.py
andy-armstrong: Maybe just add a comment here to say that this doesn't actually pick the median of all the staff scores.

openassessment/assessment/api/staff.py
dianakhuang: This is named to be consistent with the other (self/ai) versions of the same function, so I'm not sure it makes sense to change it.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,342 @@
...
+        with self.assertRaises(StaffAssessmentRequestError) as context_manager:
+            staff_assessment = staff_api.create_assessment(
+                tim_sub["uuid"],
+                "Dumbledore",
+                OPTIONS_SELECTED_DICT["most"]["options"], dict(), "",
+                invalid_rubric,
+            )
+        self.assertEqual(str(context_manager.exception), u"Rubric definition was not valid")
+
+    def test_invalid_rubric_options_exception(self):
```
andy-armstrong: I'd still like this to be DDT test with a variety of different invalid options. I think it would be clearer that all of the validation is happening, and it would be more extensible.

cahrens:
@dianakhuang @efischer19 I'm seeing the following bug:

1) Learner submits self assessment and peer assessments, peers assess (this step probably not necessary)
2) Staff does override of learner
3) Learner tries to assess more peers-- server error

This is 100% reproducible for me, and I verified that the submission that is erroring CAN be assessed by a learner without an override. @explorerleslie said that after override, the learner should still be able to assess peers (and in the current UI, there is no indication of why the assessment failed).

openassessment/xblock/staff_assessment_mixin.py
```diff
@@ -0,0 +1,71 @@
...
+            workflow_api.update_from_assessments(assessment["submission_uuid"], {})
+
+        except StaffAssessmentRequestError:
+            logger.warning(
+                u"An error occurred while submitting a staff assessment "
+                u"for the submission {}".format(data['submission_uuid']),
+                exc_info=True
+            )
+            msg = self._(u"Your staff assessment could not be submitted.")
+            return {'success': False, 'msg': msg}
```
andy-armstrong: It is a shame that these APIs end up returning a 200 with the failure hidden inside the JSON response. I'm not sure that all of the code is aware of this distinction, as it still needs to handle true server errors as well.

I know this is how all the ORA XBlock endpoints work (and maybe all XBlock endpoints in general) so it is more of an observation than a request to change anything. I have to make sure that all UI AJAX handlers know how to handle ``success: false`` as well as server errors.

andy-armstrong:
:+1: This is excellent work, @dianakhuang and @efischer19. Thanks for addressing all of my comments.

openassessment/xblock/staff_assessment_mixin.py
dianakhuang: Yeah, I'm not a big fan of this either. :crying_cat_face: I know it's an older pattern that the LMS endpoints used to use before we started using HTTP error codes. I don't know why it was reused here, but it was.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,331 @@
...
+
+        # Staff assess
+        staff_assessment = staff_api.create_assessment(
+            tim_sub["uuid"],
+            "Dumbledore",
+            OPTIONS_SELECTED_DICT[key]["options"], dict(), "",
+            RUBRIC,
+        )
+
+        # Verify assesment made, score updated, and no longer wating
```
andy-armstrong: Typo: wating

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,331 @@
...
+            OPTIONS_SELECTED_DICT[key]["options"], dict(), "",
+            RUBRIC,
+        )
+
+        # Ensure points are calculated properly
+        self.assertEqual(assessment["points_earned"], OPTIONS_SELECTED_DICT[key]["expected_points"])
+        self.assertEqual(assessment["points_possible"], RUBRIC_POSSIBLE_POINTS)
+
+        #ensure submission is marked as finished
+        self.assertTrue(staff_api.assessment_is_finished(tim_sub["uuid"], self.STEP_REQUIREMENTS))
```
andy-armstrong: There seems to be a bug on my sandbox where recording a staff score isn't enough for the workflow to say that it is finished. It would be good if this actually verified that the workflow thinks that the whole thing is finished, in addition to the Staff API considering itself finished.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,331 @@
...
+        staff_assessment = staff_api.create_assessment(
+            tim_sub["uuid"],
+            "Dumbledore",
+            OPTIONS_SELECTED_DICT[key]["options"], dict(), "",
+            RUBRIC,
+        )
+
+        # Verify assesment made, score updated, and no longer wating
+        self.assertEqual(staff_assessment["points_earned"], OPTIONS_SELECTED_DICT[key]["expected_points"])
+        self.assertTrue(staff_api.assessment_is_finished(tim_sub["uuid"], self.STEP_REQUIREMENTS_WITH_STAFF))
```
andy-armstrong: This test should also verify that the workflow believes that it is finished.

cahrens:
What are our thoughts about updating the developer doc for the new methods (http://edx.readthedocs.org/projects/edx-ora-2/en/latest/)? @catong Do you have input on this (I see these docs are marked "deprecated").

dianakhuang:
@cahrens I was actually asked by @catong to remove those docs from the repo, so there's not really a way to update them at the moment unless we make the deliberate decision to bring them back.

catong:
@cahrens All doc is in the edx-documentation repo now, and there's no current plan to resurrect the developer doc for ORA; it has been removed from any published doc for a while now. What doc content are you thinking of, and for what audience?

openassessment/assessment/test/test_staff.py
efischer19: I'll update this test in the "how a workflow defines done" PR that we're breaking out, https://github.com/edx/edx-ora2/pull/778

openassessment/assessment/test/test_staff.py
efischer19: Same, see https://github.com/edx/edx-ora2/pull/778

efischer19:
I've finished reviewing/responding. I'll keep working on the "what does it mean for a workflow to be done in the context of staff assessments" bug in https://github.com/edx/edx-ora2/pull/778, otherwise we're good to squash and merge this from my end.

:+1:

catong:
@cahrens This architecture/API ORA doc stopped being published at least a year ago and is not listed on docs.edx.org so we already consider it "gone". :-) The link you have is to the RTD project, which remains, but isn't listed on the docs site. 
API doc publishing is currently on hold pending a strategy for APIs across the organization, but it would be great for Dev to continue documenting API docstrings within the .py files if you can. We can chat if you have other questions or suggestions. Thanks!

PR #759, opened 2015-11-05
Staff scores endpoint

efischer19:
Laying out the skeleton of how I think this will go. set_staff_score
is broken out into its own method.


efischer19:
FYI @dianakhuang - I thought it made more sense to have this be a PR on top of your changes.

openassessment/workflow/models.py
```diff
@@ -352,24 +356,74 @@ def _get_steps(self):
             steps = list(self.steps.all())
         return steps
 
-    def set_score(self, score):
+    def set_staff_score(self, score, staff_name, is_required=False, is_override=False, reason=None):
```
dianakhuang: I'm not sure why we need both `is_required` and `is_override`.

openassessment/workflow/models.py
```diff
@@ -352,24 +356,74 @@ def _get_steps(self):
...
+                will be used in the event that this parameter is not provided.
 
         """
+        if is_required == is_override:
+            raise StaffAssessmentRequestError('Staff scores must provide one and only one of is_requred and is_override')
+        if not staff_name:
+            raise StaffAssessmentRequestError('Staff scores must provide the name of the staff member')
+        annotation_type = None
+        if is_required:
+            annotation_type = "Required staff score"
```
dianakhuang: I think it makes sense to make these constants and use underscores in the string constants to make the fact that they're a type a little clearer. Something like "required_staff_score".

openassessment/workflow/models.py
efischer19: Good point. I suppose we'll never have neither (if it's not required or an override, what is it?) or both, so we can assume one and treat "not that" as the other.

openassessment/workflow/models.py
efischer19: You got it

efischer19:
Previous changes w.r.t. waiting for staff assessments have been pulled out into https://github.com/edx/edx-ora2/pull/764, per offline conversation with @dianakhuang

openassessment/workflow/api.py
```diff
@@ -261,9 +261,11 @@ def update_from_assessments(submission_uuid, assessment_requirements):
 
     """
     workflow = _get_workflow_model(submission_uuid)
+    staff_required = assessment_requirements.get("staff", {}).get('staff', {}).get('required', False)
+    is_staff_override = workflow.staff_score_exists() and not staff_required
```
efischer19: I'm not 100% sure how to write this logic. What if we want to override a previously recorded required staff score? This current setup would not allow for that, we'd have to add an explicit `force_override` to this API method as well

dianakhuang: I think it might make sense to have force_override be explicit on the api level as well, and then call the api from `staff_assess` in the staff assessment mixin.

openassessment/workflow/api.py
```diff
@@ -182,7 +182,7 @@ def get_workflow_for_submission(submission_uuid, assessment_requirements):
     return update_from_assessments(submission_uuid, assessment_requirements)
 
 
-def update_from_assessments(submission_uuid, assessment_requirements):
+def update_from_assessments(submission_uuid, assessment_requirements, force_update_score=False):
```
dianakhuang: Don't forget to update the docstring for this.

openassessment/workflow/models.py
```diff
@@ -287,15 +292,14 @@ def update_from_assessments(self, assessment_requirements, force_update_score=Fa
...
-        # If the status is done or cancelled, check if score has changed.
-        if self.status == self.STATUS.done:
-            if force_update_score:
-                new_score = self.get_score(assessment_requirements, step_for_name)
-                self.set_score(new_score)
-                self.save()
-                logger.info((
-                    u"Workflow for submission UUID {uuid} has updated score."
-                ).format(uuid=self.submission_uuid))
+        # If the status is done, check if we need to override the previosu score
```
dianakhuang: NIt: `previosu` -> `previous`

openassessment/workflow/models.py
```diff
@@ -364,11 +411,24 @@ def set_score(self, score):
...
-            score["points_possible"]
-        )
+        if not self.staff_score_exists():
+            sub_api.set_score(
+                self.submission_uuid,
+                score["points_earned"],
+                score["points_possible"]
+            )
+
+    def staff_score_exists(self):
```
dianakhuang: Don't forget a docstring for this. I'm also wondering if there's anyway to clean up all the `None`s we check here.

openassessment/workflow/models.py
```diff
@@ -279,23 +284,21 @@ def update_from_assessments(self, assessment_requirements, force_update_score=Fa
...
+        if force_update_score:
+            new_score = self.get_score(assessment_requirements, step_for_name)
+            self.set_staff_score(new_score, is_override=True)
+            self.save()
+            logger.info((
+                u"Workflow for submission UUID {uuid} has updated score."
+            ).format(uuid=self.submission_uuid))
+            return
+
+        if self.status == self.STATUS.cancelled or self.status == self.STATUS.done:
```
efischer19: I had meant to ask this earlier - are we going to allow staff overrides on cancelled workflows, or should I keep the previous behavior of checking for `STATUS.cancelled` before handling `force_update_score`?

openassessment/workflow/models.py
```diff
@@ -352,6 +358,47 @@ def _get_steps(self):
...
+        """
+        Set a staff score for the workflow.
+
+        Allows for staff scores to bet set on a submission, with annotations to provide an audit trail if needed.
+        This method can be used for both required staff grading, and staff overrides.
+
+        Args:
+            score (dict): A dict containing 'points_earned', 'points_possible', and 'staff_name'.
+            staff_name (string): The name of the staff member who is giving the score.
+            is_required (bool, default False): True if setting a required staff score.
```
efischer19: outdated docs

openassessment/xblock/test/data/self_assessment_scenario.xml
```diff
@@ -43,6 +43,7 @@
         </criterion>
     </rubric>
     <assessments>
+        <assessment name="staff-assessment" required="False" />
```
efischer19: I'm a bit worried that I had to add this to make the staff assessment available in test runs, is there some other way to specify it as always available? I'm going to check into that a bit before I write up more thorough staff tests.

requirements/base.txt
```diff
@@ -6,7 +6,7 @@
 git+https://github.com/edx/XBlock.git@9c634481dfc85a17dcb3351ca232d7098a38e10e#egg=XBlock
 
 # edx-submissions
-git+https://github.com/edx/edx-submissions.git@14aeaa9e30f9a408b34ffaf6d78409dedaad015a#egg=edx-submissions==0.1.0
+git+https://github.com/edx/edx-submissions.git@87aaf568400255ad62ef3bebef6a3e6063f700e0#egg=edx-submissions==0.1.0
```
efischer19: To get this to take:

* `pip install -U edx-submissions --force-reinstall`
* `make install`

openassessment/assessment/api/peer.py
```diff
@@ -45,7 +45,7 @@ def submitter_is_finished(submission_uuid, requirements):
         bool
 
     """
-    if requirements is None:
+    if not requirements:
```
dianakhuang: Is there a reason why we're checking it this way? Generally, if we want to check for `None` we want to be explicit about that because it'll cause weird behavior with `falsy` values.

efischer19: The code around here: https://github.com/edx/edx-ora2/blob/efischer/staff_endpoint/openassessment/workflow/models.py#L203 can send `{}`, which causes errors at line 55 below if we don't check for that case as well.

Would it be better to check for `None` and `{}` individually?

dianakhuang: Nope, that should be okay. I just wanted to make sure that we wanted to behave this way when we get `{}` too.

openassessment/workflow/models.py
```diff
@@ -346,12 +355,61 @@ def _get_steps(self):
...
+            sub_dict = sub_api.get_submission_and_student(self.submission_uuid)
+            sub_api.reset_score(
+                sub_dict['student_item']['student_id'],
+                self.course_id,
+                self.item_id
+            )
+        else:
+            if reason is None:
+                reason = "A staff score was required for this submission"
+        print sub_api
```
dianakhuang: I'm guessing this was here for testing purposes and you don't need it anymore?

openassessment/assessment/api/peer.py
```diff
@@ -183,12 +184,15 @@ def get_score(submission_uuid, requirements):
...
+    assessments = [item.assessment for item in items]
 
     return {
         "points_earned": sum(
             get_assessment_median_scores(submission_uuid).values()
         ),
-        "points_possible": items[0].assessment.points_possible,
+        "points_possible": assessments[0].points_possible,
+        "contributing_assessments": [assessment.id for assessment in assessments],
+        "staff_name": None,
```
dianakhuang: Does it make a whole lot of sense to call this `staff_name`? Maybe we should just have it as `scorer_id`?

dianakhuang: Or `staff_id`?

requirements/base.txt
dianakhuang: I think what we'll want to do is to cut a new release for edx-submissions and up the version number so we won't have to force an update this way.

requirements/base.txt
```diff
@@ -6,7 +6,7 @@
 git+https://github.com/edx/XBlock.git@9c634481dfc85a17dcb3351ca232d7098a38e10e#egg=XBlock
 
 # edx-submissions
-git+https://github.com/edx/edx-submissions.git@14aeaa9e30f9a408b34ffaf6d78409dedaad015a#egg=edx-submissions==0.1.0
+git+https://github.com/edx/edx-submissions.git@3d3ba2a4ba2d64f37dd1d908c8a1d78c1ac95523#egg=edx-submissions==0.1.1
```
efischer19: confirmed to be upgradeable via `make install` on an otherwise-already-set-up devstack.

openassessment/assessment/api/peer.py
efischer19: I like `staff_id`, I'll update it

efischer19:
@dianakhuang, this is ready for full review.

I've put a few tests into https://github.com/edx/edx-ora2/blob/efischer/staff_endpoint/openassessment/assessment/test/test_staff.py - do you think we should flesh this test plan out more fully before merging, or just get the bare minimum in to unblock @cahrens and @andy-armstrong and work on adding fuller test coverage in parallel?

openassessment/workflow/models.py
```diff
@@ -346,12 +355,59 @@ def _get_steps(self):
...
+            if reason is None:
+                reason = "A staff score was required for this submission"
+        sub_api.set_score(
+            self.submission_uuid,
+            score["points_earned"],
+            score["points_possible"],
+            annotation_creator = score["staff_id"],
+            annotation_type = annotation_type,
+            annotation_reason = reason
+        )
```
efischer19: Do we need to worry about transactionality here? Theoretically a peer or other type of assessment could be recorded between `reset_score` and `set_score`. So
* does that matter, or will the fact that `staff` comes first in `get_score` priority save us anyways?
* how would we go about fixing this is if does matter? I'd think we'd need a new submissions api method that would allow for "reset and set" to happen in a single database transaction.

openassessment/assessment/test/test_staff.py
```diff
@@ -0,0 +1,226 @@
...
+        assessment = staff_api.create_assessment(
+            tim_sub["uuid"],
+            "Dumbledore",
+            ASSESSMENT_DICT['options_selected'], dict(), "",
+            RUBRIC_DICT,
+        )
+        self.assertEqual(assessment["points_earned"], 6)
+        self.assertEqual(assessment["points_possible"], 14)
+
+        #ensure submission is marked as finished now
```
dianakhuang: Are you missing another assertion here? Is this comment still accurate?

efischer19: Yeah, I forgot to write that assertion, good catch.

I've included it in the test PR I'm working on here: https://github.com/edx/edx-ora2/pull/768

openassessment/workflow/models.py
```diff
@@ -346,12 +355,59 @@ def _get_steps(self):
...
+        try:
+            self.steps.get(name=self.STATUS.staff)
+        except AssessmentWorkflowStep.DoesNotExist:
+            self.steps.add(
+                AssessmentWorkflowStep(
+                    name=self.STATUS.staff,
+                    order_num=0,
+                    assessment_completed_at=now(),
+                )
+            )
```
dianakhuang: Could you put in a comment explaining why we always want to add a staff step?

openassessment/workflow/models.py
dianakhuang: In the example you used, all the staff member would have to do would be to re-submit the grade, which would put in a new reset and a new grade. Not perfect, but definitely would work.

The problem with transactions is that they only really work in the context of views, because they're enforced through Django middleware. Also, how transactions are being handled is changing between Django 1.4 and Django 1.8, so it might not be worth noting until all those changes become the default.

openassessment/workflow/models.py
```diff
@@ -364,11 +420,27 @@ def set_score(self, score):
...
+        if not self.staff_score_exists():
+            sub_api.set_score(
+                self.submission_uuid,
+                score["points_earned"],
+                score["points_possible"]
+            )
+
+    def staff_score_exists(self):
+        """
+        Check if a staff score exists for this submission. Cached after first use.
```
dianakhuang: How is this getting cached? I'm happy to hear it is, but I'm not sure what the mechanism is for it.

dianakhuang:
That's all the comments I have from this pass.

openassessment/workflow/models.py
efischer19: Yeah, I think I'm also of the opinion that it's a) rare enough and b) easily remedied by re-submitting, so we probably don't have to worry about it.

If we do end up needing to address it, it'd be a simple enough change to the `edx-submissions` api to use `select_for_update` and `transaction.atomic` in a new endpoint.

openassessment/workflow/models.py
efischer19: Will do

openassessment/workflow/models.py
efischer19: Ack, more outdated documentation. I had tried to include that, but it wasn't working and I wasn't sure it would be invalidated regularly enough for the risk of a stale cache to outweigh the gain of not needing to make a database query on each call.

I'll look into this a bit more and try to add it, unless you think the stale-cache risk is too great.

dianakhuang: If it's not working, then I understand. It'd be nice if it could be cached, but if it's too much of a pain to invalidate it, then I'd err on the side of not caching it.

openassessment/workflow/models.py
```diff
@@ -346,12 +355,61 @@ def _get_steps(self):
...
             # peer -> self for backwards compatibility
             self.steps.add(
-                AssessmentWorkflowStep(name=self.STATUS.peer, order_num=0),
-                AssessmentWorkflowStep(name=self.STATUS.self, order_num=1)
+                AssessmentWorkflowStep(name=self.STATUS.peer, order_num=1),
+                AssessmentWorkflowStep(name=self.STATUS.self, order_num=2)
             )
             steps = list(self.steps.all())
+
+        # A staff step will always be available, to allow for staff overrides
```
efischer19: Is this statement (a staff step will always be available) true? Or do we need to selectively enable it based on course settings?

efischer19:
In case it gets lost in the impending rebase shuffle, I just merged some new tests in from https://github.com/edx/edx-ora2/pull/768

openassessment/workflow/models.py
dianakhuang: I think it is always true that a staff member can always override grades, but maybe it only makes sense to add this workflow step when a staff assessment exists?

efischer19: I'm not sure it's possible for a staff assessment to exist without this step being available? I tried just adding a `self.staff_score_exists()` check here, but that quickly blows up the stack due to recursion, since that method calls into `self._get_steps()`

dianakhuang: Bah, I understand that, but I'm wondering if this will make things weird for problems where a staff doesn't come through and override the score. Will ora2 think the problem is incomplete?

efischer19: Oh, I see what you're saying now. I think we're covered there - that's why I set `assessment_completed_at=now()` when we do this, so `is_assessment_complete` will return true. https://github.com/edx/edx-ora2/blob/efischer/staff_endpoint/openassessment/workflow/models.py#L587

Will this cover the situation you're talking about? Do you think we should do the same for `submitted_completed_at`?

dianakhuang: I think that makes sense. I guess the other concern is when we're rendering this on the front end. If the list of steps for a problem includes the staff step all the time, does this mean that it will always get rendered? Or do we want to put in an exception here?

efischer19: I'm 100% on board for adding an exception if needed. Will this "setting of the workflow steps" we're discussing here be critical to that, or will we still be able to check the problem definition separately? I guess in my head it was always:

* Does the problem require a staff grade? Check the problem definition
* Does a staff grade exist? Assume a staff grading step is in place, and check if it has a grade

I could see where this could cause problems if `self.steps` is treated as the canonical source of problem requirements, that just wasn't my understanding.

PR #760, opened 2015-11-06
Add re-grading support to the Staff Tools modal

andy-armstrong:
## [TNL-3467](https://openedx.atlassian.net/browse/TNL-3467)

This change introduces the ability to re-grade an ORA submission from the Staff Tools modal.

Note: this PR is ready for product review, but not yet for full engineering review.

### Sandbox
- [x] https://andy-armstrong.sandbox.edx.org/courses/course-v1:AndyA+AAORA+1/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/

### Testing
- [x] Data is properly encoded in HTML templates to avoid XSS
- [x] Analytics (N/A)

### Reviewers
- [x] Code review: @cahrens 
- [x] UI strings/error msgs review: @catong 
- [x] Product review: @explorerleslie 
- UX review has been deferred: @frrrances 

FYI: @cahrens @dianakhuang @efischer19 @robrap @scottrish @symbolist @muhammad-ammar @muzaffaryousaf 

### Post-review
- [x] Squash commits


cahrens:
@explorerleslie If you can take a look at http://fa.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/4dbd99f691064633a9be0daadb805ce6/ (use staff tools to look at the user "staff") and give some initial feedback on the following, I would appreciate it:

1. Use of blue titles for collapsible/expandable sections (standard in at least the LMS view)
2. Border around each of the collapsible/expandable student info sections
3. New design for "Remove submission from peer grading"

openassessment/templates/openassessmentblock/staff_area/student_info.html
```diff
@@ -2,72 +2,285 @@
...
+                    </div>   
+                </div>
+            </div>
+        </div>
 
-                <div class="staff-info__cancel-submission__content ui-toggle-visibility__content">
-                    <div class="ui-staff__content__section">
-                        <div class="wrapper--input">
+        {% if not workflow_cancellation %}
+            <div id="openassessment__staff-info__cancel__submission"
```
cahrens: @andy-armstrong Why have this ID-- openassessment__staff-info__cancel__submission? We want to avoid introducing IDs if possible, correct?

explorerleslie:
@andy-armstrong I'm not a fan of the blue titles or all the borders. It looks crowded and overwhelming to the user. We should match the wireframes: https://openedx.atlassian.net/wiki/display/TNL/%5BUX+Wireframes%5D+-+ORA+Staff+Grading?preview=/42599275/42599267/Staff-override-3.png (Sorry if this is a WIP and you're getting to it.)

The "Remove student..." box looks good with some minor visual tweaks. 

I think that this would be a good PR for @frrrances to give a look at from an incremental step -- Frances this is the "Staff" view of ORA staff grading that we have in the UX backlog.

andy-armstrong:
Thanks @explorerleslie. The blue titles and borders were from @cahrens and @talbs so I'll let them talk to those.

cahrens:
@talbs had mentioned that anything that should collapse/expand should be blue. And note that in the wireframes, both "Remove Submission from Grading" and "Assessment Override" are blue. I'm not getting why the other state state collapsible items wouldn't follow that pattern. Opinion, @frrrances?

As far as borders, "Student Response" already had a border (it was pink, and @frrrances suggested changing it to grey/white), so I made the other sections have borders to be consistent. To me it looks odd to not be able to tell where the collapsible/expandable area ends. But again, I will let @frrrances decide.

talbs:
> @talbs had mentioned that anything that should collapse/expand should be blue. And note that in the wireframes, both "Remove Submission from Grading" and "Assessment Override" are blue. I'm not getting why the other student state collapsible items wouldn't follow that pattern. Opinion, @frrrances?
> 
> As far as borders, "Student Response" already had a border (it was pink, and @frrrances suggested changing it to grey/white), so I made the other sections have borders to be consistent. To me it looks odd to not be able to tell where the collapsible/expandable area ends. But again, I will let @frrrances decide.

@cahrens, looks like you've got the whole UX team tagged on this one, so I'll leave you in the hands of the always capable @frrrances and will defer to her judgement on these crucial details. 

If it helps, **my recommendation about consistent visual treatment (i.e. blue text color) of any titles that toggle an area to collapse/expand was given in just the context of the student-facing ORA UI (in a separate piece of work), which is completely on white/light backgrounds**. For the staff UI that is being handled here, I agree with @explorerleslie that light text on the dark background is more appropriate.

Good luck and let me know if any of you need early-on ORA2 technical or design background.

cahrens:
Thanks, @talbs. @frrrances I removed the blue titles that I had added and the extra borders, so it's back to what the wireframes looked like. The sandbox is updated.

cahrens:
@frrrances I pushed up the changes we talked about (except for the input box moving down where it belongs and the changes to the last 2 sections, which @andy-armstrong will do). I've also updated the sandbox.

openassessment/xblock/static/sass/oa/elements/_staff.scss
```diff
@@ -44,6 +61,13 @@
     color: $copy-staff-color;
   }
 
+  .ui-staff__subtitle {
+    @extend %t-subheading;
+    @include fontSize($f-size-medium);
+    font-weight: 600;
```
frrrances: Found the variable for this: 
```
font-weight: %t-strong;
```
also add this inside the rule: 
```
span {
font-weight: inherit;
}
```
since you need to override the normal font-weight set on all spans.

openassessment/xblock/static/sass/oa/elements/_staff.scss
```diff
@@ -44,6 +61,13 @@
...
+    margin-bottom: ($baseline-v/4);
```
frrrances: double this to ($baseline-v/2);

openassessment/xblock/static/sass/oa/utilities/_developer.scss
```diff
@@ -135,27 +136,121 @@ $link-hover: $edx-blue-l1 !default; // from our Pattern Library http://ux.edx.or
...
-    @include padding(($baseline-v/2), ($baseline-h/2), ($baseline-v/2), ($baseline-h/2));
+    @include padding(($baseline-v/2), 0, ($baseline-v/2), 0);
     margin-bottom: ($baseline-v/2);
   }
 
-  .openassessment__student-info_list {
+  .staff-info__student__report {
     list-style-type: none;
+
+    h2.title {
```
frrrances: remove h2 selector. add some padding and boldness:
```
padding: ($baseline-v/2) ($baseline-h/2);
font-weight: %t-strong;
```
and this needs the span fix as well: 

```
span {
font-weight: inherit;
}
```

openassessment/xblock/static/sass/oa/utilities/_developer.scss
```diff
@@ -135,27 +136,121 @@ $link-hover: $edx-blue-l1 !default; // from our Pattern Library http://ux.edx.or
...
-   * LMS specific styles on HTML elements.
-   */
+  .ui-staff__action__section {
+    padding: $baseline-v ($baseline-h/2);
+    background-color: white;
+    margin-bottom: ($baseline-v/2);
+
+    .ui-staff__title {
+      text-transform: none;
+      font-size: 18px;
```
frrrances: Remove font-size, font-weight, and line-height and use typography mixins: 
``
@extend %t-large;
font-weight: %t-strong;
```

File: openassessment/xblock/static/sass/oa/utilities/_developer.scss [hunk 6]
```diff
@@ -135,27 +136,121 @@ $link-hover: $edx-blue-l1 !default; // from our Pattern Library http://ux.edx.or
...
 
-  // 'p' elements in LMS have a color set on them.
-  .student__answer__display__content p {
-    color: inherit;
+    .title {
+      color: $edx-gray-d4;
+    }
+
+    .title--sub {
+      font-size: 18px;
```
frrrances: Use extend here for font-size: 
```
@extend %t-large;
```

frrrances:
@cahrens Looks much better! I put a few small comments inline but I think it's on track now. Let me know if you need more help or when this is ready for final review.

openassessment/xblock/static/sass/oa/elements/_staff.scss
cahrens: This doesn't seem to work-- there's something else overriding it with a font-weight of 400. I'll check in what you are requesting, but I will need help getting it to apply.

openassessment/xblock/static/sass/oa/utilities/_developer.scss
cahrens: This one doesn't seem to be working either.

openassessment/xblock/static/sass/oa/utilities/_developer.scss
cahrens: @andy-armstrong you added this CSS, but I don't see that there are any elements with class title--sub within class ui-staff__action__section. I suspect that this CSS is unnecessary. Can you take a look?

openassessment/xblock/static/sass/oa/utilities/_developer.scss
cahrens: @andy-armstrong please also take a look at this feedback (I suspect much of this CSS will change or be deleted since Frances would like the last 2 sections to look like the rest when collapsed (black).

openassessment/xblock/static/sass/oa/elements/_staff.scss
cahrens: Actually, I wasn't reading the inspector correctly. It looks like it is in fact working.

openassessment/xblock/static/sass/oa/elements/_staff.scss
cahrens: Done

openassessment/xblock/static/sass/oa/utilities/_developer.scss
cahrens: I was wrong with this one too-- it is working. However, the left padding looks odd to me. I can show you when we meet later today.

openassessment/templates/openassessmentblock/oa_rubric.html
```diff
@@ -23,7 +23,8 @@ <h4 class="question__title ui-toggle-visibility__control">
                                    name="{{ criterion.name }}"
                                    id="assessment__rubric__question--{{ criterion.order_num }}__{{ option.order_num }}"
                                    class="answer__value"
-                                   value="{{ option.name }}" />
+                                   value="{{ option.name }}"
+                                   aria-labelledby="assessment__rubric__prompt--{{ criterion.order_num }}"/>
```
cahrens: @cptvitamin Let me know if there is a better way to do this.

openassessment/templates/openassessmentblock/response/oa_response.html
```diff
@@ -139,10 +139,10 @@ <h3 class="message__title">{% trans "We could not submit your response" %}</h3>
 
                 <ul class="list list--actions">
                     <li class="list--actions__item">
-                        <a aria-role="button" href="#" id="step--response__submit"
-                            class="action action--submit step--response__submit {{ submit_enabled|yesno:",is--disabled" }}">
+                        <button type="submit" id="step--response__submit"
```
cahrens: @cptvitamin I converted this anchor tag to a button. I added type="submit" because other similar buttons had that. Let me know if it is incorrect.

openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html
```diff
@@ -18,17 +18,18 @@ <h2 class="staff-info__title ui-staff__title">
...
-                            <label for="openassessment__student_username" class="label">{% trans "Enter an individual learner's username or email" %}</label>
-                            <input id="openassessment__student_username" type="text" class="value" maxlength="255">
+                        <form class="openassessment_student_info_form">
+                            <label class="label">{% trans "Enter an individual learner's username or email" %}
+                                <input type="text" class="openassessment__student_username value" maxlength="255">
+                            </label>
                             <ul class="list list--actions">
                                 <li class="list--actions__item">
-                                    <a aria-role="button" href="" id="submit_student_username" class="action--submit"><span class="copy">{% trans "Submit" %}</span></a>
+                                    <button class="action--submit action--submit-username"><span class="copy">{% trans "Submit" %}</span></button>
```
cahrens: @andy-armstrong I think this should have type="submit" because it is related to a form (?).

openassessment/templates/openassessmentblock/response/oa_response.html
cptvitamin: Excellent.  Thanks for doing that.  adding type=submit is optional as that is the implied value.  If you don't declare a type, it is treated like a submit.   This is generally what you want, and even if that is not what you want (submit buttons are used with forms) it usually doesn't cause any issues if the button is just a trigger for a javascript event and is not related to  a form.  It's up to you if you want to explicitly declare this.

openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html
cptvitamin: @andy-armstrong @cahrens  :+1:  thanks for converting these.  See https://github.com/edx/edx-ora2/pull/760#discussion_r44467716 for more info on type=submit

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -157,108 +186,146 @@
...
+         */
         cancelSubmission: function(submissionUUID) {
             // Immediately disable the button to prevent multiple requests.
             this.cancelSubmissionEnabled(false);
             var view = this;
-            var sel = $('#openassessment__student-info', this.element);
-            var comments = sel.find('#staff-info__cancel-submission__comments').val();
-            this.server.cancelSubmission(submissionUUID, comments).done(
-                function(msg) {
+            var comments = this.element.find('.cancel_submission_comments').val();
```
cahrens: This line is throwing an error.

Uncaught TypeError: this.element.find is not a function

cahrens:
When a regrading is done, or a student's submission is removed, the "Final Grade" section within the staff tools will need to update to reflect the change.

The easiest way to do the update will be to re-render the staff tools sections (though we have to think about how to handle expanding the collapsible sections).

Currently on removing a submission, a confirmation message is shown explaining how to re-enable another submission for that student. @andy-armstrong and I are considering moving that message into the Learner Response area, so that it is always available (and we can refresh data without worrying about the loss of the message). FYI @explorerleslie, as this is what is currently shown after the staff tools are manually refreshed:

image

Note that the date is broken currently in master (both in the staff tools section and the student-facing section). I will create a separate bug for that.

openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html
cahrens: I retract my comment then-- thanks @cptvitamin.

openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html
```diff
@@ -65,22 +68,22 @@ <h2 class="staff-info__title ui-staff__title">
                     </div>
 
                     <div class="staff-info__status ui-staff__content__section">
-                        <a aria-role="button" href="" id="schedule_training" class="action--submit"><span class="copy">{% trans "Schedule Example-Based Assessment Training" %}</span></a>
-                        <div id="schedule_training_message"></div>
+                        <button class="action--submit action--submit-training"><span class="copy">{% trans "Schedule Example-Based Assessment Training" %}</span></button>
```
cahrens: @andy-armstrong did you test that this schedule training stuff works now without the IDs? I'm not sure how to get into that scenario.

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
```diff
@@ -0,0 +1,52 @@
+{% load tz %}
+{% load i18n %}
+{% spaceless %}
```
cahrens: What is spaceless?

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,305 @@
...
+                            {% endfor %}
+                        {% endfor %}
+                        </tbody>
+                    </table>
+                </div>
+            </div>
+        </div>
+    {% endif %}
+
+    {% if example_based_assessment %}
```
cahrens: We should test with an example based assessment to make sure this HTML is good.

andy-armstrong:
@explorerleslie @catong This PR is ready for design review on my sandbox:

https://andy-armstrong.sandbox.edx.org/courses/course-v1:AndyA+AAORA+1/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/

andy-armstrong:
Question for folks: what should happen with staff override for a peer assessment where the user hasn't yet submitted all of their peer reviews? Currently the staff override doesn't apply which is a confusing user experience because even with the override the grade remains incomplete. @explorerleslie @frrrances @cahrens

explorerleslie:
@andy-armstrong I found a weird edge case that we'll need to address. If the same submission is shown in both the "assess peers" and "staff override" areas, any edits to the "staff override" area jump you to the "assess peers" section. I haven't been able to test if this is the case if the submissions AREN'T the same but are just both displayed.

andy-armstrong:
@explorerleslie I'd noticed that once but didn't again and thought it wasn't an issue. It turns out that clicking on the radio button itself doesn't scroll, but clicking on the label does. This should be simple to fix. If you could continue your product review now, that would be great. Thanks!

openassessment/templates/openassessmentblock/edit/oa_edit.html
```diff
@@ -24,7 +24,7 @@ <h2 class="openassessment_alert_title">{% trans "Rubric Change Impacts Settings
                 <p class="openassessment_alert_message">{% trans "A change that you made to this assessment's rubric has an impact on some examples laid out in the settings tab. For more information, go to the Settings section and fix areas highlighted in red." %}</p>
             </div>
             <a href="" rel="view" class="action openassessment_alert_close">
-                <i class="icon-remove-sign"></i>
+                <i class="icon fa fa-times-circle" aria-hidden="true"></i>
```
cahrens: Note to reviewers: this icon missed the fa conversion. I verified that it does work with the new classes.

openassessment/templates/openassessmentblock/edit/oa_edit_criterion.html
```diff
@@ -2,7 +2,7 @@
 {% spaceless %}
 <li class="openassessment_criterion is-collapsible" data-criterion="{{ criterion_name }}">
     <div class="openassessment_container_header openassessment_criterion_header view-outline">
-        <a class="action expand-collapse collapse"><i class="icon-caret-down ui-toggle-expansion"></i></a>
```
cahrens: Note to reviewers: this icon missed the fa conversion. When I changed it the icon did appear, but the functionality did not work (the section did not collapse when the icon or title beside it was clicked on). I created a bug to fix this if product desires, and am removing the icon so it does not look confusing.

https://openedx.atlassian.net/browse/TNL-3799

openassessment/templates/openassessmentblock/edit/oa_edit_prompt.html
```diff
@@ -3,7 +3,7 @@
 <li class="openassessment_prompt is-collapsible" data-prompt="{{ prompt_uuid }}">
 
     <div class="openassessment_container_header openassessment_prompt_header view-outline">
-        <a class="action expand-collapse collapse"><i class="icon-caret-down ui-toggle-expansion"></i></a>
```
cahrens: Note to reviewers: this icon missed the fa conversion. When I changed it the icon did appear, but the functionality did not work (the section did not collapse when the icon or title beside it was clicked on). I created a bug to fix this if product desires, and am removing the icon so it does not look confusing.

https://openedx.atlassian.net/browse/TNL-3799

openassessment/templates/openassessmentblock/edit/oa_training_example.html
```diff
@@ -3,9 +3,7 @@
 <li class="openassessment_training_example is-collapsible">
 
     <div class="openassessment_training_example_header view-outline">
-        <a class="action expand-collapse collapse">
```
cahrens: Note to reviewers: this icon missed the fa conversion. When I changed it the icon did appear, but the functionality did not work (the section did not collapse when the icon or title beside it was clicked on). I created a bug to fix this if product desires, and am removing the icon so it does not look confusing.

https://openedx.atlassian.net/browse/TNL-3799

andy-armstrong:
@cahrens @dianakhuang This PR is ready for review now. I've updated the sandbox too with fixes for the issue @explorerleslie reported, and for https://openedx.atlassian.net/browse/TNL-3786.

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
```diff
@@ -0,0 +1,52 @@
...
+                    <div class="staff-assessment__display">
+                        <header class="staff-assessment__display__header">
+                            <h3 class="staff-assessment__display__title">
+                                {% blocktrans %}
+                                    Response for: {{ student_username }}
+                                {% endblocktrans %}
+                            </h3>
+                        </header>
+
+                        {% include "openassessmentblock/oa_submission_answer.html" with answer=submission.answer answer_text_label="The learner's response to the question above:" %}
```
cahrens: Does answer_text_label get translated?

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
```diff
@@ -0,0 +1,52 @@
...
+
+                        {% include "openassessmentblock/oa_uploaded_file.html" with file_upload_type=file_upload_type file_url=staff_file_url header="Associated File" class_prefix="staff-assessment" show_warning="true" %}
```
cahrens: Does header get translated?

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
```diff
@@ -0,0 +1,52 @@
...
+                    </div>
+
+                    <form class="staff-assessment__assessment" method="post">
+                        {% include "openassessmentblock/oa_rubric.html" with rubric_type="staff" rubric_feedback_prompt="(Optional) What aspects of this response stood out to you? What did it do well? How could it improve?" rubric_feedback_default_text="I noticed that this response..." %}
```
cahrens: Same question about the strings here.

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
```diff
@@ -0,0 +1,52 @@
...
+                    </form>
+                </article>
+            </div>
+
+            <div class="step__actions">
+                <div class="message message--inline message--error message--error-server">
```
cahrens: Were you able to test this error condition (that the error properly displays)?

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
```diff
@@ -25,7 +25,7 @@ <h3 class="message__title">{% trans "Submission Cancelled" %}</h3>
 
                 <div class="message__content">
                   <p>
-                    {% blocktrans with removed_datetime=workflow_cancellation.created_at|utc|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
+                    {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|utc|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
```
cahrens: Please also remove the "Your response" HTML below, as noted in the TNL-3786.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,306 @@
...
+                    <p>
+                        {% blocktrans with comments=workflow_cancellation.comments %}
+                            Comments: {{ comments }}
+                        {% endblocktrans %}
+                    </p>
+                {% else %}
+                    {% include "openassessmentblock/oa_submission_answer.html" with answer=submission.answer answer_text_label="The learner's response to the question above:" %}
+                {% endif %}
+
+                {% if submission.file_url %}
```
cahrens: I wonder if the user-submitted file also should not be displayed if the user response was cancelled. @explorerleslie?

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,306 @@
...
+        <div class="staff-info__status ui-staff__content__section wrapper--ui--collapse">
+            <div class="ui-staff ui-toggle-visibility is--collapsed">
+                <h2 class="staff-info__title ui-staff__subtitle ui-toggle-visibility__control">
+                    <i class="icon fa fa-caret-right" aria-hidden="true"></i>
+                    <span>{% trans "Peer Assessments for This Learner" %}</span>
+                </h2>
+                <div class="ui-toggle-visibility__content">
+                    {% for assessment in peer_assessments %}
+                        {% with peer_num=forloop.counter %}
+                            <h4 class="title--sub"> {% trans "Peer" %} {{ peer_num }}: </h4>
```
cahrens: Not introduced by us, but this is certainly bad for 18n. It should be a blocktrans with placeholder.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,306 @@
...
+        <div class="staff-info__status ui-staff__content__section wrapper--ui--collapse">
+            <div class="ui-staff ui-toggle-visibility is--collapsed">
+                <h2 class="staff-info__title ui-staff__subtitle ui-toggle-visibility__control">
+                    <i class="icon fa fa-caret-right" aria-hidden="true"></i>
+                    <span>{% trans "Peer Assessments Completed by This Learner" %}</span>
+                </h2>
+                <div class="ui-toggle-visibility__content">
+                    {% for assessment in submitted_assessments %}
+                        {% with peer_num=forloop.counter %}
+                            <h4 class="title--sub">{% trans "Assessment" %} {{ peer_num }}:</h4>
```
cahrens: Same issue here with 18n.

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
cahrens: https://code.djangoproject.com/ticket/1067

Useful for being about to put commands on multiple lines.

openassessment/xblock/staff_area_mixin.py
```diff
@@ -214,16 +214,20 @@ def render_student_info(self, data, suffix=''):  # pylint: disable=W0613
...
-    def get_student_info_path_and_context(self, student_username):
+    def get_student_info_path_and_context(self, student_username, expanded_view=None):
         """
-        Get the proper path and context for rendering the the student info
-        section of the staff debug panel.
+        Get the proper path and context for rendering the student info
+        section of the staff area.
 
         Args:
             student_username (unicode): The username of the student to report.
```
cahrens: Add documentation about expanded_view.

openassessment/xblock/staff_area_mixin.py
```diff
@@ -278,18 +282,22 @@ def get_student_info_path_and_context(self, student_username):
         if "example-based-assessment" in assessment_steps:
             example_based_assessment = ai_api.get_latest_assessment(submission_uuid)
 
-        workflow_cancellation = workflow_api.get_assessment_workflow_cancellation(submission_uuid)
-        if workflow_cancellation:
-            workflow_cancellation['cancelled_by'] = self.get_username(workflow_cancellation['cancelled_by_id'])
+        workflow = self.get_workflow_info(submission_uuid=submission_uuid)
```
cahrens: There is no longer a need to add the 'cancelled_by" information?

openassessment/xblock/staff_area_mixin.py
```diff
@@ -377,8 +385,8 @@ def cancel_submission(self, data, suffix=''):
                 "success": True,
                 'msg': self._(
                     u"The learner submission has been removed from peer assessment. "
-                    u"The learner receives a grade of zero unless you reset "
-                    u"the learner's attempts for the problem to allow them to "
+                    u"The learner receives a grade of zero unless you delete "
```
cahrens: This makes me think that we need to do some rebasing.

openassessment/xblock/static/js/spec/lms/oa_staff_area.js
```diff
@@ -281,6 +247,101 @@ describe('OpenAssessment.StaffAreaView', function() {
...
+            it('can submit a staff grade override', function() {
+                var staffArea = createStaffArea(),
+                    $assessment, $submitButton;
+                chooseStudent(staffArea, 'testStudent');
+                $assessment = $('.wrapper--staff-assessment', staffArea.element)
+                $submitButton = $('.action--submit', $assessment);
+                fillAssessment($assessment);
+
+                // Submit the assessment
+                server.studentTemplate = 'oa_staff_cancelled_submission.html';
```
cahrens: This test is testing the wrong thing-- cancellation instead of staff override (see expects below as well!).

openassessment/xblock/staff_area_mixin.py
andy-armstrong: I moved all that logic into the new helper function ``get_workflow_cancellation_info`` that I use below. It allowed me to share the bug fix between the student and staff views.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -157,108 +198,148 @@
...
-                    $('#reschedule_unfinished_tasks_message', view.element).text(msg);
-                }
-            ).fail(function(errMsg) {
-                    $('#reschedule_unfinished_tasks_message', view.element).text(errMsg);
-                });
+            this.server.rescheduleUnfinishedTasks()
+                .done(function(msg) {
+                    $('.reschedule_unfinished_tasks_message', view.element).text(msg);
+                }).fail(function(errMsg) {
+                $('.reschedule_unfinished_tasks_message', view.element).text(errMsg);
```
cahrens: Nit: spacing.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -157,108 +198,148 @@
...
+            var comments = $('.cancel_submission_comments', this.element).val();
+            this.server.cancelSubmission(submissionUUID, comments)
+                .done(function(msg) {
                     $('.cancel-submission-error').html('');
-                    $('#openassessment__staff-info__cancel__submission', view.element).html(msg);
-                }
-            ).fail(function(errMsg) {
+                    view.loadStudentInfo({expanded_view: 'final-grade'})
+                        .done(function() {
+                            $('.openassessment__staff-info__cancel__submission', view.element).html(msg);
```
cahrens: I put a note about this above-- do we really need to show the msg here? Aren't we reloading and showing the message in the final grade section instead?

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -157,108 +198,148 @@
...
+
+        /**
+         * Enable/disable the staff assessment submit button.
+         *
+         * @param enabled If specified, sets the state of the button.
+         * @returns {boolean} Whether the button is enabled
+         */
+        staffSubmitEnabled: function(enabled) {
+            var button = $('.wrapper--staff-assessment .action--submit', this.element);
+            if (typeof enabled === 'undefined') {
```
cahrens: This is a confusing method. It does two different things depending on whether enabled is undefined-- it either returns the state of the button (which is what is documented), or it toggles the button state (and has no explicit return). At the very least, it should always return the state of the button (I would lean towards 2 different method).

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -157,108 +198,148 @@
...
+        submitStaffAssessment: function(submissionID, rubric) {
+            // Send the assessment to the server
+            var view = this;
+            var baseView = this.baseView;
+            baseView.toggleActionError('staff', null);
+            view.staffSubmitEnabled(false);
+
+            this.server.staffAssess(
+                rubric.optionsSelected(), rubric.criterionFeedback(), rubric.overallFeedback(), submissionID
+                )
```
cahrens: Nit: spacing.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -157,108 +198,148 @@
...
+                .done(function() {
+                    view.loadStudentInfo({expanded_view:  'final-grade'});
+                })
+                .fail(function(errorMessage) {
+                    baseView.toggleActionError('staff', errorMessage);
```
cahrens: Do we have test coverage of the failure case?

openassessment/xblock/static/js/src/oa_server.js
```diff
@@ -26,57 +22,45 @@ if (typeof OpenAssessment.Server === "undefined" || !OpenAssessment.Server) {
...
                 $.ajax({
-                    url: url,
-                    type: "POST",
-                    dataType: "html"
-                }).done(function(data) {
-                    defer.resolveWith(view, [data]);
-                }).fail(function() {
-                    defer.rejectWith(view, [gettext('This section could not be loaded.')]);
-                });
+                        url: url,
```
cahrens: Nit: I'm not sure why you changed the spacing in these blocks. The way it was before looks right to me.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
andy-armstrong: This is a direct copy of every other class's submit handling logic, but I agree that it is confusing. I'll look into breaking them all into two methods.

andy-armstrong: Note that it is a contract with oa_rubric.js which expects a single function that can do both things.

openassessment/xblock/static/js/src/oa_server.js
```diff
@@ -84,504 +68,463 @@ if (typeof OpenAssessment.Server === "undefined" || !OpenAssessment.Server) {
...
+         * @param payload The assessment payload
+         * @returns {promise} A promise which resolves with no arguments if successful,
+         *     and which fails with an error message otherwise.
+         */
+        submitAssessment: function(assessmentType, payload) {
+            var url = this.url(assessmentType);
+            return $.Deferred(function(defer) {
+                $.ajax({ type: "POST", url: url, data: JSON.stringify(payload), contentType: jsonContentType })
+                    .done(function(data) {
+                            if (data.success) {
```
cahrens: Nit: to me, this appears to be indented an extra level.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,306 @@
...
+                {% else %}
+                    {% include "openassessmentblock/oa_submission_answer.html" with answer=submission.answer answer_text_label="The learner's response to the question above:" %}
+                {% endif %}
+
+                {% if submission.file_url %}
+                    <a href="{{ submission.file_url }}" class="submission--file">
+                        {% trans "The file associated with this response." %}
+                    </a>
+                    <span>{%  trans "Caution: This file was uploaded by another course learner and has not been verified, screened, approved, reviewed or endorsed by edX. If you decide to access it, you do so at your own risk." %}</span>
+                {% endif %}
```
catong: This text comes from Legal so I'm not suggesting edits other than adding a serial comma after "reviewed": "...has not been verified, screened, approved, reviewed, or endorsed by edX."

openassessment/xblock/static/sass/oa/utilities/_developer.scss
```diff
@@ -129,33 +134,115 @@ $link-hover: $edx-blue-l1 !default; // from our Pattern Library http://ux.edx.or
...
     width: $max-width/2;
   }
 
-  /**
-   * The follow styles are bound for the "shame" file. This is done to override
-   * LMS specific styles on HTML elements.
-   */
+  // staff assessments
+  .wrapper--staff-assessment {
+    margin-top: ($baseline-v/2);
```
cahrens: Was all this staff-assessment CSS copied from the student view? If so, is there a way to share the class definitions so we don't need 2 copies (I'm worried about it getting out of sync)?

openassessment/xblock/staff_area_mixin.py
cahrens: :+1:

openassessment/xblock/submission_mixin.py
```diff
@@ -44,6 +44,7 @@ class SubmissionMixin(object):
         'msh1xml', 'msh2xml', 'action', 'apk', 'app', 'bin', 'command', 'csh',
         'ins', 'inx', 'ipa', 'isu', 'job', 'mst', 'osx', 'out', 'paf', 'prg',
         'rgs', 'run', 'sct', 'shb', 'shs', 'u3p', 'vbscript', 'vbe', 'workflow',
+        'htm', 'html',
```
cahrens: Does this indicate a rebase issue?

andy-armstrong: Yes. I had accidentally rebased this branch against master without rebasing ora-staff-grading first. The diff has gone away now.

test/acceptance/tests.py
```diff
@@ -285,6 +297,122 @@ def test_staff_area_panel(self, panel_name, button_label):
...
+        # Check the learner's current score.
+        self.staff_area_page.expand_learner_report_sections()
+        self.staff_area_page.verify_learner_final_score("Final grade: 6 out of 8")
+
+        # Do staff override and wait for final score to change.
+        self.staff_area_page.assess("staff", [0, 1])
+
+        # Verify that the new student score is different from the original one.
+        # TODO: uncomment this after hooked up to the API. Also verify other state if appropriate.
+        # self.staff_area_page.verify_learner_final_score("Final grade: 1 out of 8")
```
cahrens: I need to re-enable this code now.

cahrens:
I noticed that once a staff override is submitted, the peer response section in "staff info" is empty. I don't think that was intentional, as students can still see the peer responses. @andy-armstrong can you please look into this?

Also, note that once a staff assessment is submitted, it is impossible to see the staff score breakdown and comments. It is also very difficult to deduce that the Final Score came from a staff override. This should be improved by https://openedx.atlassian.net/browse/TNL-3761, but @explorerleslie, you should be thinking about how you want this to look for the MVP.

explorerleslie:
@andy-armstrong @cahrens hrm. We may want to consider adding an additional staff info section for staff grading that appears if there is a staff grade present, to match the pattern we use for peer and self, etc. Then the info is always present somewhere. (I didn't catch this before because I thought we would be loading up the same form every time and that's where you could see the staff grade.) What do you think?

cahrens:
@explorerleslie I think it would make sense to have a staff assessment section. I'll create a story, and we can groom it today. We will need to think about MVP though, as delivery date is starting to become an issue.

explorerleslie:
@cahrens thanks. This does not need to be part of the 12/15 deadline. If it were released mid-January that would be fine.

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
andy-armstrong: Good question. oa_submission_answer.html does call ``trans`` on the parameter, but I don't know whether the string will be scraped. This convention is used by all the assessments so they will all need to be fixed if not.

explorerleslie:
with my last comment, I'm :+1: on the interaction for this PR.

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
andy-armstrong: Yes, header does get translated by oa_uploaded_file.html, but again I'm not sure how or if these strings get scraped. I'm doing a translation update now to see if it is working.

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
andy-armstrong: There two were *not* translated so I've fixed the code.

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
andy-armstrong: Done

openassessment/templates/openassessmentblock/staff_area/oa_staff_assessment.html
```diff
@@ -0,0 +1,51 @@
+{% load i18n %}
+{% spaceless %}
+
+    {% block body %}
+    <div class="ui-toggle-visibility__content">
+        <div class="wrapper--staff-assessment">
+            <div class="step__instruction">
+                <p>{% trans "Allows you to override the current learner's grade using the problem's rubric." %}</p>
```
catong: @andy-armstrong Can we change this to "Override this learner's current grade using the problem's rubric."

andy-armstrong: Sure thing... done.

openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html
catong: @andy-armstrong I don't know if it can be addressed in this PR, but the display is not ideal for the username/email field - whatever text gets entered in the field is cut off at the bottom. 
ora_coursestafftools

openassessment/xblock/static/js/src/lms/oa_staff_area.js
andy-armstrong: No. :-) I've added one, and also added a test for the failure case for canceling a submission.

openassessment/xblock/static/js/src/oa_server.js
andy-armstrong: It was an over-aggressive clean up. There were a few lines where the ``.done`` appeared so late in the line that it didn't seem clear to me. However, the majority of these were fine as is so I've reverted them back.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
andy-armstrong: Done

openassessment/xblock/static/sass/oa/utilities/_developer.scss
andy-armstrong: It isn't duplicated as the other assessments are all styled as being top level steps. This styling is unique to showing inline assessments in the staff area.

andy-armstrong: I should have said that I have the same concern that the sass is getting out of control, and that there are a lot of changes in the _developer.scss. I don't know the best time to try to sort that out.

andy-armstrong:
@cahrens @dianakhuang @catong I've addressed your feedback in the last commit (https://github.com/edx/edx-ora2/commit/8596c4dcee18eac3ad75369b204ba74797c85278). Please re-review.

andy-armstrong:
@cahrens I haven't tackled the i18n questions yet, as updating the strings has already taken four hours and still hasn't completed. :-(

andy-armstrong:
@cahrens I've pushed up a tiny commit to remove the last vestiges of Studio collapsing. Let me know if you had other things you wanted to do here.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
cahrens: @explorerleslie or @catong  Do you know how to configure example based assessments? We'd like to make sure that the HTML didn't break, but @andy-armstrong (and I) don't really understand example based assessments.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
cahrens: @explorerleslie Do you have an opinion on this? It wasn't part of our work, but it would be easy to fix now.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,316 @@
...
+            <div class="ui-staff ui-toggle-visibility is--collapsed">
+                <h2 class="staff-info__title ui-staff__subtitle ui-toggle-visibility__control">
+                    <i class="icon fa fa-caret-right" aria-hidden="true"></i>
+                    <span>{% trans "Peer Assessments for This Learner" %}</span>
+                </h2>
+                <div class="ui-toggle-visibility__content">
+                    {% if peer_assessments %}
+                        {% for assessment in peer_assessments %}
+                            {% with peer_num=forloop.counter %}
+                                <h4 class="title--sub"> {% trans "Peer" %} {{ peer_num }}: </h4>
```
cahrens: The comment on this got lost--

Not introduced by us, but this is certainly bad for 18n. It should be a blocktrans with placeholder.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
explorerleslie: Huh. Yeah, if it's easy, we should fix this now, @cahrens. Thanks for flagging!

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,316 @@
...
+            <div class="ui-staff ui-toggle-visibility is--collapsed">
+                <h2 class="staff-info__title ui-staff__subtitle ui-toggle-visibility__control">
+                    <i class="icon fa fa-caret-right" aria-hidden="true"></i>
+                    <span>{% trans "Peer Assessments Completed by This Learner" %}</span>
+                </h2>
+                <div class="ui-toggle-visibility__content">
+                    {% if submitted_assessments %}
+                        {% for assessment in submitted_assessments %}
+                            {% with peer_num=forloop.counter %}
+                                <h4 class="title--sub">{% trans "Assessment" %} {{ peer_num }}:</h4>
```
cahrens: The comment on this got lost--

Not introduced by us, but this is certainly bad for 18n. It should be a blocktrans with placeholder.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
explorerleslie: Example based assessment is either AI grading or student training -- do you know which? 

If it's student training, the docs are here: http://edx.readthedocs.org/projects/edx-partner-course-staff/en/latest/exercises_tools/open_response_assessments/CreateORAAssignment.html#step-5-select-assignment-steps

If it's AI, don't test -- no one is using it right now, and it's a huge pain to set up...needs further work before it's ready for prime time.

openassessment/xblock/static/js/spec/lms/oa_staff_area.js
```diff
@@ -281,6 +255,137 @@ describe('OpenAssessment.StaffAreaView', function() {
...
+                fillAssessment($assessment);
+
+                // Submit the assessment
+                server.studentTemplate = 'oa_staff_graded_submission.html';
+                submitAssessment(staffArea);
+
+                // Verify that the student info is visible and shows the correct score
+                $gradeSection = $('.staff-info__student__grade', staffArea.element);
+                expect($('.ui-toggle-visibility', $gradeSection)).not.toHaveClass('is--collapsed');
+                expect($('p', $gradeSection).first().text().trim()).toBe(
```
cahrens: Can you also test the score before the override (up around line 357) so we verify that the score was different before?

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -157,108 +194,148 @@
...
+
+            this.server.staffAssess(
+                rubric.optionsSelected(), rubric.criterionFeedback(), rubric.overallFeedback(), submissionID
+            ).done(function() {
+                // Note: we ignore any message returned from the server and instead
+                // re-render the student info with the "Learner's Final Grade"
+                // section expanded. This section will show the learner's
+                // final grade and in the future should include details of
+                // the staff override itself.
+                view.loadStudentInfo({expanded_view:  'final-grade'});
```
cahrens: Nit: extra space on line 334.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,316 @@
...
+                                    </tbody>
+                                </table>
+                                <h4 class="title--sub">{% trans "Overall Feedback" %}</h4>
+                                <div class="student__answer__display__content">
+                                    {{ assessment.feedback|linebreaks }}
+                                </div>
+                            {% endwith %}
+                        {%  endfor %}
+                    {% else %}
+                        <p>{% trans "This learner currently has no peer assessments." %}</p>
```
cahrens: @andy-armstrong I like these changes. But you should flag it for @catong and @explorerleslie so they are also aware.

openassessment/xblock/static/js/fixtures/templates.json
```diff
@@ -682,8 +684,52 @@
         "output": "oa_staff_area.html"
     },
     {
-        "template": "openassessmentblock/staff_area/student_info.html",
+        "template": "openassessmentblock/staff_area/oa_student_info.html",
```
cahrens: I'm confused-- there is another block below for oa_student_info.html (line 761). Are there supposed to be 2? If so, how does this work?

cahrens:
@andy-armstrong I reviewed your code review commit and just had a few notes. Looking great! These are the remaining things that I am aware of:

1) I think we need @frrrances to give a final CSS review (though she is not currently tagged on the PR).
2) I need to retest the issue I thought I had found about peer reviews disappearing once the sandbox is updated.
3) We need to figure out how to test the training scenario.
4) Remove the unnecessary CSS related to the expand/collapse code removal.

openassessment/templates/openassessmentblock/oa_rubric.html
```diff
@@ -55,15 +56,16 @@ <h4 class="question__title ui-toggle-visibility__control">
             </div>
         </li>
         {% endfor %}
-        <li class="wrapper--input field field--textarea assessment__rubric__question assessment__rubric__question--feedback" id="assessment__rubric__question--feedback">
-            <label class="question__title" for="assessment__rubric__question--feedback__value">
-                <span class="question__title__copy">{{ rubric_feedback_prompt }}</span>
+        <li class="wrapper--input field field--textarea assessment__rubric__question assessment__rubric__question--feedback">
+            <label class="question__title" for="{{ rubric_type }}__assessment__rubric__question--feedback__value">
+                <span class="question__title__copy">{% trans rubric_feedback_prompt %}</span>
```
dianakhuang: Do we know that this will send the right string to `gettext` for translation (i.e. are there other places we do it in ora2)? I'm not super familiar with how it works for ora2.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,316 @@
...
+                        {% endblocktrans %}
+                    </p>
+                {% else %}
+                    {% include "openassessmentblock/oa_submission_answer.html" with answer=submission.answer answer_text_label="The learner's response to the question above:" %}
+                {% endif %}
+
+                {% if submission.file_url %}
+                    <a href="{{ submission.file_url }}" class="submission--file">
+                        {% trans "The file associated with this response." %}
+                    </a>
```
cahrens: @andy-armstrong @explorerleslie OK'd moving this section up into the else clause so it isn't shown if the submission was cancelled.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -124,29 +122,32 @@ OpenAssessment.BaseView.prototype = {
...
+     * @param {string} errorMessage An optional error message to use instead of the default.
+     */
+    showLoadError: function(stepName, errorMessage) {
+        if (!errorMessage) {
+            errorMessage = gettext('Unable to load');
+        }
+        var $container = $('#openassessment__' + stepName);
+        $container.toggleClass('has--error', true);
+        $container.find('.step__status__value i').removeClass().addClass('icon fa fa-exclamation-triangle');
+        $container.find('.step__status__value .copy').html(errorMessage);
```
robrap: @andy-armstrong Can our error messages contain html?  I am guessing this needs to be escaped.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -124,29 +122,32 @@ OpenAssessment.BaseView.prototype = {
...
         if (container === null) {
-            if (msg !== null) { console.log(msg); }
+            if (message !== null) { console.log(message); }
         }
 
         else {
             // Insert the error message
-            var msgHtml = (msg === null) ? "" : msg;
+            var msgHtml = (message === null) ? "" : message;
             $(container + " .message__content", element).html('<p>' + msgHtml + '</p>');
```
robrap: @andy-armstrong Is msgHtml really html or is it plain text?  I imagine it needs to be escaped and should not be named msgHtml, but not sure.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -157,108 +194,148 @@
...
-                });
+            var comments = $('.cancel_submission_comments', this.element).val();
+            this.server.cancelSubmission(submissionUUID, comments).done(function() {
+                // Note: we ignore any message returned from the server and instead
+                // re-render the student info with the "Learner's Final Grade"
+                // section expanded. This section will show that the learner's
+                // submission was cancelled.
+                view.loadStudentInfo({expanded_view: 'final-grade'});
+            }).fail(function(errorMessage) {
+                $('.cancel-submission-error').html(errorMessage);
```
robrap: @andy-armstrong Escaping again.

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
```diff
@@ -24,26 +24,17 @@
...
-                <div class="step__content">
-                    <article class="submission__answer__display">
-                        <h3 class="submission__answer__display__title">{% trans "Your Response" %}</h3>
-
-                        <div class="submission__answer__display__content">
-                            {{ student_submission.answer.text|linebreaks }}
-                        </div>
-                    </article>
+                    <p>
+                        {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|utc|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
```
robrap: @andy-armstrong Not sure if this is going to require a separate story (or stories), but working on XSS presentation with Ned, we realized a couple of things:
1. Where we use trans and blocktrans we don't seem to be escaping the translation, so we need to establish a best practice.
2. The [i18n wiki page](http://edx.readthedocs.org/projects/edx-developer-guide/en/latest/internationalization/i18n.html) doesn't include Django templates
3. The way we are handling "blocktrans" doesn't seem like it would work right for translations, and also might introduce double escaping.  Either way, the best practice should be documented in the wiki when it is established.
Thanks!

openassessment/xblock/workflow_mixin.py
```diff
@@ -110,21 +112,27 @@ def update_workflow_status(self, submission_uuid=None):
...
+
         Returns:
             dict
 
         Raises:
             AssessmentWorkflowError
         """
-        if not self.submission_uuid:
-            return {}
+        if not submission_uuid:
```
dianakhuang: I know that this isn't your code originally, but I'd probably recommend checking for `is not None` explicitly here.

dianakhuang:
Those were my only real comments. It's looking great, and I'm super happy to see the cleanup going on.

andy-armstrong:
@frrrances This PR is almost ready to land so it would be great to get your review on it. Unfortunately we're having issues with the sandbox so I'll let you know when that is back up.

andy-armstrong:
@frrrances @cahrens @dianakhuang The sandbox is back in action:

https://andy-armstrong.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/4dbd99f691064633a9be0daadb805ce6/

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,316 @@
...
+                                        <th abbr="Criterion" scope="col">{% trans "Criterion" %}</th>
+                                        <th abbr="Selected Option" scope="col">{% trans "Selected Option" %}</th>
+                                        <th abbr="Points" scope="col">{% trans "Points" %}</th>
+                                        <th abbr="Points Possible" scope="col">{% trans "Points Possible" %}</th>
+                                    </tr>
+                                    </thead>
+
+                                    <tbody>
+                                    {% for criterion in rubric_criteria %}
+                                        {% for part in self_assessment.parts %}
```
efischer19: Just spitballing, but shouldn't this be `for part in assessment.parts` to match `for assessment in peer_assessments` above? Could this be the cause of the mysterious "self as peer" bug we've been chasing?

cahrens: Ah, definitely! I looked that over myself, but completely missed it. Thanks, @efischer19!

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
andy-armstrong: Done

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
andy-armstrong: Fixed

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
andy-armstrong: Fixed

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
andy-armstrong: @catong @explorerleslie I modified this section so that it always displays if the problem is configured to support peer assessments. Before this change, it meant that the section would not be shown, so there was no way for staff to tell that this problem supports peer assessments but just happens not to have any.

@catong Let me know if you think the language for this is clear.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -0,0 +1,318 @@
...
+                                    </tbody>
+                                </table>
+                                <h4 class="title--sub">{% trans "Overall Feedback" %}</h4>
+                                <div class="student__answer__display__content">
+                                    {{ assessment.feedback|linebreaks }}
+                                </div>
+                            {% endwith %}
+                        {%  endfor %}
+                    {% else %}
+                        <p>{% trans "This learner has not completed any peer assessments." %}</p>
```
andy-armstrong: @catong @explorerleslie This is another example where I show a section with  'no assessments' message, rather than hiding the section entirely.

openassessment/xblock/workflow_mixin.py
andy-armstrong: Done

openassessment/templates/openassessmentblock/oa_rubric.html
andy-armstrong: I'm not 100% sure about this, but this is the idiom used elsewhere in this file and others. The question to my mind is whether the string being passed as a parameter to this template will be processed by the string scraper. My intuition says it wouldn't be, but I don't know for certain yet. I think we probably need an i18n clean up story for ORA before we try to land the feature.

openassessment/xblock/static/js/fixtures/templates.json
andy-armstrong: The way it works is that you can pass more than one set of parameters through the same template. The import distinction is the "output" which is the name of the resulting HTML for the template with the given context. This allows us to have many variations of student info for the different situations that we want to test.

openassessment/xblock/static/js/spec/lms/oa_staff_area.js
andy-armstrong: Done

openassessment/xblock/static/js/src/lms/oa_base.js
andy-armstrong: Good catch. There's a lot of similar code in here that probably needs to be rethought for security reasons. I'm thinking we should create another story to tackle it all.

For this particular change, I don't think I ended up using it. I'm going to revert this change.

andy-armstrong: Strangely there are some callers that pass a message as a second parameter, even though I only introduced that as an option with this change. Given that, I've chosen to just escape the string.

openassessment/xblock/static/js/src/lms/oa_base.js
andy-armstrong: I've removed the confusingly named variable, and I'm now escaping the string.

andy-armstrong:
@cahrens @dianakhuang @robrap @frrrances @catong @explorerleslie I've implemented as much as I'm going to tonight, and I've updated the sandbox with the latest code.

To @cahrens's points:
1) I've tagged @frrrances to give a final CSS review (as much as she has time for)
2) I think we've fixed the issue with peer assessments disappearing, because it was accidentally rendering self assessments.
3) We still need to figure out how to test the example-based assessments (but not training as it isn't truly supported yet)
4) I haven't looked at removing the unnecessary CSS related to the expand/collapse code removal. Note that we do still use expand/collapse in the XBlock so we'll have to be careful to only excise what is truly no longer used.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
catong: @andy-armstrong I'd like to suggest a minor tweak: "This learner has not received any peer assessments."

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
catong: @andy-armstrong I'd also like to suggest a minor tweak here: "This learner has not performed any peer assessments."

cahrens:
@andy-armstrong the CSS I'm proposing to remove in _developer.css is strictly scoped to the HTML that you already edited. For example, 

````
.openassessment_training_example.collapsed {
    .openassessment_training_example_body {
      display: none;
    }
    .openassessment_training_example_header {
      .action.expand-collapse {
        @include float(left);
        .ui-toggle-expansion {
          @include transform(rotate(bidi-rotate-angle(-90deg)));
          @include transform-origin(50% 50%);
          .icon-caret-down:before {
            color: #009fe6;
          }
          .icon-caret-down:before:hover {
            color: $edx-gray-d3;
            cursor: pointer;
          }
        }
      }
    }
  }
````
If you aren't able to get to this, I can do it later today.

cahrens:
@andy-armstrong I removed the dead CSS and talked with @frrrances. She would prefer to review the CSS closer to the end. I'll add a task to our "cleanup" story.

openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html
cahrens: I was at least able to press these buttons and verify that the appropriate confirmation messages show.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
cahrens: @andy-armstrong I was able to test this-- looks fine.

andy-armstrong:
:+1: for all of your commits, @cahrens. Awesome acceptance and a11y tests, as always!

cahrens:
:+1: After you fix the messages we've been talking about (and squash away). With fast build times, perhaps this can be merged by 2?

PR #763, opened 2015-11-09
add custom a11y rules in accessibility tests

clytwynec:


clytwynec:
@benpatterson @jzoldak  Please review.

This adds the custom a11y rules to the ORA2 bok choy builds. See http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/470/

scripts/test-acceptance.sh
```diff
@@ -33,5 +33,15 @@ cd test/logs
 
 test_name="${1:-acceptance}"
 
+if [[ "${test_name}" = "accessibility" ]]; then
```
clytwynec: The machines that these tests run on in jenkins have an old version of npm which doesn't seem to work with the requirements in this repo.  There is a devops ticket for this, but it may be better to wait until some of the other work they have going on now is done before attempting to install the custom rules via npm.      

This bit of code is a work around for that.  It will work both in jenkins and for local runs of the tests.

benpatterson:
:+1:

scripts/test-acceptance.sh
jzoldak: Please include this information in a TODO comment in the file.

jzoldak:
:+1:

PR #765, opened 2015-11-09
fixing acceptance tests

clytwynec:


clytwynec:
@benpatterson @andy-armstrong @cahrens  It looks like the acceptance tests have been silently failing since PR
https://github.com/edx/edx-ora2/pull/743.  (It shows a blue dot, but if you look at the console log, the tests actually failed.)  This corrects:
* The exit status by storing it as a variable in the script  (This was something I missed when adding the a11y tests originally)
* The failures caused by the PhantomJS / SauceLabs combo (the "No json object could be decoded" errors) by not using PhantomJS anymore.

There is now a test that is actually failing, but it doesn't fail locally. (See [this build](http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/454/console))  It will need more investigation, but I'd like to get the fix from this PR in ASAP.

benpatterson:
:+1: from me

scripts/test-acceptance.sh
```diff
@@ -30,10 +30,8 @@ fi
 
 mkdir -p test/logs
 cd test/logs
-export SELENIUM_BROWSER=phantomjs
```
andy-armstrong: I liked that the tests don't pop a browser when run locally. However, I know that @cahrens preferred it to show Chrome so that she could see what it was doing. Maybe this needs to be a setting, or maybe folks can just set SELENIUM_BROWSER themselves locally.

clytwynec: Yea,  You should still be able to choose locally by setting SELENIUM_BROWSER.

andy-armstrong:
Thanks for jumping on this @clytwynec. I figured it was a sandbox issue so I'm glad that that wasn't the case. :+1: once we decide how to handle the SELENIUM_BROWSER for developers versus Travis.

PR #766, opened 2015-11-10
Remove django-celery

muhammad-ammar:


muhammad-ammar:
@feanil @symbolist @nedbat @doctoryes @muzaffaryousaf @macdiesel Please review. Thanks!

doctoryes:
I don't know much about ORA2. But looks like we're using Celery 3.1 now, which supports Django:
http://docs.celeryproject.org/en/latest/whatsnew-3.1.html#django-supported-out-of-the-box
And tests pass - and @muhammad-ammar knows what he's doing!
So, :+1:

muhammad-ammar:
@doctoryes acceptance tests not run on travis. Currently my local ora2 test environment is not working. I will try to fix the env tomorrow and then picture will be more clear. Other than that i @feanil can tell if these changes are good.

feanil:
Changes look good to me.

muhammad-ammar:
Acceptance tests are passing. http://jenkins.edx.org:8080/job/ora2-acceptance-tests/468/console

openassessment/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+Celery needs to be loaded so that task registration and discovery can work correctly.
+"""
+# This will make sure the app is always imported when Django starts so that shared_task will use this app.
+from .celery import APP as CELERY_APP
```
symbolist: Don't we want to import this only in `settings.test`? When run as part of edxapp, the settings defined in edx-platform should apply.

openassessment/celery.py
```diff
@@ -0,0 +1,23 @@
...
+from __future__ import absolute_import
+
+import os
+
+from celery import Celery
+
+from django.conf import settings
+
+# set the default Django settings module for the 'celery' program.
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'proj.settings')
```
symbolist: If we run via `manage.py` this will already be set. No need to override it. Besides `proj.settings` is not a valid settings file.

symbolist: A better place for this file will be in the `settings` directory.

symbolist: And then it can be imported in `test.py`.

muhammad-ammar:
@symbolist Addressed you feedback.

muhammad-ammar:
@symbolist Please have a final look. Thanks!

symbolist:
Looks great now! :+1:

PR #769, opened 2015-11-12
Fix message wording about how student can resubmit problem.

cahrens:
https://openedx.atlassian.net/browse/TNL-3773

@catong and @andy-armstrong please review

@catong here is how it looks now:

image


cahrens:
Note that the we show an "old mongo" style location in the instructor dashboard. We should consider fixing that as well (though it is in edx-platform, not specific to ORA).

image

With "split", this looks more like "block-v1:ora+ora+ora+type@openassessment+block@6e78fa380b33424586d19ae858ae1be0"

andy-armstrong:
:+1:

catong:
:+1:

PR #768, opened 2015-11-12
Staff assessment api test additions

efischer19:
Increasing test coverage. Making this a separate PR so as not to interfere with ongoing review. I won't be changing behavior in any way, just writing some tests based on the following:

| Name | Stmts | Miss | Branch | Branch Misses | Cover | Missing |
|---|---|---|---|---|---|---|
| staff_assessment_mixin | 56 | 17 | 8 | 0 | 73% | 81-89, 99-112 |
| api.staff | 60 | ~~21~~ 0 | 6 | ~~1~~ 0 | ~~67~~ ~~70~~ ~~82~~ 100% | ~~58~~, ~~120-126~~, ~~157-160~~, ~~223-~~~~231-236~~ |
| workflow.models | 257 | 19 | 98 | 15 | 90% | 247-248, 357-361, 442, 474-475, 517-521, 544-549, 711, 718 |

Attack plan:
* ~~staff api~~:
  * ~~add a staff grade required test~~ or two
  * ~~add invalid rubric and invalid rubric selection tests~~
  * ~~mock out some database error~~
    * ~~get latest staff assessment~~
    * ~~get scores by criteria~~
    * ~~create_assessment~~
* staff mixin:
  * leave for later front-end tests
* workflow models:
  * get_score and skip over non-existent staff assessment
  * staff_score_exists True, somehow
  * something about get_score not getting a staff score #248
  * backwards-compatible mock None #357-361
  * non related, but could get us to 100% coverage:
    * cancel workflow that already has score
    * databaseerrors in workflow cancellation, get by uuid
    * get nonexistent workflow by uuid
    * unicode methods of cancellation object

PR #772, opened 2015-11-12
Staff grading test coverage (on hold)

efischer19:
Continuing from https://github.com/edx/edx-ora2/pull/768, with a goal of increasing test coverage. I won't be changing behavior in any way, just writing some tests based on the following:

| Name | Stmts | Miss | Branch | Branch Misses | Cover | Missing |
|---|---|---|---|---|---|---|
| workflow.models | 257 | 19 | 98 | 14 | 91% | 247-248, 357-361, 444, 476-477, 519-523, 546-551, 713, 720 |

Attack plan:
* get_score and skip over non-existent staff assessment
* staff_score_exists True, somehow
* something about get_score not getting a staff score #248
* backwards-compatible mock None #357-361
* non related, but could get us to 100% coverage:
  * cancel workflow that already has score
  * databaseerrors in workflow cancellation, get by uuid
  * get nonexistent workflow by uuid
  * unicode methods of cancellation object


efischer19:
This PR is not under active development currently, it may be in the future.

PR #773, opened 2015-11-13
Adding html to black list for custom file types.

muzaffaryousaf:
This PR adds `html & htm` extension to the black list for custom type file uploads.

Sandbox: https://html-blacklisted.sandbox.edx.org/

PR #774, opened 2015-11-13
Adding html to black list.

symbolist:
TNL-3767


symbolist:
Reviewed in https://github.com/edx/edx-ora2/pull/773.

PR #778, opened 2015-11-17
Provisionally Done work

efischer19:
Intended to address the issue of a staff grade's presence interacting with peer assessment requirements when determining whether a submission is complete.

FYI @dianakhuang

PR #779, opened 2015-11-18
Defining done in context of staff assessments

efischer19:
Supersedes https://github.com/edx/edx-ora2/pull/778

Currently under ~~active development~~ review

### DIFF
--- a/openassessment/assessment/test/test_staff.py
+++ b/openassessment/assessment/test/test_staff.py
@@ -78,8 +78,10 @@ class TestStaffAssessment(CacheResetTest):
         self.assertEqual(assessment["points_earned"], OPTIONS_SELECTED_DICT[key]["expected_points"])
         self.assertEqual(assessment["points_possible"], RUBRIC_POSSIBLE_POINTS)
 
-        # ensure submission is marked as finished
+        # Ensure submission and workflow are marked as finished
         self.assertTrue(staff_api.assessment_is_finished(tim_sub["uuid"], self.STEP_REQUIREMENTS))
+        workflow = workflow_api.get_workflow_for_submission(tim_sub["uuid"], self.STEP_REQUIREMENTS)
+        self.assertEqual(workflow["status"], "done")
 
     @data(*ASSESSMENT_SCORES_DDT)
     def test_create_assessment_required(self, key):
@@ -104,6 +106,8 @@ class TestStaffAssessment(CacheResetTest):
         # Verify assesment made, score updated, and no longer waiting
         self.assertEqual(staff_assessment["points_earned"], OPTIONS_SELECTED_DICT[key]["expected_points"])
         self.assertTrue(staff_api.assessment_is_finished(tim_sub["uuid"], self.STEP_REQUIREMENTS_WITH_STAFF))
+        workflow = workflow_api.get_workflow_for_submission(tim_sub["uuid"], self.STEP_REQUIREMENTS_WITH_STAFF)
+        self.assertEqual(workflow["status"], "done")
 
     @data(*ASSESSMENT_SCORES_DDT)
     def test_create_assessment_score_overrides(self, key):
@@ -182,7 +186,7 @@ class TestStaffAssessment(CacheResetTest):
 
         # Verify both assessment and workflow report correct score
         self.assertEqual(staff_assessment["points_earned"], OPTIONS_SELECTED_DICT[staff_score]["expected_points"])
-        workflow = workflow_api.get_workflow_for_submission(tim_sub["uuid"], self.STEP_REQUIREMENTS)
+        workflow = workflow_api.get_workflow_for_submission(tim_sub["uuid"], requirements)
         self.assertEqual(workflow["score"]["points_earned"], OPTIONS_SELECTED_DICT[staff_score]["expected_points"])
 
     @data(*ASSESSMENT_TYPES_DDT)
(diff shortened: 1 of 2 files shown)


efischer19:
@andy-armstrong @dianakhuang This is ready for review. Andy - the "is this workflow done?" checks in `test_staff.py` that you were asking previously about have been included in this PR.

FYI @cahrens @robrap @scottrish

cahrens:
Does in include fixing the bug where a staff override prevents the learner from subsequently completing peer assessments?

efischer19:
I think it should, we're doing things much more smartly now. I'll write a quick test to confirm that fact.

cahrens:
Yes, a test would be great-- thanks!

openassessment/assessment/test/test_staff.py
```diff
@@ -210,23 +218,68 @@ def test_create_assessment_does_not_block(self, after_type, after_assess):
...
-        requirements = self.STEP_REQUIREMENTS
-        if after_type == 'peer':
-            requirements = {"peer": {"must_grade": 0, "must_be_graded_by": 1}}
 
         # Verify both assessment and workflow report correct score (workflow should report previous value)
         self.assertEqual(assessment["points_earned"], unscored_assessment["expected_points"])
         workflow = workflow_api.get_workflow_for_submission(tim_sub["uuid"], requirements)
         self.assertEqual(workflow["score"]["points_earned"], OPTIONS_SELECTED_DICT[staff_score]["expected_points"])
 
+    def test_provisionally_done(self):
```
efischer19: Actually nevermind @cahrens - this particular test already includes a user making a peer assessment after having a staff assessment present. So yes, the bug should be fixed by this PR.

openassessment/workflow/models.py
```diff
@@ -348,7 +360,8 @@ def update_from_assessments(self, assessment_requirements):
             score = self.get_score(assessment_requirements, step_for_name)
             # If we found a score, then we're done
             if score is not None:
-                self.set_score(score)
+                if score.get("staff_id", None) is None:
```
dianakhuang: I don't think you could just write this as `score.get("staff_id") is None`.

efischer19: I'm not grok-ing this properly - do you mean I should write this as you've described?

openassessment/assessment/test/test_staff.py
```diff
@@ -196,6 +200,10 @@ def test_create_assessment_does_not_block(self, after_type, after_assess):
         if after_type == 'staff':
             return
 
+        requirements = self.STEP_REQUIREMENTS
```
cahrens: @efischer19 I'm having trouble understanding this test change-- can you explain?

openassessment/workflow/models.py
```diff
@@ -82,6 +82,9 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
         DEFAULT_ASSESSMENT_SCORE_PRIORITY
     )
 
+    # Blocking steps are steps that must be completed by the submitter, even if a staff assesment is present
```
cahrens: Nit: typo-- "assesment".

openassessment/workflow/models.py
```diff
@@ -82,6 +82,9 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
...
+    BLOCKING_STEPS = ["peer"]
```
cahrens: Is peer the only blocking step? What about if a self review is required?

openassessment/assessment/test/test_staff.py
efischer19: These are the same requirements we were already using in the second workflow update below. Previously, we had switched the requirements from `self.STEP_REQUIREMENTS` (which is `{}`) on the first check to the actual requirements on the second check.

Modifying the requirements mid-workflow seemed to be illogical, and was causing a weird issue with done status at one point during development, so I fixed the test to use the same requirements throughout.

openassessment/workflow/models.py
```diff
@@ -316,17 +322,23 @@ def update_from_assessments(self, assessment_requirements):
 
         new_staff_score = self.get_score(assessment_requirements, {'staff': step_for_name.get('staff', None)})
         if new_staff_score:
-            old_score = self.score
+            old_score = sub_api.get_latest_score_for_submission(self.submission_uuid)
             if not old_score or old_score['points_earned'] != new_staff_score['points_earned']:
```
cahrens: Is it possible that old_score['points_earned'] == new_staff_score['points_earned'], but old_score was not a staff score? If so, don't we still want to go through this block even though the numerical grade is the same?

(I wasn't part of the original review, so sorry if I'm completely confused here).

openassessment/workflow/models.py
```diff
@@ -357,7 +369,8 @@ def update_from_assessments(self, assessment_requirements):
             score = self.get_score(assessment_requirements, step_for_name)
             # If we found a score, then we're done
             if score is not None:
-                self.set_score(score)
+                if score.get("staff_id") is None:
```
cahrens: I'm not following this-- a comment would be helpful.

openassessment/workflow/models.py
```diff
@@ -316,17 +322,23 @@ def update_from_assessments(self, assessment_requirements):
...
                 ).format(uuid=self.submission_uuid))
-                staff_step = step_for_name.get('staff')
-                staff_step.assessment_completed_at=now()
-                staff_step.save()
-                self.status = self.STATUS.done
+
+                # Update the assessment and submitter_completed_at fields for all steps
+                # All steps are considered "assessment complete", as the staff score will override all
+                # Steps in BLOCKING_STEPS may still require the submitter to fulfill their obligations
+                for step in steps:
```
cahrens: Honestly, this part is also confusing me, but talking over it in person tomorrow with the new version will likely clear it up for me.

PR #780, opened 2015-11-18
self.STATUS AttributeError

efischer19:
~~This was previously trying to catch the wrong exception type.~~

Nope, had to do with where I check and update `self.STATUS`. In a prior version, I had explicitly added `'staff'` to the class globals in the default values. However, as @andy-armstrong noted in the review process, we were banking on those values never being overridden.

To ensure that `'staff'` was always present, I made it so that we check for it in `start_workflow()`, which I (erroneously) thought was always called when getting a workflow.

Turns out, you can also get an `AssessmentWorkflow` object from the database, in which case the `'staff'` checks were never made, meaning `self.STATUS` never got updated to include `staff`, leading to the `AttributeError`s seen in sandboxes and devstacks.

### DIFF
--- a/openassessment/workflow/models.py
+++ b/openassessment/workflow/models.py
@@ -97,6 +97,25 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
         # TODO: In migration, need a non-unique index on (course_id, item_id, status)
 
     @classmethod
+    def include_staff_in_class(cls):
+        if 'staff' not in cls.STEPS:
+            new_list = ['staff']
+            new_list.extend(cls.STEPS)
+            cls.STEPS = new_list
+            cls.STATUS_VALUES = cls.STEPS + cls.STATUSES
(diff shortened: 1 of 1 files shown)


efischer19:
Currently working on writing a test to skip `start_workflow()`, to ensure the validity of this proposed fix. All the current tests pass through that method, which is why we never caught this before.

efischer19:
@dianakhuang @andy-armstrong @cahrens - Here's what I think is the cause of the `AttributeError`s that have been showing up.

andy-armstrong:
:+1: Thanks for tracking this down, @efischer19. Can we merge it without a test just to get the feature branch unlocked? Or are you close to having a test?

dianakhuang:
:+1: from me.

efischer19:
I'm not super close to getting a test written (creating a workflow without actually calling `start_workflow` is tricky), but I'd be more than happy to merge without if we can confirm this actually fixes things.

PR #781, opened 2015-11-18
Speed up the translation process

andy-armstrong:
This is a simple fix to speed up the translation process. After talking to Sarina, we shouldn't be generating strings for every language, because all we upload is the English versions. I've also added a few more directories to the list of those that are ignored.

@cahrens @sarina Please review.

### DIFF
--- a/scripts/i18n-push.sh
+++ b/scripts/i18n-push.sh
@@ -21,14 +21,20 @@
 
 cd `dirname $BASH_SOURCE` && cd ..
 
+# Note we only extract the English language strings so they can be
+# sent to Transifex. All other language strings will be pulled
+# down from Transifex and so don't need to be generated.
+
 echo "Extracting i18n strings..."
-python manage.py makemessages --all
-python manage.py makemessages --all -d djangojs
+django-admin.py makemessages --locale=en --ignore="edx-ora2/*" --ignore="build/*" --ignore="node_modules/*"
+
+echo "Extracting client-side i18n strings..."
+django-admin.py makemessages --locale=en --ignore="edx-ora2/*" --ignore="build/*" --ignore="node_modules/*" -d djangojs
 
 echo "Generating dummy strings..."
 i18n_tool dummy
 
-read -p "Push strings to transifex? [y/n]  " RESP
+read -p "Push strings to Transifex? [y/n]  " RESP
 if [ "$RESP" = "y" ]; then
     i18n_tool transifex push
     echo " == Pushed strings to Transifex"


scripts/i18n-push.sh
```diff
@@ -21,9 +21,12 @@
...
 cd `dirname $BASH_SOURCE` && cd ..
 
+# Note we only extract the English language strings so they can be
+# sent to Transifex. All other language strings will be pulled
+# down from Transifex and so don't need to be generated.
 echo "Extracting i18n strings..."
-python manage.py makemessages --all
-python manage.py makemessages --all -d djangojs
+python manage.py makemessages -l en --all
+python manage.py makemessages -l en --all -d djangojs
```
sarina: Why are you only extracting and pushing JS strings? There aren't strings in templates or Python files? https://docs.djangoproject.com/en/1.4/ref/django-admin/#django-admin-option---domain

openassessment/locale/config.yaml
```diff
@@ -74,11 +74,19 @@ dummy_locales:
 
 # Directories we don't search for strings.
 ignore_dirs:
```
sarina: I don't think this configuration does anything if you're not using `i18n_tool extract`. If you're using just makemessages you need to explicitly pass `--ignore` option. https://docs.djangoproject.com/en/1.4/ref/django-admin/#django-admin-option---ignore

sarina: See https://github.com/edx/i18n-tools/blob/master/i18n/extract.py#L100 for how `i18n_tool` explicitly uses the `ignore_dirs`

scripts/i18n-push.sh
andy-armstrong: The default domain for makemessages is 'django' which covers *.py, *.html and *.txt. So the line before this one captures everything but the JS strings.

openassessment/locale/config.yaml
andy-armstrong: That's very strange. I wonder what this configuration is there for then! It makes me think that we should probably switch to i18n-tool instead of trying to emulate it...

sarina: Yeah I might see if you can watch the output on this branch and see if it's actually excluding these directories.

scripts/i18n-push.sh
sarina: Ah ok :+1:

andy-armstrong:
@sarina Here's a revised version that follows your advice, and that also switches to use django-admin as the old code didn't work since rebasing on 1.8. Is this more what you had in mind? I used this to update the translations this evening and it seemed to work as desired.

sarina:
Yes, this looks better. You may want to migrate this repo to using i18n_tools explicitly, so that all repos are using the same tool; that said, this seems fine.

screen shot 2015-11-24 at 8 37 46 am

:+1:

andy-armstrong:
Thanks @sarina. 

FYI, I did try updating to using i18n-tool, but I couldn't work out the correct way to use it. It would leave partial files in place of the actual ones, and then the push process would fail. For now I think what I have works okay.

@cahrens This is a very simple change, so please review it when you have a chance.

sarina:
Ah ok, well if you want to tackle that we can. I think you just have to set
the segmenting rules appropriately.

cahrens:
:+1:

PR #783, opened 2015-11-18
What is done?

efischer19:
Take 2


openassessment/workflow/models.py
```diff
@@ -316,17 +316,20 @@ def update_from_assessments(self, assessment_requirements):
 
         new_staff_score = self.get_score(assessment_requirements, {'staff': step_for_name.get('staff', None)})
         if new_staff_score:
-            old_score = self.score
-            if not old_score or old_score['points_earned'] != new_staff_score['points_earned']:
+            old_score = sub_api.get_latest_score_for_submission(self.submission_uuid)
+            if not old_score or old_score['created_at'] != new_staff_score['created_at']:
```
efischer19: Scores don't have a UID, but `created_at` should be a good enough approximation to determine whether we need to update.

openassessment/workflow/models.py
```diff
@@ -316,17 +316,20 @@ def update_from_assessments(self, assessment_requirements):
...
                     u"Workflow for submission UUID {uuid} has updated score using staff assessment."
                 ).format(uuid=self.submission_uuid))
-                staff_step = step_for_name.get('staff')
-                staff_step.assessment_completed_at=now()
-                staff_step.save()
-                self.status = self.STATUS.done
+
+                # Update the assessment_completed_at field for all steps
+                # All steps are considered "assessment complete", as the staff score will override all
+                for step in steps:
```
efischer19: By setting `assessment_complete` to a value, each step will individually report that it is done, so the status is being properly updated when a staff grade is present.

openassessment/workflow/models.py
```diff
@@ -357,7 +360,9 @@ def update_from_assessments(self, assessment_requirements):
             score = self.get_score(assessment_requirements, step_for_name)
             # If we found a score, then we're done
             if score is not None:
-                self.set_score(score)
+                # Only set the score if it's not a staff score, in which case it will have already been set above
+                if score.get("staff_id") is None:
```
efischer19: This is to prevent setting a staff score twice, since it was already checked/set above, before the `if status is done` line.

efischer19:
@cahrens This PR is the previous commit cherry-picked, with review feedback on top. Ready for another look whenever you're ready.

FYI @dianakhuang @andy-armstrong

cahrens:
Thanks, @efischer19, I'll look at this in the morning. Can you also create a sandbox with this and the christina/collapse-sections branch so we can verify that the behavior is what we want?

efischer19:
Superseded by https://github.com/edx/edx-ora2/pull/787

PR #784, opened 2015-11-18
Update translations from Transifex

andy-armstrong:


andy-armstrong:
@sarina Since this is my first time doing a Transifex update, would you mind spot checking this for me. It looks like it is okay. I had to fix a couple of invalid strings on Transifex and I verified that the fixes came through.

One thing I noticed is that I'm adding a whole lot of zero length files. Is this because there are more languages than before, or should the script not be adding them?

FYI @cahrens @dianakhuang @efischer19

sarina:
@andy-armstrong yeah you're pulling every language on transifex. In platform, we only pull a few key target languages: https://github.com/edx/edx-platform/blob/master/conf/locale/config.yaml#L3

sarina:
LGTM :+1: 

If you want, you could redo it only pulling the few target languages that we pull in platform. Overall it doesn't matter much, though.

PR #786, opened 2015-11-19
Move initialization code to __init__

efischer19:
@andy-armstrong I'm not able to reproduce the exact error, but this seems like it will fix things.

FYI @cahrens @dianakhuang


dianakhuang:
:+1: from me.

andy-armstrong:
:+1: Exactly what I was thinking! Thanks!

PR #787, opened 2015-11-19
Fix done definition

efischer19:
Moving to a new PR to point at existing frontend branch.

Fixes issues with defining when a workflow is done in the context of staff assessments. Included initial (reverted) commit, and review feedback.

@cahrens @dianakhuang @andy-armstrong 

Supersedes https://github.com/edx/edx-ora2/pull/783


openassessment/assessment/test/test_staff.py
```diff
@@ -78,8 +78,10 @@ def test_create_assessment_not_required(self, key):
         self.assertEqual(assessment["points_earned"], OPTIONS_SELECTED_DICT[key]["expected_points"])
         self.assertEqual(assessment["points_possible"], RUBRIC_POSSIBLE_POINTS)
 
-        # ensure submission is marked as finished
+        # Ensure submission and workflow are marked as finished
         self.assertTrue(staff_api.assessment_is_finished(tim_sub["uuid"], self.STEP_REQUIREMENTS))
```
cahrens: Nit: I'd consider a helper method here for these 3 lines, since I assume everyplace we check assessment_is_finished we should also check the workflow status.

openassessment/assessment/test/test_staff.py
```diff
@@ -118,7 +122,7 @@ def test_create_assessment_score_overrides(self, key):
             initial_assessment = OPTIONS_SELECTED_DICT["none"]
 
         # Create assessment
-        tim_sub, tim = self._create_student_and_submission("Tim", "Tim's answer")
+        tim_sub, tim = self._create_student_and_submission("Tim", "Tim's answer", override_steps=['self'])
```
cahrens: Consider renaming variable to be clearer.

openassessment/workflow/models.py
```diff
@@ -309,17 +312,20 @@ def update_from_assessments(self, assessment_requirements):
 
         new_staff_score = self.get_score(assessment_requirements, {'staff': step_for_name.get('staff', None)})
         if new_staff_score:
-            old_score = self.score
-            if not old_score or old_score['points_earned'] != new_staff_score['points_earned']:
+            old_score = sub_api.get_latest_score_for_submission(self.submission_uuid)
+            if not old_score or not old_score.get('staff_id') or old_score['points_earned'] != new_staff_score['points_earned']:
```
dianakhuang: This is kind of complicated, and I would suggest explaining what each clause is testing for (old score is a staff score, etc.)

efischer19:
feedback addressed

cahrens:
@efischer19 what is the status of this PR? I'm seeing on my branch off of ora-staff-grading that the workflow is marked "waiting" in the following scenario:

1) Learner submits response
2) Learner completes all steps required
3) Peer assessment is required and learner has NOT been assessed by peer(s)
4) Staff override has been submitted

It is my understanding that once a staff override has been submitted, the final grade should only be unavailable if the learner has not completed all of the tasks HE is required to do. I'm wondering if this PR addresses the above bug.

efischer19:
@cahrens This PR should fix the behavior you're describing. We've been sorting out integration errors and 1.8 issues, but I think those are largely addressed now. 

I'll get this rebased and updated on my sandbox shortly, and we'll see how that looks.

efischer19:
Superseded by https://github.com/edx/edx-ora2/pull/797

PR #793, opened 2015-11-23
Remove part of api test

efischer19:
This test is trying to confirm behavior when an api cannot be loaded.
It is no longer possible for an AssessmentWorkflow to have been created
without going through __init__(), so the latter part of this test is
now trying to confirm assumptions about an impossible state.

I recommend we just remove it. @andy-armstrong, thoughts?


andy-armstrong:
Thanks!

PR #794, opened 2015-11-23
Miscellaneous small improvements

andy-armstrong:
* Bump version numbers
* Introduce JSCS code style validation (http://jscs.info/)
* Address style failures

Note: I'm introducing JSCS because JSHint has deprecated all of their code style validations. They are switching to linting for code correctness only. See http://jshint.com/docs/options/#maxlen


andy-armstrong:
@cahrens @dianakhuang @efischer19 please review. You may want to compare this with ?w=1 as there are a lot of whitespace changes

dianakhuang:
:+1:

andy-armstrong:
@cahrens @efischer19 One thing to note is that PyCharm has support for JSCS and will automatically detect the .jscsrc file. I've been using this to help clean up the validation failures.

efischer19:
:+1: from me as well

.jshintrc
```diff
@@ -27,7 +27,7 @@
     "unused"          : true,      // Warns when you define and never use your variables.
     "strict"          : false,     // Requires all functions to run in ECMAScript 5's strict mode.
     "trailing"        : true,      // Makes it an error to leave a trailing whitespace in your code.
-    "maxlen"          : 120,       // Lets you set the maximum length of a line.
+    //"maxlen"        : 120,       // Lets you set the maximum length of a line.
```
cahrens: Any reason why we don't just delete maxlen? Is it worth a comment about why jshint can no longer check max line length?

cahrens:
Just a question about clarifying why maxlen is commented out. Otherwise :+1:

.jshintrc
andy-armstrong: Other places seemed to comment out options so that's what I did, but in this case we will never want to re-enable it so I agree that it makes more sense to remove it. I've done so, squashed and pushed it up.

PR #802, opened 2015-12-02
Add in StaffGradingWorkflow

dianakhuang:
## [TNL-3833](https://openedx.atlassian.net/browse/TNL-3833)

Add in the StaffGradingWorkflow for tracking ungraded submissions for each problem.

### Testing
- [x] Database migrations are backwards-compatible

### Reviewers

### Devops assistance
- [ ] @maxrothman - just as an FYI, we're adding a new table with this PR.

FYI: @robrap , @cahrens 

### Post-review
- [x] Squash commits

### DIFF
--- a/openassessment/assessment/api/staff.py
+++ b/openassessment/assessment/api/staff.py
@@ -3,11 +3,14 @@ Public interface for staff grading, used by students/course staff.
 """
 import logging
 from django.db import DatabaseError, IntegrityError, transaction
+from django.utils.timezone import now
 from dogapi import dog_stats_api
+
+from submissions import api as submissions_api
 
 from openassessment.assessment.models import (
     Assessment, AssessmentFeedback, AssessmentPart,
-    InvalidRubricSelection
+    InvalidRubricSelection, StaffWorkflow,
 )
 from openassessment.assessment.serializers import (
     AssessmentFeedbackSerializer, RubricSerializer,
@@ -54,6 +57,72 @@ def assessment_is_finished(submission_uuid, requirements):
     if requirements and requirements.get('staff', {}).get('required', False):
         return bool(get_latest_staff_assessment(submission_uuid))
     return True
+
+
+def on_start(submission_uuid):
+    """
+    Create a new staff workflow for a student item and submission.
+
+    Creates a unique staff workflow for a student item, associated with a
+    submission.
+
+    Args:
+        submission_uuid (str): The submission associated with this workflow.
+
+    Returns:
+        None
+
+    Raises:
+        StaffAssessmentInternalError: Raised when there is an internal error
+            creating the Workflow.
+
+    """
+    try:
+        submission = submissions_api.get_submission_and_student(submission_uuid)
+        workflow, __ = StaffWorkflow.objects.get_or_create(
+            course_id=submission['student_item']['course_id'],
+            item_id=submission['student_item']['item_id'],
+            submission_uuid=submission_uuid
+        )
+    except DatabaseError:
+        error_message = (
+            u"An internal error occurred while creating a new staff "
+            u"workflow for submission {}"
+            .format(submission_uuid)
+        )
+        logger.exception(error_message)
+        raise StaffAssessmentInternalError(error_message)
+
+
+def on_cancel(submission_uuid):
+    """
+    Cancel the staff workflow for submission.
+
+    Sets the cancelled_at field in staff workflow.
+
+    Args:
+        submission_uuid (str): The submission UUID associated with this workflow.
+
+    Returns:
+        None
+
+    """
+    try:
+        workflow = StaffWorkflow.objects.get(submission_uuid=submission_uuid)
+        workflow.cancelled_at = now()
+        workflow.save(update_fields=['cancelled_at'])
+    except StaffWorkflow.DoesNotExist:
+        # If we can't find a workflow, then we don't have to do anything to
+        # cancel it.
+        pass
+    except DatabaseError:
+        error_message = (
+            u"An internal error occurred while cancelling the staff"
+            u"workflow for submission {}"
+            .format(submission_uuid)
+        )
+        logger.exception(error_message)
+        raise StaffAssessmentInternalError(error_message)
 
 
 def get_score(submission_uuid, requirements):
@@ -160,6 +229,58 @@ def get_assessment_scores_by_criteria(submission_uuid):
         raise StaffAssessmentInternalError(error_message)
 
 
+def get_submission_to_assess(course_id, item_id, scorer_id):
+    """Get a submission for staff evaluation.
+
+    Retrieves a submission for assessment for the given staff member.
+
+    Args:
+        course_id (str): The course that we would like to fetch submissions from.
+        item_id (str): The student_item (problem) that we would like to retrieve submissions for.
+        scorer_id (str): The user id of the staff member scoring this submission
+
+    Returns:
+        dict: A student submission for assessment. This contains a 'student_item',
+            'attempt_number', 'submitted_at', 'created_at', and 'answer' field to be
+            used for assessment.
+
+    Raises:
+        StaffAssessmentInternalError: Raised when there is an internal error
+            retrieving staff workflow information.
+
+    Examples:
+        >>> get_submission_to_assess("a_course_id", "an_item_id", "a_scorer_id")
+        {
+            'student_item': 2,
+            'attempt_number': 1,
+            'submitted_at': datetime.datetime(2014, 1, 29, 23, 14, 52, 649284, tzinfo=<UTC>),
+            'created_at': datetime.datetime(2014, 1, 29, 17, 14, 52, 668850, tzinfo=<UTC>),
+            'answer': u'The answer is 42.'
+        }
+
+    """
+    student_submission_uuid = StaffWorkflow.get_submission_for_review(course_id, item_id, scorer_id)
+    if student_submission_uuid:
+        try:
+            submission_data = submissions_api.get_submission(student_submission_uuid)
+            return submission_data
+        except submissions_api.SubmissionNotFoundError:
+            error_message = (
+                u"Could not find a submission with the uuid {}"
+            ).format(student_submission_uuid)
+            logger.exception(error_message)
+            raise StaffAssessmentInternalError(error_message)
+    else:
+        logger.info(
+            u"No submission found for staff to assess ({}, {})"
+            .format(
+                course_id,
+                item_id,
+            )
+        )
+        return None
+
+
 def create_assessment(
     submission_uuid,
     scorer_id,
@@ -215,6 +336,11 @@ def create_assessment(
         >>> create_assessment("Tim", options_selected, criterion_feedback, feedback, rubric_dict)
     """
     try:
+        try:
+            scorer_workflow = StaffWorkflow.objects.get(submission_uuid=submission_uuid)
+        except StaffWorkflow.DoesNotExist:
+            scorer_workflow = None
+
         assessment = _complete_assessment(
             submission_uuid,
             scorer_id,
@@ -222,7 +348,8 @@ def create_assessment(
             criterion_feedback,
             overall_feedback,
             rubric_dict,
-            scored_at
+            scored_at,
+            scorer_workflow
         )
         return full_assessment_dict(assessment)
 
@@ -250,7 +377,8 @@ def _complete_assessment(
         criterion_feedback,
         overall_feedback,
         rubric_dict,
-        scored_at
+        scored_at,
+        scorer_workflow
 ):
     """
     Internal function for atomic assessment creation. Creates a staff assessment
@@ -295,4 +423,7 @@ def _complete_assessment(
     # match the rubric.
     AssessmentPart.create_from_option_names(assessment, options_selected, feedback=criterion_feedback)
 
+    # Close the active assessment
+    if scorer_workflow is not None:
+        scorer_workflow.close_active_assessment(assessment, scorer_id)
     return assessment
--- a/openassessment/assessment/migrations/0002_staffworkflow.py
+++ b/openassessment/assessment/migrations/0002_staffworkflow.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.utils.timezone
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('assessment', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='StaffWorkflow',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('scorer_id', models.CharField(max_length=40, db_index=True)),
+                ('course_id', models.CharField(max_length=40, db_index=True)),
+                ('item_id', models.CharField(max_length=128, db_index=True)),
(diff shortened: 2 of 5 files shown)


openassessment/assessment/api/staff.py
```diff
@@ -56,6 +58,42 @@ def assessment_is_finished(submission_uuid, requirements):
...
+
+    Args:
+        submission_uuid (str): The submission associated with this workflow.
+
+    Returns:
+        None
+
+    Raises:
+        SubmissionError: There was an error retrieving the submission.
+        PeerAssessmentInternalError: Raised when there is an internal error
```
robrap: copy and paste issue.

openassessment/assessment/api/staff.py
```diff
@@ -56,6 +58,42 @@ def assessment_is_finished(submission_uuid, requirements):
...
+        submission = submissions_api.get_submission_and_student(submission_uuid)
+        workflow, __ = StaffWorkflow.objects.get_or_create(
+            course_id=submission['student_item']['course_id'],
+            item_id=submission['student_item']['item_id'],
+            submission_uuid=submission_uuid
+        )
+        workflow.save()
+    except DatabaseError:
+        error_message = (
+            u"An internal error occurred while creating a new peer "
```
robrap: You may want to search for "peer" and see what you turn up. :)

openassessment/assessment/api/staff.py
```diff
@@ -5,9 +5,11 @@
 from django.db import DatabaseError, IntegrityError, transaction
 from dogapi import dog_stats_api
 
+from submissions import api as submissions_api
+
 from openassessment.assessment.models import (
     Assessment, AssessmentFeedback, AssessmentPart,
-    InvalidRubricSelection
+    InvalidRubricSelection, StaffWorkflow
```
andy-armstrong: Nit: add a trailing comma

openassessment/assessment/api/staff.py
```diff
@@ -160,6 +198,62 @@ def get_assessment_scores_by_criteria(submission_uuid):
...
 
 
+def get_submission_to_assess(course_id, item_id, scorer_id):
+    """Get a submission for staff evaluation.
+
+    Retrieves a submission for assessment for the given staff member. This will
+    not return a submission submitted by the requesting scorer.
+
+    Args:
+        submission_uuid (str): The submission UUID from the student
```
robrap: Can staff be a student in the same class?  Or is all of this just peer leftovers?

openassessment/assessment/models/staff.py
```diff
@@ -0,0 +1,117 @@
...
+
+class StaffWorkflow(models.Model):
+    """Internal Model for tracking Staff Assessment Workflow
+
+    This model can be used to determine the following information required
+    throughout the Staff Assessment Workflow:
+
+    1) Get next submission that requires assessment.
+    2) Does a submission have a staff assessments?
+    4) Does this staff member already have a submission open for assessment?
```
robrap: very minor: skipped 3

openassessment/assessment/api/staff.py
```diff
@@ -56,6 +58,41 @@ def assessment_is_finished(submission_uuid, requirements):
...
+    """
+    try:
+        submission = submissions_api.get_submission_and_student(submission_uuid)
+        workflow, __ = StaffWorkflow.objects.get_or_create(
+            course_id=submission['student_item']['course_id'],
+            item_id=submission['student_item']['item_id'],
+            submission_uuid=submission_uuid
+        )
+        workflow.save()
+    except DatabaseError:
```
andy-armstrong: Nit: can we capture the exception and include it in the error message. I don't like how ORA squelches the real errors and leaves you with generic messages in the logs.

openassessment/assessment/models/staff.py
```diff
@@ -0,0 +1,117 @@
...
+
+from django.db import models, DatabaseError
+from django.utils.timezone import now
+
+from openassessment.assessment.models.base import Assessment
+from openassessment.assessment.errors import StaffAssessmentInternalError
+
+
+class StaffWorkflow(models.Model):
+    """Internal Model for tracking Staff Assessment Workflow
```
robrap: I think the first line should follow """, even though that's not how Google does it.

openassessment/assessment/api/staff.py
```diff
@@ -56,6 +58,41 @@ def assessment_is_finished(submission_uuid, requirements):
     return True
 
 
+def on_start(submission_uuid):
+    """Create a new staff workflow for a student item and submission.
```
andy-armstrong: Nit: my understanding from Ned is that our standard is to wrap the first line, even though that isn't the Google standard.

openassessment/assessment/models/staff.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Returns:
+            True/False
+        """
+        return bool(self.cancelled_at)
+
+    @classmethod
+    def get_submission_for_review(cls, course_id, item_id, scorer_id):
+        """
+        Find a submission for peer assessment. This function will find the next
```
robrap: Same issue with hunting for "peer"

andy-armstrong: Typo: staff not peer

openassessment/assessment/models/staff.py
```diff
@@ -0,0 +1,117 @@
...
+        submission that requires assessment, excluding any submission that has been
+        completely graded, or is actively being reviewed by other students.
```
andy-armstrong: Nit: a comment above says that it won't return the staff member's own submissions either. Is that an issue? It might be good to disallow staff from reviewing their own submissions, but I think Leslie said she wasn't concerned about that.

openassessment/assessment/api/staff.py
```diff
@@ -160,6 +197,62 @@ def get_assessment_scores_by_criteria(submission_uuid):
...
 
 
+def get_submission_to_assess(course_id, item_id, scorer_id):
+    """Get a submission for staff evaluation.
+
+    Retrieves a submission for assessment for the given staff member. This will
+    not return a submission submitted by the requesting scorer.
+
+    Args:
+        submission_uuid (str): The submission UUID from the student
```
andy-armstrong: This whole comment string needs revising.

openassessment/assessment/api/staff.py
```diff
@@ -160,6 +197,62 @@ def get_assessment_scores_by_criteria(submission_uuid):
         raise StaffAssessmentInternalError(error_message)
 
 
+def get_submission_to_assess(course_id, item_id, scorer_id):
+    """Get a submission for staff evaluation.
+
+    Retrieves a submission for assessment for the given staff member. This will
+    not return a submission submitted by the requesting scorer.
```
andy-armstrong: Nit: does this really not return submissions from the scorer? I don't see any logic for this.

openassessment/assessment/models/staff.py
```diff
@@ -0,0 +1,117 @@
...
+                    models.Q(scorer_id='') | models.Q(grading_started_at__lte=timeout),
+                    course_id=course_id,
+                    item_id=item_id,
+                    grading_completed_at=None,
+                    cancelled_at=None,
+                )
+            if not staff_workflows:
+                return None
+
+            workflow = staff_workflows[0]
```
robrap: If there is a race condition here where multiple another staff might overwrite after a first staff initiates grading, do we protect against this later?  Or can the original staff possibly update with the wrong scorer_id?

openassessment/assessment/api/staff.py
efischer19: My 2 cents - I agree with Andy and Ned, wrapping the first line looks better IMO

openassessment/assessment/api/staff.py
```diff
@@ -56,6 +58,41 @@ def assessment_is_finished(submission_uuid, requirements):
...
+
+    """
+    try:
+        submission = submissions_api.get_submission_and_student(submission_uuid)
+        workflow, __ = StaffWorkflow.objects.get_or_create(
+            course_id=submission['student_item']['course_id'],
+            item_id=submission['student_item']['item_id'],
+            submission_uuid=submission_uuid
+        )
+        workflow.save()
```
efischer19: Fun fact I've picked up from heavily debugging SQL behind django models - `get_or_create(..)` will include a call to `save(force_insert=True)` if creating, so this `save()` is unneeded.

openassessment/assessment/models/staff.py
```diff
@@ -0,0 +1,117 @@
...
+    throughout the Staff Assessment Workflow:
+
+    1) Get next submission that requires assessment.
+    2) Does a submission have a staff assessments?
+    4) Does this staff member already have a submission open for assessment?
+    5) Close open assessments when completed.
+
+    """
+    # Amount of time before a lease on a submission expires
+    TIME_LIMIT = timedelta(hours=8)
```
efischer19: Would it be worthwhile to let this value be configurable, rather than hardcoded?

openassessment/assessment/models/staff.py
dianakhuang: We don't do a lot to protect against that race condition. We do try to 'lock' the workflow to the current scorer for 8 hours, but if two threads are executing this at the same time, then it's possible they'll grab the same entry and the person who gets it 'last' will win. I think this is okay, especially when staff sizes aren't particularly large.

openassessment/assessment/models/staff.py
dianakhuang: Currently it's not configurable for `PeerWorkflow`s, so I kept the logic mostly the same.

openassessment/assessment/test/test_staff.py
```diff
@@ -117,6 +123,9 @@ def test_create_assessment_required(self, key):
         # Verify assesment made, score updated, and no longer waiting
         self.assertEqual(staff_assessment["points_earned"], OPTIONS_SELECTED_DICT[key]["expected_points"])
         self._verify_done_state(tim_sub["uuid"], self.STEP_REQUIREMENTS_WITH_STAFF)
+        # Verify that a StaffWorkflow step has been marked as complete
+        workflow = StaffWorkflow.objects.get(submission_uuid=tim_sub['uuid'])
```
efischer19: Now that we're on django 1.8, refreshing a previously-fetch object from the database is possible: https://docs.djangoproject.com/en/1.8/ref/models/instances/#refreshing-objects-from-database

efischer19:
Done with my first pass, everything is looking good so far. Let me know when you're ready for final thumbs.

openassessment/assessment/migrations/0002_staffworkflow.py
```diff
@@ -0,0 +1,33 @@
...
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('scorer_id', models.CharField(max_length=40, db_index=True)),
+                ('course_id', models.CharField(max_length=40, db_index=True)),
+                ('item_id', models.CharField(max_length=128, db_index=True)),
+                ('submission_uuid', models.CharField(unique=True, max_length=128, db_index=True)),
+                ('created_at', models.DateTimeField(default=django.utils.timezone.now, db_index=True)),
+                ('grading_completed_at', models.DateTimeField(null=True, db_index=True)),
+                ('grading_started_at', models.DateTimeField(null=True, db_index=True)),
+                ('cancelled_at', models.DateTimeField(null=True, db_index=True)),
```
andy-armstrong: Nit: do we need to support canceling a staff workflow? I don't see any logic that will cause this to happen. Does it get cancelled automatically when the submission is cancelled from the staff tools?

openassessment/assessment/test/test_staff.py
dianakhuang: Oh man. I've been wanting this forever. Thanks for pointing it out!

openassessment/assessment/test/test_staff.py
```diff
@@ -382,6 +391,53 @@ def test_database_create_error_handling(self, mock_create):
...
+            patched_get_submission.side_effect = sub_api.SubmissionNotFoundError('Failed')
+            with self.assertRaises(staff_api.StaffAssessmentInternalError):
+                submission = staff_api.get_submission_to_assess(tim['course_id'], tim['item_id'], tim['student_id'])
+
+    def test_no_available_submissions(self):
+        tim_sub, tim = self._create_student_and_submission("Tim", "Tim's answer")
+        # Use a non-existent course and non-existent item.
+        submission = staff_api.get_submission_to_assess('test_course_id', 'test_item_id', tim['student_id'])
+        self.assertIsNone(submission)
+
```
andy-armstrong: If staff workflows can be cancelled, it would be good to add a test for the behavior in that case.

openassessment/assessment/migrations/0002_staffworkflow.py
dianakhuang: Good call. It will automatically try to cancel it if it exists in the api, but the staff api currently doesn't support it. I'll add that method in.

andy-armstrong:
Looks great @dianakhuang. I'm done with my first pass.

openassessment/assessment/api/staff.py
dianakhuang: I believe in all of these cases, we use `logger.exception`, which will log the original traceback and error as well as the generic error message. The pattern is a little odd, but I think it ultimately works the way we would like.

dianakhuang:
@efischer19 @andy-armstrong I've added cancellation support.

openassessment/assessment/api/staff.py
```diff
@@ -3,11 +3,14 @@
 """
 import logging
 from django.db import DatabaseError, IntegrityError, transaction
+from django.utils.timezone import now
 from dogapi import dog_stats_api
 
+from submissions import api as submissions_api
+
```
efischer19: Just out of curiosity, is there any standard pattern as to how to space out `import` statements? (blank line between or no)? I usually just go with "whatever looks easy to read", but if there's a standard I'd love to start using that.

openassessment/assessment/api/staff.py
```diff
@@ -56,6 +59,72 @@ def assessment_is_finished(submission_uuid, requirements):
...
+        submission_uuid (str): The submission UUID associated with this workflow.
+
+    Returns:
+        None
+
+    """
+    try:
+        workflow = StaffWorkflow.objects.get(submission_uuid=submission_uuid)
+        workflow.cancelled_at = now()
+        workflow.save()
```
efischer19: If only one field is being updated, you can use `update_fields` for a slight performance increase: https://docs.djangoproject.com/en/1.8/ref/models/instances/#specifying-which-fields-to-save

openassessment/assessment/api/staff.py
dianakhuang: I general, I try to put spaces around distinct categories of imports. Like, group together all the standard libraries, group together the django imports, group together any other libraries, and then group together any of the imports from the current project, etc.

efischer19:
One minor note about `update_fields`, otherwise :+1: from me

openassessment/assessment/api/staff.py
efischer19: That's about how I end up doing it too, good to know I'm not alone in my thinking!

andy-armstrong: Well, since you asked...

1. I follow the PEP8 guidance to group logical sets of imports (see https://github.com/edx/edx-platform/wiki/Python-Guidelines)
2. I put a line between each logical group
3. Within the group, I sort the imports alphabetically

Your mileage may vary as to whether this is too anal :-)

andy-armstrong:
:+1: Merge away!

PR #805, opened 2015-12-03
quick 'n dirty fix

efischer19:
@cahrens This ought to get you working again while we decide what the proper fix is for https://openedx.atlassian.net/browse/TNL-3865. @dianakhuang can you be the second set of thumbs?

Also I guess I accidentally fixed some non-empty lines? I can revert those if you want, but it seems like a "better than I found it" situation so if you don't mind I'll just leave it.

### DIFF
--- a/openassessment/xblock/staff_assessment_mixin.py
+++ b/openassessment/xblock/staff_assessment_mixin.py
@@ -52,7 +52,8 @@ class StaffAssessmentMixin(object):
                 create_rubric_dict(self.prompts, self.rubric_criteria_with_labels)
             )
             self.publish_assessment_event("openassessmentblock.staff_assessment", assessment)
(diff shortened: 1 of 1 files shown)


dianakhuang:
:+1: Though I definitely think we'll need to do something smarter about this later.

efischer19:
Agreed, this smells pretty funky to me. Definitely needs more attention in the future.

efischer19:
Argh, buttons! Agreed @dianakhuang, this "fix" smells pretty funky, we'll need to be smarter for a long-term solution

efischer19:
@dianakhuang I had to make a change to the commit - still the same idea, but now with less `KeyError` after running tests locally.

dianakhuang:
:+1:

efischer19:
Superseded by https://github.com/edx/edx-ora2/pull/806

PR #806, opened 2015-12-03
Fix falsy check for requirements

efischer19:
This makes `student_training.py` match what we did for `peer.py` back in https://github.com/edx/edx-ora2/pull/759. It seems silly that passing `None` returns false, while an empty dict causes errors.

For further discussion - is it the correct decision to always return `False` if there are no requirements? In my mind, no requirements means "nothing needs to be done" so the answer to the question "is the submitter finished?" is then `True`.

That doesn't make sense for what we're doing with staff grading though, in these cases it's more "we don't know the requirements so we can't rightly make any decision about whether the submitter is finished".

@dianakhuang - Let's discuss this further in the morning. This seems cleaner than https://github.com/edx/edx-ora2/pull/805, but we could also choose to have `None` be passed here(https://github.com/edx/edx-ora2/blob/ora-staff-grading/openassessment/xblock/staff_assessment_mixin.py#L55), or some other solution involving requirements.

### DIFF
--- a/openassessment/assessment/test/test_student_training_api.py
+++ b/openassessment/assessment/test/test_student_training_api.py
@@ -229,7 +229,7 @@ class StudentTrainingAssessmentTest(CacheResetTest):
             with self.assertRaises(StudentTrainingInternalError):
                 training_api.assess_training_example(self.submission_uuid, EXAMPLES[0]['options_selected'])
 
-    @ddt.data({}, {'num_required': 'not an integer!'})
+    @ddt.data({'some_other_field': 0}, {'num_required': 'not an integer!'})
     def test_submitter_is_finished_invalid_requirements(self, requirements):
         with self.assertRaises(StudentTrainingRequestError):
             training_api.submitter_is_finished(self.submission_uuid, requirements)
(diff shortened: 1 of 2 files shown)


openassessment/assessment/test/test_student_training_api.py
```diff
@@ -229,7 +229,7 @@ def test_assess_training_example_database_error(self):
             with self.assertRaises(StudentTrainingInternalError):
                 training_api.assess_training_example(self.submission_uuid, EXAMPLES[0]['options_selected'])
 
-    @ddt.data({}, {'num_required': 'not an integer!'})
+    @ddt.data({'some_other_field': 0}, {'num_required': 'not an integer!'})
```
efischer19: This test is now "`if requirements is not None or empty, but does not contain the field we expect` instead of `if requirements is not None`".

I'm not a huge fan of changing longstanding tests to pass when they fail this way, I'm a bit worried that maybe some other codepath depends on the old behavior. On the other hand, it seems super funky to have `student_training` error out on `{}`, while `peer` handles the same input just fine.

efischer19:
superseded by https://github.com/edx/edx-ora2/pull/807

PR #807, opened 2015-12-04
Use workflow_requirements()

efischer19:
~~@dianakhuang, here's another option for how to handle this situation, if we decide to not use https://github.com/edx/edx-ora2/pull/806. `workflow_requirements()` is defined here, and will ensure we have some value for student_training if that module exists: https://github.com/edx/edx-ora2/blob/ora-staff-grading/openassessment/xblock/workflow_mixin.py#L66~~

### DIFF
--- a/openassessment/assessment/api/peer.py
+++ b/openassessment/assessment/api/peer.py
@@ -45,7 +45,7 @@ def submitter_is_finished(submission_uuid, requirements):
         bool
 
     """
-    if not requirements:
+    if requirements is None:
         return False
 
     try:
@@ -59,6 +59,8 @@ def submitter_is_finished(submission_uuid, requirements):
         return False
     except PeerWorkflow.DoesNotExist:
         return False
+    except KeyError:
+        raise PeerAssessmentRequestError(u'Requirements dict must contain "must_grade" key')
 
 
 def assessment_is_finished(submission_uuid, requirements):
--- a/openassessment/workflow/models.py
+++ b/openassessment/workflow/models.py
@@ -201,30 +201,14 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
             score = sub_api.get_latest_score_for_submission(self.submission_uuid)
         return score
 
-    def status_details(self, assessment_requirements):
+    def status_details(self):
         status_dict = {}
         steps = self._get_steps()
         for step in steps:
-            api = step.api()
-            if api is not None:
-                # If an assessment module does not define these functions,
-                # default to True -- that is, automatically assume that the user has
-                # met the requirements.  This prevents students from getting "stuck"
-                # in the workflow in the event of a rollback that removes a step
-                # from the problem definition.
-                submitter_finished_func = getattr(api, 'submitter_is_finished', lambda submission_uuid, reqs: True)
-                assessment_finished_func = getattr(api, 'assessment_is_finished', lambda submission_uuid, reqs: True)
-
-                status_dict[step.name] = {
-                    "complete": submitter_finished_func(
-                        self.submission_uuid,
-                        assessment_requirements.get(step.name, {})
-                    ),
-                    "graded": assessment_finished_func(
-                        self.submission_uuid,
-                        assessment_requirements.get(step.name, {})
-                    ),
-                }
+            status_dict[step.name] = {
+                "complete": step.is_submitter_complete(),
+                "graded": step.is_assessment_complete(),
+            }
         return status_dict
 
     def get_score(self, assessment_requirements, step_for_name):
--- a/openassessment/xblock/test/test_peer.py
+++ b/openassessment/xblock/test/test_peer.py
@@ -33,8 +33,54 @@ class TestPeerAssessment(XBlockHandlerTestCase):
 
     @scenario('data/over_grade_scenario.xml', user_id='Bob')
     def test_load_peer_student_view_with_dates(self, xblock):
+
+        self._sally_and_hal_grade_each_other_helper(xblock)
+
+        # If Over Grading is on, this should now return Sally or Hal's response to Bob.
+        student_item = xblock.get_student_item_dict()
+        submission = xblock.create_submission(student_item, (u"Bob's answer 1", u"Bob's answer 2"))
+        workflow_info = xblock.get_workflow_info()
+        self.assertEqual(workflow_info["status"], u'peer')
+
+        # Validate Submission Rendering.
+        request = namedtuple('Request', 'params')
+        request.params = {}
+        peer_response = xblock.render_peer_assessment(request)
+        self.assertIsNotNone(peer_response)
+        self.assertNotIn(submission["answer"]["parts"][0]["text"].encode('utf-8'), peer_response.body)
+        self.assertNotIn(submission["answer"]["parts"][1]["text"].encode('utf-8'), peer_response.body)
+
+        # Validate Peer Rendering.
+        self.assertTrue("Sally".encode('utf-8') in peer_response.body or
+            "Hal".encode('utf-8') in peer_response.body)
+
+    @mock.patch('openassessment.xblock.workflow_mixin.WorkflowMixin.workflow_requirements')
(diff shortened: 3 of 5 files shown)


dianakhuang:
Yeah, I think this is the cleanest solution as of right now. :+1:

openassessment/assessment/api/peer.py
```diff
@@ -45,7 +45,7 @@ def submitter_is_finished(submission_uuid, requirements):
         bool
 
     """
-    if not requirements:
+    if requirements is None:
```
efischer19: I think both `peer` and `student_training` should at least be handled consistently, yeah?

efischer19:
I'm curious about this comment on `update_from_assessments`, repasted below: https://github.com/edx/edx-ora2/blob/efischer/training_reqs_take3/openassessment/workflow/models.py#L289

```
By convention, if `assessment_requirements` is `None`, then assessment
modules that need requirements should automatically say that they're incomplete.
This allows us to update the workflow even when we don't know the
current state of the problem.  For example, if we're updating the workflow
at the completion of an asynchronous call, we won't necessarily know the
current state of the problem, but we would still want to update assessments
that don't have any requirements.
```

I wonder, should we consider this update-on-staff-assess to be a case of updating "even when we don't know the current state of the problem", and pass `None`? Or can we consider the fact that `staff_assess` can call into `workflow_requirements()` "knowing the current state of the problem", and update submitter requirements accordingly?

openassessment/workflow/models.py
```diff
@@ -201,30 +201,14 @@ def score(self):
             score = sub_api.get_latest_score_for_submission(self.submission_uuid)
         return score
 
-    def status_details(self, assessment_requirements):
+    def status_details(self):
         status_dict = {}
         steps = self._get_steps()
         for step in steps:
-            api = step.api()
-            if api is not None:
```
efischer19: This whole block is more-or-less duplicated by https://github.com/edx/edx-ora2/blob/efischer/training_reqs_take3/openassessment/workflow/models.py#L636, and both blocks are executed (but yield slightly different results!) when `api.update_from_assessments()` is called.

If `status_details()` is to remain a "update, then get details" method, it should call the step's `update()` method instead of duplicating the logic. In that case, we should also update `api.update_from_assessments()` to only call this method.

I personally prefer this solution, where `update()` (writes data) and `status_details()` (reads data) are separated.

efischer19:
So as I see it, there are a few questions to answer before this gets merged:
* Is there any drawback to de-duplicating the "update steps" logic, as seen in https://github.com/edx/edx-ora2/commit/44c60ba8d06306913f369419b4f92f93f6567de6?
* Should `staff_assess()` pass in it's best guess at requirements, or `None`?

I'm leaning towards passing `None`. I see as as the staff assessment saying to the other assessment types "Hey, if you can be auto-completed, you're done, but I can't make any guarantees about what requirements may or may not have been provided/fulfilled".

Note that since `status_details` now queries the `AssessmentWorkflowStep` object instead of bypassing it to hit the underlying APIs directly, staff_assessments will actually be doing something when they mark `assessment_completed_at` here: https://github.com/edx/edx-ora2/blob/efischer/training_reqs_take3/openassessment/workflow/models.py#L313.

Previously, this method (https://github.com/edx/edx-ora2/blob/efischer/training_reqs_take3/openassessment/workflow/models.py#L600) on the `AssessmentWorkflowStep` object could return `True`, while the underlying API could return `False`.

efischer19:
@dianakhuang @ormsbee, could you give this a review?

Dave - I'm mostly curious if you had any background as to why this method: https://github.com/edx/edx-ora2/blob/master/openassessment/workflow/models.py#L183 duplicated much of the logic found in this method: https://github.com/edx/edx-ora2/blob/master/openassessment/workflow/models.py#L535 (called from `model.update_from_assessment()` here: https://github.com/edx/edx-ora2/blob/master/openassessment/workflow/models.py#L290).

I've elected to remove the duplication and use the `AssessmentWorkflowStep` as the canonical location for status information, unless there's reason not to.

dianakhuang:
It looks reasonable to me, but I'd like to see what @ormsbee has to say about it.

ormsbee:
@efischer19 and I talked this over in person a bit. As long as there are tests for the case where the assessment requirement changes as the student is going through the process and there is and that it properly pushes them to the next step on read (e.g. "original peer req was 5 assessments, student gets 4, new requirement is 3 = student gets advanced to the next step when they load their problem"), it looks good to me.

efischer19:
Note from talking with @ormsbee: check, and add if not present, a test case along the lines of:
* problem set up with (relatively) high requirements
* submitter does some of their required duties (peer assessing, what have you), but not all of them
* problem requirements are relaxed, submitter is now above the threshold
* when loading the page, submitter sees their status as finished, thanks to the new requirements.

It's my belief that this will be the case, as the only place that caused unit test failures when removing the requirements parameter from `status_details()` was a codepath that led back to `update_from_assessments()`, but I'll write a test to confirm.

efischer19:
Test has been added, and passes. @ormsbee @dianakhuang, could I get final review thumbs before merging this?

dianakhuang:
:+1: from me.

openassessment/assessment/api/peer.py
ormsbee: Can the `requirements` dict never be empty? I'd prefer to keep the more general version unless there's a specific reason to change it.

openassessment/xblock/test/test_peer.py
```diff
@@ -33,8 +33,54 @@ class TestPeerAssessment(XBlockHandlerTestCase):
...
+        request = namedtuple('Request', 'params')
+        request.params = {}
+        peer_response = xblock.render_peer_assessment(request)
+        self.assertIsNotNone(peer_response)
+        self.assertNotIn(submission["answer"]["parts"][0]["text"].encode('utf-8'), peer_response.body)
+        self.assertNotIn(submission["answer"]["parts"][1]["text"].encode('utf-8'), peer_response.body)
+
+        # Validate Peer Rendering.
+        self.assertTrue("Sally".encode('utf-8') in peer_response.body or
+            "Hal".encode('utf-8') in peer_response.body)
```
ormsbee: The `encode()` of "Sally" and "Hal" here are superfluous, no?

ormsbee: Oh, I see, you just moved this from below. Consider this a nitpick then. :-)

ormsbee:
Minor issue/question on whether `requirements` dict can be empty (as opposed to `None`). :+1: if either that's changed back to the falsy check, or if you're confident that it should not be a concern.

openassessment/assessment/api/peer.py
efischer19: I could go either way. This was recently changed in our feature branch (Note master still uses `None` here https://github.com/edx/edx-ora2/blob/master/openassessment/assessment/api/peer.py#L48), to fix a bug. Then it popped up again in `student_training.py` here https://github.com/edx/edx-ora2/blob/master/openassessment/assessment/api/student_training.py#L44, which was the impetus for this PR to begin. As long as we're consistent between the locations I'm happy.

I just defaulted to what was here before we started doing any staff grading work, do you think we should change both locations to `if not requirements`, or leave it as `if requirements is None`? Changing it to the more general case *seems* to be okay, but I can't be 100% confident that the different handling of `None` and `{}` wasn't intended at some point.

ormsbee: If it's an empty dict with the code as is, it'll throw a `KeyError` below when it tries to access `"must_grade"`. I'd like to see some handling of that, if we're sticking with the explicit `None` check.

Anyhow, you've clearly thought this through more than I have. I'm :+1: either way you want to go on this.

efischer19: Oh yeah good point, I thought we were handling that exception already. I'll update that to match `student_training.py`, but otherwise leave the explicit `None` check.

Thanks!

PR #810, opened 2015-12-07
Full workflow bok choy tests

cahrens:
TNL-3819

Adds 2 new "full workflow" tests-- one with staff assessment at the end of the process, and one with staff assessment at the beginning. These tests cover the following previously-existing bugs: unable to submit a peer assessment after staff override, wrong assessment information shown in staff tools, unable to submit staff override for a problem with training.

@andy-armstrong and @efischer19 please review.

### DIFF
--- a/test/acceptance/tests.py
+++ b/test/acceptance/tests.py
@@ -63,13 +63,28 @@ class OpenAssessmentTest(WebAppTest):
         'file_upload':
             u'courses/{test_course_id}/courseware/'
             u'57a3f9d51d424f6cb922f0d69cba868d/bb563abc989340d8806920902f267ca3/'.format(test_course_id=TEST_COURSE_ID),
+        'full_workflow':
+            u'courses/{test_course_id}/courseware/'
+            u'676026889c884ac1827688750871c825/181ea9ff144c4766be44eb8cb360e34f/'.format(test_course_id=TEST_COURSE_ID),
     }
 
     SUBMISSION = u"This is a test submission."
     LATEX_SUBMISSION = u"[mathjaxinline]( \int_{0}^{1}xdx \)[/mathjaxinline]"
     OPTIONS_SELECTED = [1, 2]
     STAFF_OVERRIDE_OPTIONS_SELECTED = [0, 1]
+    STAFF_OVERRIDE_SCORE = 1
+    STAFF_OVERRIDE_EXISTS = "COMPLETE"
+    STAFF_OVERRIDE_LEARNER_STEPS_NOT_COMPLETE = "YOU MUST COMPLETE THE STEPS ABOVE TO VIEW YOUR GRADE"
+    STAFF_AREA_SCORE = "Final grade: {} out of 8"
+    STAFF_OVERRIDE_STAFF_AREA_NOT_COMPLETE = "The problem has not been completed."
     EXPECTED_SCORE = 6
+    STUDENT_TRAINING_OPTIONS = [
+        [1, 2],
+        [0, 2]
+    ]
+
+    LEARNER_EMAIL = "learner@foo.com"
+    LEARNER_PASSWORD = "learner_password"
 
     def setUp(self, problem_type, staff=False):
         """
@@ -92,10 +107,25 @@ class OpenAssessmentTest(WebAppTest):
         self.staff_asmnt_page = AssessmentPage('staff-assessment', self.browser, self.problem_loc)
         self.grade_page = GradePage(self.browser, self.problem_loc)
 
+    def login_user(self, learner, email=LEARNER_EMAIL, password=LEARNER_PASSWORD):
+        """
+        Logs in an already existing user.
+
+        Args:
+            learner (str): the username of the user.
+            email (str): email (if not specified, LEARNER_EMAIL is used).
+            password (str): password (if not specified, LEARNER_PASSWORD is used).
+        """
+        auto_auth_page = AutoAuthPage(
+            self.browser, email=email, password=password, username=learner,
+            course_id=self.TEST_COURSE_ID, staff=True
+        )
+        auto_auth_page.visit()
+
     def do_self_assessment(self):
         """
-        Submits a self assessment, verifies the grade, and returns the username of the student
-        for which the self assessment was submitted.
+        Creates a user, submits a self assessment, verifies the grade, and returns the username of the
+        learner for which the self assessment was submitted.
         """
         self.auto_auth_page.visit()
         username = self.auto_auth_page.get_username()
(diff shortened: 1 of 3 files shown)


test/acceptance/tests.py
```diff
@@ -63,13 +63,26 @@ class OpenAssessmentTest(WebAppTest):
...
     }
 
     SUBMISSION = u"This is a test submission."
     LATEX_SUBMISSION = u"[mathjaxinline]( \int_{0}^{1}xdx \)[/mathjaxinline]"
     OPTIONS_SELECTED = [1, 2]
     STAFF_OVERRIDE_OPTIONS_SELECTED = [0, 1]
+    STAFF_OVERRIDE_SCORE = 1
+    STAFF_OVERRIDE_EXISTS = "COMPLETE"
+    STAFF_OVERRIDE_LEARNER_STEPS_NOT_COMPLETE = "YOU MUST COMPLETE THE STEPS ABOVE TO VIEW YOUR GRADE"
+    STAFF_OVERRIDE_STAFF_AREA_SCORE = "Final grade: 1 out of 8"
```
andy-armstrong: Can you make this a parameterized string so there can be other tests with a different final grade? I think it would also make the test clearer.

test/acceptance/tests.py
```diff
@@ -63,13 +63,26 @@ class OpenAssessmentTest(WebAppTest):
...
+    STAFF_OVERRIDE_STAFF_AREA_NOT_COMPLETE = "The problem has not been completed."
     EXPECTED_SCORE = 6
+    EXPECTED_STAFF_AREA_SCORE = "Final grade: 6 out of 8"
```
andy-armstrong: Ditto for this one. Having hard-coded scores seems limiting.

test/acceptance/tests.py
```diff
@@ -103,22 +116,84 @@ def do_self_assessment(self):
...
+        # Check that we've completed student training
+        try:
+            self.student_training_page.wait_for_complete()
+        except BrokenPromise:
+            self.fail("Student training was not marked complete.")
+
+    def do_peer_assessment(self, count=1, options=OPTIONS_SELECTED):
+        """
+        Does the specified number of peer assessments.
+        Args:
```
andy-armstrong: Nit: add a blank line before the args.

test/acceptance/tests.py
```diff
@@ -592,6 +623,210 @@ def test_file_upload(self):
...
+    STAFF_AREA_SELF_ASSESSMENT = ['Good', u'5', u'5', u'Excellent', u'3', u'3']
+
+    SUBMITTED_ASSESSMENT = [0, 3]
+    STAFF_AREA_SUBMITTED = ['Poor', u'', u'0', u'5', u'Excellent', u'', u'3', u'3']
+
+    def setUp(self):
+        super(FullWorkflowTest, self).setUp('full_workflow', staff=True)
+        self.staff_area_page = StaffAreaPage(self.browser, self.problem_loc)
+
+    def login_user(self, learner, email=LEARNER_EMAIL, password=LEARNER_PASSWORD):
```
andy-armstrong: I'd like to see these helper methods in a base class so that we can add further tests that perform submissions etc.

test/acceptance/tests.py
```diff
@@ -592,6 +623,210 @@ def test_file_upload(self):
...
+        self.login_user(learner)
+
+        # Learner 1 does peer assessment of learner 2 to complete workflow.
+        self.do_peer_assessment(options=self.SUBMITTED_ASSESSMENT)
+
+        # If learner does not now have a score, it means that "extra" submissions are in the system,
+        # and more need to be scored. Create additional learners and have them grade until learner has
+        # a grade (stopping after 5 additional learners).
+        count = 0
+        while self.grade_page.wait_for_page().score is None and count < 6:
```
andy-armstrong: It would be great if this logic was in a helper method so it can be reused, and so that it doesn't complicate this test.

test/acceptance/tests.py
```diff
@@ -592,6 +623,210 @@ def test_file_upload(self):
...
+            self_assessment: the expected fields in the self assessment section
+        """
+        self.staff_area_page.visit()
+        self.staff_area_page.show_learner(username)
+        self.staff_area_page.expand_learner_report_sections()
+        self.assertEqual(peer_assessments, self.staff_area_page.status_text('peer__assessments'))
+        self.assertEqual(submitted_assessments, self.staff_area_page.status_text('submitted__assessments'))
+        self.assertEqual(self_assessment, self.staff_area_page.status_text('self__assessment'))
+
+    def test_staff_override_at_end(self):
```
andy-armstrong: We'd talked about using DDT to test staff overrides of different types of problems (peer-only, peer-and-self, ai etc). I could see that maybe this would be redundant, but it does seem like there's a risk that the behavior might be different in only one situation.

I'm thinking it would be useful to have named configured scenarios, and then we could use DDT to apply tests like this to each one in turn.

andy-armstrong: I wonder if we can piggyback on the scenarios created for the Python tests, e.g.

https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/test/data/self_only_scenario.xml

andy-armstrong: Never mind. @cahrens and I spoke about this and this isn't what we had in mind. This is the most complex scenario so should give us confidence in the simpler ones too.

andy-armstrong:
These are great tests, @cahrens. I have some refactoring suggestions but :+1:

test/acceptance/tests.py
cahrens: For the acceptance criteria, it was my understanding that we wanted a full workflow with all the individual parts (self training, peer grading, self grading, and staff override). We also wanted to have the staff override at the beginning and the end, and wished to cover all bugs that had previously been found.

Doing multiple simpler scenarios via ddt isn't feasible due to the setup of the test (using the ORA test course), and it will substantially increase bok choy test time. We do already have targeted staff override tests for self assessment only (test_staff_override in StaffAreaTest)  and peer assessment only (see PeerAssessmentTestStaffOverride).

cahrens:
@andy-armstrong last commit responds to your refactoring requests.

test/acceptance/tests.py
```diff
@@ -592,6 +641,211 @@ def test_file_upload(self):
...
+    @retry()
+    @attr('acceptance')
+    def test_staff_override_at_end(self):
+        """
+        Scenario: complete workflow with staff override at the very end
+
+        Given that I have created a submission, completed training, and done a self assessment
+        And a second learner has also created a submission, training, and self assessment
+        Then I can assess a learner
+        And when another learner can assess me
```
efischer19: Nit: "When another learner assesses me"

efischer19:
:+1: Looks great, I'm glad to have this additional testing coverage

cahrens:
@andy-armstrong I squashed now, so here is my cleanup commit from your review, if you are interested: https://github.com/edx/edx-ora2/commit/f14ce53fc0e9652943676041dd99a099b97e94db

PR #814, opened 2015-12-08
Enable staff grading in studio

efischer19:
## [TNL-3469](https://openedx.atlassian.net/browse/TNL-3469)

Allows course authors to enable staff grading step in studio, and shows that as expected on LMS. See commit log for detailed changes.

### Sandbox
- [X] https://efischer19.sandbox.edx.org/dashboard

### Testing
- [x] Data is properly encoded in HTML templates to avoid XSS
- [X] Unit, integration, acceptance tests as appropriate

Note: I would like to add a "required staff grade" step to the full workflow bok choy test, once the code for staff to assess responses that way is finished up. I think it would be very good for test coverage if we had 2 "big" tests:
- The pre-existing test, with training, self, and peer steps, followed by a staff override
- A new test, with training, self, peer, and staff required steps. This would ensure that staff grading steps play nicely with other types (And confirm the correct handling of https://openedx.atlassian.net/browse/TNL-3899)

### Reviewers
- [x] Code review: @cahrens 
- [ ] Code review: @andy-armstrong  
- [X] Product review: @explorerleslie 
- [ ] Doc review: @catong 

FYI: @robrap

### Post-review
- [x] Squash commits


openassessment/xblock/validation.py
```diff
@@ -58,27 +63,40 @@ def _is_valid_assessment_sequence(assessments):
...
     sequence = [asmnt.get('name') for asmnt in assessments]
-    return sequence in valid_sequences
+    required = ['staff-assessment', 'peer-assessment', 'self-assessment']
+    optional = ['example-based-assessment', 'student-training']
+
+    # at least one of required?
+    if not any(name in required for name in sequence):
+        return False
+
+    # nothing expect what appears in required or optional
```
efischer19: typo: "except"

efischer19:
@dianakhuang and/or @andy-armstrong, could I get a quick sanity check on this code so far? I have it working up to the point where the setting can be set in the GUI without error.

FYI @robrap @cahrens

openassessment/xblock/validation.py
```diff
@@ -48,8 +48,13 @@ def _duplicates(items):
 def _is_valid_assessment_sequence(assessments):
     """
     Check whether the sequence of assessments is valid.
-    For example, we currently allow self-assessment after peer-assessment,
-    but do not allow peer-assessment before self-assessment.
+    TODO: this documentation is horribly outdated, what defines a valid sequence anymore?
```
efischer19: I'm particularly interested what people think here - I feel that explicitly hard-coding all the possible sequences is already excessively unwieldy, are the rules I'm working with valid? Should there be any other validation assumptions?

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
```diff
@@ -0,0 +1,66 @@
...
+    <div class="openassessment_inclusion_wrapper">
+        <input type="checkbox" id="include_staff_assessment"
+                {% if assessments.staff_assessment %} checked="true" {% endif %}>
+        <label for="include_staff_assessment">{% trans "Step: Staff Assessment" %}</label>
+    </div>
+    <div class="openassessment_assessment_module_editor">
+        <p id="staff_assessment_description_closed" class="openassessment_description_closed {% if assessments.staff_assessment %} is--hidden {% endif %}">
+            {% trans "Staff members assess learners' responses using the rubric for the assignment." %}
+        </p>
+        <div id="staff_assessment_settings_editor" class="assessment_settings_wrapper {% if not assessments.staff_assessment %} is--hidden {% endif %}">
```
cahrens: I'm pretty sure that all that is required for staff grading is a simple checkbox (see wireframes-- https://openedx.atlassian.net/wiki/display/TNL/%5BUX+Wireframes%5D+-+ORA+Staff+Grading). Dates and times don't need to be set.

openassessment/xblock/validation.py
```diff
@@ -58,27 +63,41 @@ def _is_valid_assessment_sequence(assessments):
...
-        ['example-based-assessment', 'peer-assessment', 'self-assessment'],
-        ['example-based-assessment', 'self-assessment', 'peer-assessment'],
-        ['example-based-assessment', 'student-training', 'peer-assessment'],
-        ['example-based-assessment', 'student-training', 'peer-assessment', 'self-assessment'],
-        ['example-based-assessment', 'student-training', 'self-assessment', 'peer-assessment'],
-    ]
-
     sequence = [asmnt.get('name') for asmnt in assessments]
-    return sequence in valid_sequences
+    required = ['staff-assessment', 'peer-assessment', 'self-assessment']
```
cahrens: Was it necessary to change this code? It is my understanding that staff assessment will always be at the end.

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
explorerleslie: That is correct.

openassessment/xblock/validation.py
efischer19: It wasn't strictly necessary - I just thought that expanding the list of explicitly hard-coded combinations to check was quickly becoming unwieldy. I could just expand the list above, and included `staff-assessment` as needed.

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
efischer19: Cool, I'll get this pulled out. :+1:

openassessment/xblock/validation.py
cahrens: I just wanted to make sure you knew that staff assessment is always at the end (which probably greatly simplifies the logic about what a valid sequence is).

efischer19:
How much should we be concerned with hiding, or restricting the use of staff tools while in studio vs lms? For example, when I'm authoring an ORA component with staff grading enabled, I see (and presumably can use) the staff tools:

screen shot 2015-12-09 at 10 51 27 am

@cahrens @explorerleslie

explorerleslie:
@efischer19 we haven't cared in the past, and as far as I know, no xblock / xmodule shows differently in studio vs. lms. The We should test to make sure nothing breaks, but I'd be surprised if anything did. Only if something breaks do I think we should look at hiding it.

efischer19:
Sounds good, thanks!

cahrens:
@efischer19 the collapse/expand behavior combined with checking is very strange and confusing. When you select the staff grading step, it also appears to collapse it (I reality, I think it is expanding to an empty area). I realize the collapse/expand behavior is the same for all steps. Can you put something in the "expanded" state so it doesn't look like selecting it also somehow disables it? Perhaps just the same text as in the "collapsed/unselected" state?

efischer19:
@cahrens I'll look into it, that should be possible

cahrens:
FYI that the staff grade step is getting correctly saved on the server! You will notice that a "staff grade" step appears and disappears as you toggle the setting in Studio and re-publish (in a user who does not have a staff override).

openassessment/xblock/static/js/src/studio/oa_edit_fields.js
```diff
@@ -152,8 +152,8 @@ OpenAssessment.ToggleControl.prototype = {
     },
 
     show: function() {
-        this.shownSection.removeClass('is--hidden');
         this.hiddenSection.addClass('is--hidden');
+        this.shownSection.removeClass('is--hidden');
```
efischer19: By re-ordering this so that we always remove `is--hidden` last, it's possible to use the same element for `shownSection` and `hiddenSection`, to keep an element from changing when using a ToggleControl.

efischer19:
@cahrens Huzzah! I went ahead and modified the expand/collapse behavior, if you want to re-cherry-pick and try it out.

cahrens:
@efischer19 It's better, but when you deselect the step, it causes a strange scrolling behavior (the step scrolls so that I can no longer see it).

openassessment/xblock/static/js/src/studio/oa_edit_validation_alert.js
```diff
@@ -86,6 +86,7 @@ OpenAssessment.ValidationAlert.prototype = {
 
             $('.oa_editor_content_wrapper', this.editorElement).each(function() {
                 $(this).css(styles);
+                $(this).scrollTop($(this).scrollTop() + alertHeight); //keep our relative scroll position the same
```
efischer19: @cahrens This fixes the weird scrolling behavior you were seeing. This was actually always happening, with every assessment type, but all the other types shrink in height when the validation alert is displayed so it never mattered.

The intent here is to tell the `editorElement` to set its scroll position in such a way as to not change what is displayed when the alert appears.

openassessment/xblock/validation.py
```diff
@@ -145,6 +161,12 @@ def validate_assessments(assessments, current_assessments, is_released, _):
             if assessment_dict.get('algorithm_id') not in ['ease', 'fake']:
                 return (False, _('The "algorithm_id" value must be set to "ease" or "fake"'))
 
+        # Staff grading must be required if it is the only step
+        if assessment_dict.get('name') == 'staff-assessment' and len(assessments) == 1:
+            required = assessment_dict.get('required')
+            if not required: # Captures both None and explicit False cases, both are invalid
+                return (False, _('The "required" value must be true if staff assessment is the only step.'))
```
efischer19: Note: I added this validation step after realizing that the addition of a `required` field made it possible to do this. A test case has been added to confirm this behavior as well.

efischer19:
The error @dianakhuang and I are seeing on sandboxes looks to be an issue with javascript files not getting updated properly. @andy-armstrong, have you ever seen anything like this?

The javascript that's getting run is not including `staffAssessmentView` as it should be doing here: https://github.com/edx/edx-ora2/blob/efischer/enable_staff_grading/openassessment/xblock/static/js/src/studio/oa_edit.js#L41

screen shot 2015-12-11 at 10 53 44 am


Eventually, we get to this block of code (https://github.com/edx/edx-ora2/blob/efischer/enable_staff_grading/openassessment/xblock/static/js/src/studio/oa_edit_settings.js#L250), and get an error `Uncaught TypeError: Cannot read property 'isEnabled' of undefined` when trying to access `view.assessmentViews['oa_staff_assessment_editor'].isEnabled()`.

screen shot 2015-12-11 at 10 55 52 am

efischer19:
@cahrens @dianakhuang All acceptance criteria from the TNL story have been addressed in this PR, can you review when you get a chance?

openassessment/assessment/test/test_staff.py
```diff
@@ -33,7 +33,7 @@ class TestStaffAssessment(CacheResetTest):
     """
 
     STEP_REQUIREMENTS = {}
-    STEP_REQUIREMENTS_WITH_STAFF = {'staff': {'required': True}}
+    STEP_REQUIREMENTS_WITH_STAFF = {'required': True}
```
dianakhuang: I'm guessing this is related to the bug fixes, but why did this need to change?

efischer19: You're right - it is related to the bug changes.

I had been carelessly double-unpacking the expected `{'staff': {'required': True}}` dictionary, first here: https://github.com/edx/edx-ora2/blob/efischer/enable_staff_grading/openassessment/workflow/models.py#L252, then again here: https://github.com/edx/edx-ora2/blob/ora-staff-grading/openassessment/assessment/api/staff.py#L57. That second location has been updated to this: https://github.com/edx/edx-ora2/blob/efischer/enable_staff_grading/openassessment/assessment/api/staff.py#L61

So this is to have the assessment level test call into its api with the actual, not the expected-but-unverified, value being passed in by the workflow model.

explorerleslie:
@efischer19 can we make staff grading default to OFF? Currently it's checked by default, but this will only be used for a small number of courses.

openassessment/assessment/api/staff.py
```diff
@@ -54,18 +54,27 @@ def assessment_is_finished(submission_uuid, requirements):
     Returns:
         True if a staff assessment has been completed for this submission or if not required.
     """
-    if requirements and requirements.get('staff', {}).get('required', False):
```
robrap: Are requirements no longer separated out into staff-specific requirements?  Or just in this method?  Should the param be staff_requirements?

Note: I'm commenting as I read, so this may become clear later.

openassessment/workflow/models.py
```diff
@@ -245,7 +252,7 @@ def get_score(self, assessment_requirements, step_for_name):
                         requirements = assessment_requirements.get(assessment_step_name, {})
```
robrap: Possibly rename to step_requirements to be more specific.

openassessment/xblock/static/js/spec/studio/oa_edit_assessments.js
```diff
@@ -250,4 +250,25 @@ describe("OpenAssessment edit assessment views", function() {
         it("Loads a description", function() { testLoadXMLExamples(view); });
         it("shows an alert when disabled", function() { testAlertOnDisable(view); });
     });
+
+    describe("OpenAssessment.EditStaffAssessmentView", function() {
+        var view = null;
+
+        beforeEach(function() {
+            var element = $("#oa_stafff_assessment_editor").get(0);
```
robrap: Too many f's in staff.

openassessment/xblock/validation.py
```diff
@@ -58,27 +62,37 @@ def _is_valid_assessment_sequence(assessments):
...
-        ['student-training', 'self-assessment', 'peer-assessment'],
-        ['example-based-assessment'],
-        ['example-based-assessment', 'self-assessment'],
-        ['example-based-assessment', 'peer-assessment'],
-        ['example-based-assessment', 'peer-assessment', 'self-assessment'],
-        ['example-based-assessment', 'self-assessment', 'peer-assessment'],
-        ['example-based-assessment', 'student-training', 'peer-assessment'],
-        ['example-based-assessment', 'student-training', 'peer-assessment', 'self-assessment'],
-        ['example-based-assessment', 'student-training', 'self-assessment', 'peer-assessment'],
-    ]
```
robrap: Does this method not have a test?  It seems like it would warrant one.  Also, the above seems like it could be used in the test of sequences to determine if it does pass.

In the original comment, it read "...do not allow peer-assessment before self-assessment", but the valid_sequences allow for this.  Which is correct, the code or the comment?

openassessment/xblock/validation.py
```diff
@@ -47,9 +47,13 @@ def _duplicates(items):
...
 def _is_valid_assessment_sequence(assessments):
     """
-    Check whether the sequence of assessments is valid.
-    For example, we currently allow self-assessment after peer-assessment,
-    but do not allow peer-assessment before self-assessment.
+    Check whether the sequence of assessments is valid. The rules enforced are:
+        -must have one of staff-, peer-, self-, or example-based-assessment listed
+        -in addition to those, only student-training is a valid entry
+        -no duplicate entries
+        -if staff-assessment is present, it must come last
```
robrap: I thought staff could grade at any time?  Is that still true?  Does this prohibit them, or does it just define the default order?

openassessment/assessment/api/staff.py
efischer19: Yeah, they're unpacked by step prior to hitting this code, I had previously been double-unpacking. (See https://github.com/edx/edx-ora2/blob/efischer/enable_staff_grading/openassessment/workflow/models.py#L252)

Good idea about renaming, that ought to make things clearer.

openassessment/workflow/models.py
efischer19: :+1:

openassessment/xblock/static/js/spec/studio/oa_edit_assessments.js
efischer19: Hmm, yeah that does seem odd. I would have thought we'd get a failure with this typo.

Regardless, I'll get this fixed.

openassessment/xblock/validation.py
efischer19: The previous code was correct, self then peer has been allowed in master since https://github.com/edx/edx-ora2/commit/0c0935ddb8d5a16d7ad41d4c05ee9b2b9cd65808.

There are a whole series of validation tests run on this logic, they basically try to load a series of valid/invalid json and verify that they are correctly accepted or rejected. See https://github.com/edx/edx-ora2/blob/efischer/enable_staff_grading/openassessment/xblock/test/test_validation.py#L21

openassessment/xblock/validation.py
efischer19: You're correct, the grading can take place at any time.

This order will make staff grading appear last on the LMS problem visually, but will not affect the ability of staff tools to grab the assessment and grade it at any point in the process.

efischer19:
@explorerleslie I'll get the default changed to off for this setting, no problem.

openassessment/xblock/static/js/spec/studio/oa_edit_assessments.js
```diff
@@ -250,4 +250,24 @@ describe("OpenAssessment edit assessment views", function() {
         it("Loads a description", function() { testLoadXMLExamples(view); });
         it("shows an alert when disabled", function() { testAlertOnDisable(view); });
     });
+
+    describe("OpenAssessment.EditStaffAssessmentView", function() {
+        var view = null;
+
+        beforeEach(function() {
+            view = new OpenAssessment.EditStaffAssessmentView();
```
efischer19: Spoke with @andy-armstrong about this. Since the Backbone code being tested doesn't rely on interacting with the DOM at all, I've removed the `element` reference here to make what is being tested more clear.

explorerleslie:
:+1:

efischer19:
@robrap Just to be super explicit with the 'XSS' step above - if there's no user-added display strings, only hard-coded data, we're clear from an XSS perspective yes?

openassessment/assessment/api/staff.py
```diff
@@ -26,7 +26,7 @@
 STAFF_TYPE = "ST"
 
 
-def submitter_is_finished(submission_uuid, requirements):
+def submitter_is_finished(submission_uuid, staff_requirements):
     """
     Determine if the submitter has finished their requirements for staff
     assessment. Always returns True.
```
cahrens: Update name of variable in doc string.

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
```diff
@@ -0,0 +1,17 @@
+{% load i18n %}
+{% load tz %}
```
cahrens: Nit: tz not needed here (spaceless probably not either, but I'm OK with that one staying).

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
```diff
@@ -0,0 +1,17 @@
...
+{% spaceless %}
+<li class="openassessment_assessment_module_settings_editor" id="oa_staff_assessment_editor">
+    <div class="drag-handle action"></div>
+    <div class="openassessment_inclusion_wrapper">
+        <input type="checkbox" id="include_staff_assessment"
+                {% if assessments.staff_assessment.required %} checked="true" {% endif %}>
```
cahrens: Nit: spacing.

openassessment/assessment/api/staff.py
efischer19: This function is called polymorphically here, so all the api types need to accept the same arguments in their method signature.

https://github.com/edx/edx-ora2/blob/efischer/enable_staff_grading/openassessment/workflow/models.py#L663

I'll get the docstring updated though.

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
```diff
@@ -0,0 +1,17 @@
...
+        <label for="include_staff_assessment">{% trans "Step: Staff Assessment" %}</label>
+    </div>
+    <div class="openassessment_assessment_module_editor">
+        <p id="staff_assessment_description" class="openassessment_description">
+            {% trans "Staff members assess learners' responses using the rubric for the assignment." %}
```
robrap: @efischer19 In terms of XSS, the guideline is to escape everything, user data or not.  For example, in Mako templates, this is what will enable us to eventually switch to escaping by default.

This template seems to be a Django template, which is already safe by default, so I think all of these {% %} variables will be properly escaped.

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
```diff
@@ -0,0 +1,17 @@
+{% load i18n %}
+{% load tz %}
+{% spaceless %}
+<li class="openassessment_assessment_module_settings_editor" id="oa_staff_assessment_editor">
+    <div class="drag-handle action"></div>
+    <div class="openassessment_inclusion_wrapper">
+        <input type="checkbox" id="include_staff_assessment"
+                {% if assessments.staff_assessment.required %} checked="true" {% endif %}>
+        <label for="include_staff_assessment">{% trans "Step: Staff Assessment" %}</label>
```
cahrens: I believe the preferred syntax is to next the input within the label (correct, @cptvitamin?). This removes the necessity of using an ID to link the two together, and we want to avoid IDs if at all possible.

That being said, I'm sure you following the format of the other templates. So if you wish to leave this and we fix them all at once, I'm OK with that.

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
```diff
@@ -0,0 +1,17 @@
...
+    </div>
+    <div class="openassessment_assessment_module_editor">
+        <p id="staff_assessment_description" class="openassessment_description">
```
cahrens: Can you use the class instead of the ID (and remove the ID)?

openassessment/templates/openassessmentblock/edit/oa_edit_staff_assessment.html
cahrens: Yes, the Django templates are "safe by default".

efischer19:
@catong Sorry I forgot to tag you earlier - could I get a doc review of the user-facing strings in this PR? I believe the only new ones are in this file https://github.com/edx/edx-ora2/pull/814/files#diff-fdf8d135bc8ca5ac3bc0c18f02ba57ea

catong:
Thanks @efischer19 What's your timeframe for this PR? OK if I review by tomorrow?

efischer19:
That sounds great @catong, if we could get this merged tomorrow morning that'd be super. Thanks!

openassessment/xblock/static/js/src/studio/oa_edit_assessment.js
```diff
@@ -621,3 +621,104 @@ OpenAssessment.EditExampleBasedAssessmentView.prototype = {
...
+
+    /**
+     Return a description of the assessment.
+
+     Returns:
+     object literal
+     **/
+    description: function() {
+        return {
+            required: this.isEnabled(),
```
cahrens: How is "required" used? Can it be deleted?

openassessment/xblock/static/js/src/studio/oa_edit_assessment.js
```diff
@@ -621,3 +621,104 @@ OpenAssessment.EditExampleBasedAssessmentView.prototype = {
...
+    /**
+     Get or set whether the assessment is enabled.
+
+     Args:
+     isEnabled (boolean, optional): If provided, set the enabled state of the assessment.
+
+     Returns:
+     boolean
+     ***/
+    isEnabled: function(isEnabled) {
```
cahrens: This code is crying out for a base class that the various editors can extend (since it seems like there is a lot of boiler-plate code). However, I reluctantly say "not now".

openassessment/xblock/studio_mixin.py
```diff
@@ -171,8 +171,17 @@ def update_editor_context(self, data, suffix=''):
 
         # Check that the editor assessment order contains all the assessments.  We are more flexible on example-based.
         if set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) != (set(data['editor_assessments_order']) - {'example-based-assessment'}):
-            logger.exception('editor_assessments_order does not contain all expected assessment types')
-            return {'success': False, 'msg': self._('Error updating XBlock configuration')}
+            # Backwards compatibility: "staff-assessment" may not be present.
+            # If that is the only problem with this data, just add it manually and continue.
+            if set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) == (
+                # Check the given set, minus example-based, plus staff
+                (set(data['editor_assessments_order']) - {'example-based-assessment'}) | {'staff-assessment'}
```
cahrens: Nit: create a variable for set(data['editor_assessments_order']) - {'example-based-assessment'} so the comment can live with it (I was wondering why we don't care about example-based-assessment, but see the comment explaining "we are more flexible" above).

openassessment/xblock/test/data/invalid_assessments.json
```diff
@@ -118,6 +118,23 @@
         "current_assessments": null,
         "is_released": false
     },
+    "remove_staff_mid_flight": {
+        "assessments": [
+            {
```
cahrens: How is this invalid?

openassessment/xblock/test/data/update_from_xml_error.json
```diff
@@ -321,6 +321,26 @@
         ]
     },
 
+    "optional_self_assessment": {
```
cahrens: How is this related to the staff assessment step?

cahrens:
@efischer19 looking good-- just a few minor nits/questions. I'm done reviewing.

openassessment/xblock/static/js/src/studio/oa_edit_assessment.js
efischer19: `description` gets serialized and passed to the backend as the requirements for a given step. Since staff assessment is unique in that the presence of the step does not tell us any information about requirements, we need to explicitly put that information here.

Stack trace, if anyone is curious:
* `description` value is grabbed here: https://github.com/edx/edx-ora2/blob/1583a5d2fe6379dd10512502094d465b0ca20c67/openassessment/xblock/static/js/src/studio/oa_edit_settings.js
* that function if called from here: https://github.com/edx/edx-ora2/blob/1583a5d2fe6379dd10512502094d465b0ca20c67/openassessment/xblock/static/js/src/studio/oa_edit.js#L202
* the return value is passed to here: https://github.com/edx/edx-ora2/blob/1583a5d2fe6379dd10512502094d465b0ca20c67/openassessment/xblock/static/js/src/oa_server.js#L450
* which updates the xblock definitions here: https://github.com/edx/edx-ora2/blob/2bd3d4017491c30e4fdd077829db69b2b8e33d9c/openassessment/xblock/studio_mixin.py#L148

openassessment/xblock/test/data/invalid_assessments.json
efischer19: From the acceptance criteria: `Once a problem is live, ensure that you cannot remove staff grading`. This test case is trying to change this (`assessments` and `current_assessments` represent the new and previous states, respectively)

ORA2 is already set up to disallow removing an assessment type after the problem is published (see the block here: https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/validation.py#L148), so we get this for free. This is just adding a test case to make the staff-specific behavior be explicitly tested. 

`remove_peer_mid_flight` does the same thing just below this, but for peer assessment.

openassessment/xblock/test/data/update_from_xml_error.json
efischer19: It tests the new enforcement that only staff assessment steps can have a `required` value.

Relevant enforcement: https://github.com/edx/edx-ora2/blob/efischer/enable_staff_grading/openassessment/xblock/xml.py#L575

cahrens:
:+1:

efischer19:
The diff doesn't show it due to being a binary file, but the test course attached to this PR has had 2 sections with corresponding subsections and ORA problems added. I've verified the staff required bok choy test on this PR passes against my sandbox with the newest course.

screen shot 2015-12-15 at 3 27 13 pm

openassessment/xblock/static/js/src/studio/oa_edit_assessment.js
```diff
@@ -621,3 +621,104 @@ OpenAssessment.EditExampleBasedAssessmentView.prototype = {
     validationErrors: function() { return []; },
     clearValidationErrors: function() {}
 };
+
+/**
```
andy-armstrong: Nit: this isn't a valid JSDoc doc string (I know that the whole file uses this Python-like convention, but still...). I've been fixing a file at a time as I run into such problems, but if you'd prefer just don't introduce any new occurrences.

http://usejsdoc.org/about-getting-started.html

openassessment/xblock/static/js/src/studio/oa_edit_assessment.js
```diff
@@ -621,3 +621,104 @@ OpenAssessment.EditExampleBasedAssessmentView.prototype = {
...
+ OpenAssessment.EditStaffAssessmentView
+
+ **/
+OpenAssessment.EditStaffAssessmentView = function(element) {
+    this.element = element;
+    this.name = "staff-assessment";
+
+    // Configure the toggle checkbox to enable/disable this assessment
+    new OpenAssessment.ToggleControl(
+        $("#include_staff_assessment", this.element),
```
andy-armstrong: Nit: use classes rather than ids in your HTML.

openassessment/xblock/static/js/src/studio/oa_edit_assessment.js
```diff
@@ -621,3 +621,104 @@ OpenAssessment.EditExampleBasedAssessmentView.prototype = {
...
+     **/
+    validationErrors: function() {
+        return [];
+    },
+
+    /**
+     Clear all validation errors from the UI.
+     **/
+    clearValidationErrors: function() {
+    },
```
andy-armstrong: Nit: add a comment along the lines of:

    // do nothing

openassessment/xblock/studio_mixin.py
```diff
@@ -170,9 +170,19 @@ def update_editor_context(self, data, suffix=''):
             return {'success': False, 'msg': self._('Error updating XBlock configuration')}
 
         # Check that the editor assessment order contains all the assessments.  We are more flexible on example-based.
-        if set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) != (set(data['editor_assessments_order']) - {'example-based-assessment'}):
-            logger.exception('editor_assessments_order does not contain all expected assessment types')
-            return {'success': False, 'msg': self._('Error updating XBlock configuration')}
+        given_without_example_based = set(data['editor_assessments_order']) - {'example-based-assessment'}
+        if set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) != (given_without_example_based):
```
andy-armstrong: Nit: are the parens around ``given_without_example_based`` necessary?

andy-armstrong:
Excellent stuff @efischer19! I love PRs where there are more lines of tests than lines of code. I have some very minor nits that you can address or ignore.

efischer19:
This is one (of possibly many, just the first I noticed when trying it) place where switching from `id` to `class` in HTML needs to be done wholesale, not piecemeal. https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/static/js/src/studio/oa_edit_settings.js#L258

All other nits addressed though.

efischer19:
@catong Since this is going to be a small doc review relative to the remainder of this PR, we're going ahead with the merge and will file a new PR for the doc updates if needed. I'll update here and tag you.

PR #816, opened 2015-12-11
JS debugging

efischer19:
I was able to reproduce the sandbox error locally, by reverting this file to `master`. So it definitely looks like an issue where the js files aren't being updated before being served in sandboxes.

FYI @andy-armstrong @dianakhuang


efischer19:
Ruling out `collectstatic` as a culprit; I forced that to happen on my sandbox and restarted workers, still saw the error.

PR #817, opened 2015-12-11
Django 1.8 modernizations

efischer19:
According to https://docs.djangoproject.com/en/1.9/ref/templates/upgrading/,
there have been some changes made to TEMPLATES in settings files that we
should be using. This commit contains those, along with a few minor fixes
to allow tests to be run as expected (without cluttering stdout via logging,
with coverage on python tests) that aren't yet in master.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -16,6 +16,7 @@ if __name__ == "__main__":
         import logging
         logging.captureWarnings(True)
         sys.argv.append('--noinput')
+        sys.argv.append('--logging-clear-handlers')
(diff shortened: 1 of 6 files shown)


efischer19:
I came across this while digging into a separate issue. It should probably be merged at some point, but I'm not terribly worried about it right now.

This didn't fix the issue I was investigating, unfortunately.

efischer19:
Closing, I don't believe this to be relevant any longer

PR #821, opened 2015-12-15
Doc review for enable staff grading in Studio

efischer19:
@catong Here's the PR I mentioned.

I've added some blank lines to flag the strings that need to be looked at. Thanks!

Sandbox: https://studio-efischer19.sandbox.edx.org/home/

FYI @cahrens @andy-armstrong

### DIFF
--- a/openassessment/templates/openassessmentblock/edit/oa_edit.html
+++ b/openassessment/templates/openassessmentblock/edit/oa_edit.html
@@ -157,9 +157,9 @@
 
             <p class="openassessment_description" id="openassessment_step_select_description">
                 {% if 'example_based_assessment' in editor_assessments_order %}
-                    {% trans "In this assignment, you can include steps for learner training, peer assessment, self assessment, and example based assessment. Select the steps that you want below, and then drag them into the order that you want. If you include an example based assessment step, it must precede all other steps. If you include a learner training step, it must precede peer and self assessment steps." %}
+                    {% trans "In this assignment, you can include learner training, peer assessment, self assessment, example based assessment, and staff assessment steps. Select the steps that you want below, and then drag them into the order you want. If you include an example based assessment step, it must precede all other steps. If you include a learner training step, it must precede peer and self assessment steps. If you include a staff assessment step, it must be the final step. After you publish an ORA assignment, you cannot change the type and number of assessment steps." %}
                 {% else %}
-                    {% trans "In this assignment, you can include steps for learner training, peer assessment, and self assessment. Select the steps that you want below, and then drag them into the order that you want. If you include a learner training step, it must precede all other steps." %}
+                    {% trans "In this assignment, you can include learner training, peer assessment, self assessment, and staff assessment steps. Select the steps that you want below, and then drag them into the order you want. If you include a learner training step, it must precede all other steps. If you include a staff assessment step, it must be the final step. After you publish an ORA assignment, you cannot change the type and number of assessment steps." %}
                 {% endif %}
             </p>
             <ol id="openassessment_assessment_module_settings_editors">
--- a/openassessment/xblock/validation.py
+++ b/openassessment/xblock/validation.py
@@ -96,11 +96,9 @@ def _is_valid_assessment_sequence(assessments):
 
 def validate_assessments(assessments, current_assessments, is_released, _):
     """
-    Check that the assessment dict is semantically valid.
-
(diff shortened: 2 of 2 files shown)


catong:
Thanks @efischer19 The strings in oa_edit_staff_assessment.html look great, no edits. 
In looking at the sandbox though, I see that there's an intro paragraph that probably should be updated to include reference to staff assessment. Screenshot attached. String is in oa_edit.html, I think.
oa_edit_intro

Suggested rewrite: "In this assignment, you can include learner training, peer assessment, self assessment, and staff assessment steps. Select the steps that you want below, and then drag them into the order you want. If you include a learner training step, it must precede all other steps. If you include a staff assessment step, it must be the final step." (@efischer19 Is this true? Or do we automatically display the staff step last in the LMS problem, regardless of where it's placed in the Settings selection?)

efischer19:
Sounds good @catong, I'll get that updated here in a bit. Your suggested edit is true, staff must always come last (code is in place to enforce this restriction).

Question for @explorerleslie - we also enforce a rule that the type and number of assessments cannot be changed after the problem is published. Should that be called out here as well? Otherwise it only shows up as an error message if you try to change those items:

screen shot 2015-12-16 at 11 58 23 am

explorerleslie:
@efischer19 do we call out that rule for peer or self? or is this new with staff? if we already have copy for  for peer and self, we should duplicate that here. If not, can you create a separate story to do it for all three?

efischer19:
@catong Can you take a quick look at the additional language around changing assessments steps after a problem has been released? Screenshot attached, sandbox is updated if you want to check there as well.

screen shot 2015-12-16 at 1 20 19 pm

FYI @explorerleslie

openassessment/templates/openassessmentblock/edit/oa_edit.html
```diff
@@ -157,9 +157,9 @@ <h2 class="openassessment_alert_title">{% trans "Rubric Change Impacts Settings
...
                 {% else %}
-                    {% trans "In this assignment, you can include steps for learner training, peer assessment, and self assessment. Select the steps that you want below, and then drag them into the order that you want. If you include a learner training step, it must precede all other steps." %}
+                    {% trans "In this assignment, you can include learner training, peer assessment, self assessment, and staff assessment steps. Select the steps that you want below, and then drag them into the order you want. If you include a learner training step, it must precede all other steps. If you include a staff assessment step, it must be the final step. The type and number of assessment steps cannot be changed after a problem has been released." %}
```
efischer19: @catong Just a heads up - I had to modify this string twice, with a slight difference depending on whether example-based is included or not.

openassessment/templates/openassessmentblock/edit/oa_edit.html
```diff
@@ -157,9 +157,9 @@ <h2 class="openassessment_alert_title">{% trans "Rubric Change Impacts Settings
 
             <p class="openassessment_description" id="openassessment_step_select_description">
                 {% if 'example_based_assessment' in editor_assessments_order %}
-                    {% trans "In this assignment, you can include steps for learner training, peer assessment, self assessment, and example based assessment. Select the steps that you want below, and then drag them into the order that you want. If you include an example based assessment step, it must precede all other steps. If you include a learner training step, it must precede peer and self assessment steps." %}
+                    {% trans "In this assignment, you can include learner training, peer assessment, self assessment, example based assessment, and staff assessment steps. Select the steps that you want below, and then drag them into the order you want. If you include an example based assessment step, it must precede all other steps. If you include a learner training step, it must precede peer and self assessment steps. If you include a staff assessment step, it must be the final step. The type and number of assessment steps cannot be changed after a problem has been released." %}
```
catong: @efischer19 @explorerleslie 
Minor edits to the last sentence to make it active rather than passive voice:
"After you publish an ORA assignment, you cannot change the type and number of assessment steps."

openassessment/templates/openassessmentblock/edit/oa_edit.html
catong: Same edit as above.

openassessment/templates/openassessmentblock/edit/oa_edit.html
efischer19: Thanks! Updated

catong:
@efischer19 This looks great, thanks! :+1:

PR #822, opened 2015-12-16
Staff grading bok choy tests

efischer19:
Sorry for the late update to this top comment!

New bok choy tests for required staff grading. The 2 relevant tests are `test_staff_assessment`, which had previously used the staff override mechanic as a temporary hack, and `test_train_self_peer_staff`, which runs through a full workflow of training, self, peer, and staff assessment.

The latter involved some refactoring to share common code with the previous full workflow tests, and both my new tests make use of some new helper or page methods to accomplish required staff grading.

Reviewers: @andy-armstrong and @cahrens


scripts/test-acceptance.sh
```diff
@@ -53,4 +53,4 @@ if [[ "${test_name}" = "accessibility" ]]; then
 fi
 
 echo "Running acceptance tests from ${test_name}.py against the sandbox..."
-python ../acceptance/${test_name}.py
+python ../acceptance/${test_name}.py StaffAssessmentTest.test_staff_assessment
```
efischer19: TODO: remove once done with new tests.

openassessment/templates/openassessmentblock/staff_area/oa_staff_override_assessment.html
```diff
@@ -39,7 +39,7 @@ <h3 class="message__title">{% trans "We could not submit your assessment" %}</h3
                 <ul class="list list--actions">
                     <li class="list--actions__item">
                         <button type="submit" class="action action--submit is--disabled">
-                            <span class="copy">{% trans "Submit your assessment" %}</span>
+                            <span class="copy">{% trans "Submit Assessment" %}</span>
```
efischer19: This update makes this text and https://github.com/edx/edx-ora2/blob/efischer/new_bok_choy/openassessment/templates/openassessmentblock/staff_area/oa_staff_grade_learners_assessment.html#L42 consistent.

test/acceptance/pages.py
```diff
@@ -557,11 +570,24 @@ def learner_response(self):
...
+            )
+            self.q(css=self._bounded_selector(sel)).first.click()
+        self.submit_assessment(continue_after)
+
+    def submit_assessment(self, continue_after=False):
         """
         Submit a staff assessment of the problem.
         """
-        self.submit(button_css=".wrapper--staff-assessment .action--submit")
+        filter_text = "Submit Assessment"
```
efischer19: This is why the above consistency is needed.

test/acceptance/tests.py
```diff
@@ -762,25 +783,26 @@ def verify_submission_has_peer_grade(self, learner, max_attempts=5):
...
             count += 1
             self.do_submission_training_self_assessment("extra_{}@looping.com".format(count), None)
             self.do_peer_assessment(options=self.PEER_ASSESSMENT)
             self.login_user(learner)
             self.grade_page.visit()
 
-        self.assertIsNotNone(
-            self.grade_page.wait_for_page().score,
+        self.assertTrue(
+            peer_page.is_complete,
```
efischer19: This allows the full workflow with staff required test to proceed properly after peer grading finishes.

test/acceptance/tests.py
```diff
@@ -923,6 +945,76 @@ def test_staff_override_at_beginning(self):
         ])
 
 
+class FullWorkflowRequiredTest(OpenAssessmentTest, FullWorkflowMixin):
```
efischer19: I'm going to try and refactor some of the common code out of this and `FullWorkflowOverrideTest`, as well as add some further assertions. I just wanted to get it functioning all the way through first.

efischer19:
The test course needs to be updated with this PR as well - previously I had created the `All Assessment Type Section` with a custom problem, which broke all the bok choy tests.

The useful problem is live on my sandbox at https://efischer19.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb/

test/acceptance/tests.py
```diff
@@ -923,6 +963,45 @@ def test_staff_override_at_beginning(self):
...
+
+        # Now do a staff assessment.
+        self.do_staff_assessment(options_selected=self.STAFF_OVERRIDE_OPTIONS_SELECTED)
+
+        # At this point, the learner sees the score (1).
+        self.browser.refresh()
+        self._verify_staff_grade_section(self.STAFF_GRADE_EXISTS, None)
+        self.assertEqual(self.STAFF_OVERRIDE_SCORE, self.grade_page.wait_for_page().score)
+
+        # Note that PEER ASSESSMENT isn't shown here - it gets cut off the page in this case
```
efischer19: @andy-armstrong This was what piqued my curiosity in planning earlier. Is this expected, or should peer be showing up as well?

screen shot 2015-12-17 at 12 54 34 pm

efischer19:
@cahrens @andy-armstrong These tests are ready for review when you get a moment. I feel like there may need to be some additional assertions made in the FullWorkflowRequired test, but I'm not quite sure what or where those should be.

test/acceptance/tests.py
efischer19: Nevermind - talked to Christina, realized that having a calculated peer grade is not a prerequisite for having a final grade in this scenario, so it makes sense that there wouldn't be a peer grade section in this view.

openassessment/templates/openassessmentblock/staff_area/oa_staff_override_assessment.html
cahrens: @efischer19 I asked Carol about this change in the main PR. I would remove this from your branch.

test/acceptance/pages.py
```diff
@@ -491,6 +491,19 @@ def show_learner(self, username):
...
+    def expand_staff_grading_section(self):
+        self.click_staff_toolbar_button("staff-grading")
+        self.q(css=self._bounded_selector(".staff__grade__control")).first.click()
+        self.wait_for_element_visibility("#staff__assessment__rubric__question--0__0", "staff grading is present")
+
+    def submissions_available(self):
+        found = self.q(
+            css=self._bounded_selector(".staff__grade__content")
+        )
+        if found.text[0] == "No more assessments can be graded at this time.":
```
cahrens: This text has changed on my branch.

test/acceptance/pages.py
```diff
@@ -491,6 +491,19 @@ def show_learner(self, username):
         submit_button.first.click()
         self.wait_for_element_visibility(".staff-info__student__report", "Student report is present")
 
+    def expand_staff_grading_section(self):
```
cahrens: Please add documentation for the new methods.

test/acceptance/tests.py
```diff
@@ -66,9 +66,12 @@ class OpenAssessmentTest(WebAppTest):
         'file_upload':
             u'courses/{test_course_id}/courseware/'
             u'57a3f9d51d424f6cb922f0d69cba868d/bb563abc989340d8806920902f267ca3/'.format(test_course_id=TEST_COURSE_ID),
-        'full_workflow':
+        'full_workflow_override':
```
cahrens: Make this full_workflow_staff_override and full_workflow_staff_required so it is clear what the last word relates to.

test/acceptance/pages.py
```diff
@@ -557,11 +570,24 @@ def learner_response(self):
...
         """
         Submit a staff assessment of the problem.
         """
-        self.submit(button_css=".wrapper--staff-assessment .action--submit")
+        filter_text = "Submit Assessment"
+        if continue_after:
+            filter_text += " and Grade Another Learner"
+        self.q(css=self._bounded_selector("button.action--submit")).filter(text=filter_text).first.click()
 
     def cancel_submission(self):
```
cahrens: Can you add coverage for the "ungraded" and "checked out" numbers? That might be tricky because there can be extra submissions in the system, so we might just need to get the numbers from the start of the test and increment/decrement from that point on.

test/acceptance/tests.py
```diff
@@ -762,25 +783,28 @@ def verify_submission_has_peer_grade(self, learner, max_attempts=5):
             max_attempts: the maximum number of times an additional peer grading should be done
         """
         count = 0
-        while self.grade_page.wait_for_page().score is None and count < (max_attempts + 1):
+        peer_page = AssessmentPage("peer-assessment", self.browser, self.problem_loc)
```
cahrens: Is self.peer_page pointing to the wrong thing?

test/acceptance/tests.py
```diff
@@ -762,25 +783,28 @@ def verify_submission_has_peer_grade(self, learner, max_attempts=5):
...
+class FullWorkflowBaseTest(OpenAssessmentTest, FullWorkflowMixin):
     """
-    Tests of complete workflows, combining multiple required steps together.
+    Base class for common functionality in full workflow tests.
     """
-    def setUp(self):
-        super(FullWorkflowTest, self).setUp('full_workflow', staff=True)
+
+    # Note: this will cause a KeyError if not called from a child class
+    def setUp(self, problem_loc_key="override_plz"):
```
cahrens: Can we not pass problem_type like in OpenAssessmentTest (and why do we need to provide a default value)?

test/acceptance/tests.py
```diff
@@ -875,6 +913,8 @@ def test_staff_override_at_beginning(self):
         And I see my staff override score
         And all fields in the staff area tool are correct
         """
+        # Note: this test skips self.do_train_self_peer(), as it needs more fine-grained validation and control.
```
cahrens: I don't understand this comment-- I intentionally don't want to do the training, self, and peer steps at the beginning of the test.

test/acceptance/tests.py
```diff
@@ -923,6 +963,45 @@ def test_staff_override_at_beginning(self):
...
+
+    @retry()
+    @attr('acceptance')
+    def test_train_self_peer_staff(self):
+        """
+        Scenario: complete workflow that included staff required step.
+
+        Given that I have created a submission, completed training, and done a self assessment
+        And a second learner has also created a submission, training, and self assessment
+        Then I can assess a learner
```
cahrens: These steps aren't correct-- no other learner is assessing this one.

I would recommend 2 different test cases, similar to what was done in the previous "full workflow" tests. In one case, staff grading happens early and peer grading doesn't have to happen to me in order for me to see a score. This is what you have here, though you need to update the BDD description.

In the second case, staff grading happens at the end, after all required steps have been done (including assessment of me by peers). We want to verify that a final grade still doesn't show, until the staff grade is done.

We also need coverage of "Submit and Grade Another Learner" functionality. Although your helper method is written to support that, I don't think any of the tests are guaranteed to actually go through that workflow (you aren't creating multiple submissions in the test case).

As mentioned above, we should have coverage on the ungraded and checked out numbers.

And finally, an accessibility test should be added with the staff grading form in the expanded state.

test/acceptance/tests.py
```diff
@@ -923,6 +963,45 @@ def test_staff_override_at_beginning(self):
...
+        self.do_staff_assessment(options_selected=self.STAFF_OVERRIDE_OPTIONS_SELECTED)
+
+        # At this point, the learner sees the score (1).
+        self.browser.refresh()
+        self._verify_staff_grade_section(self.STAFF_GRADE_EXISTS, None)
+        self.assertEqual(self.STAFF_OVERRIDE_SCORE, self.grade_page.wait_for_page().score)
+
+        # Note that PEER ASSESSMENT isn't shown here - it gets cut off the page in this case
+        self.verify_grade_entries([
+            [(u"STAFF GRADE - 0 POINTS", u"Poor"), (u"STAFF GRADE - 1 POINT", u"Fair")],
```
cahrens: Add a comment with the bug number.

cahrens:
Please add the test course to this PR (you said it needs to be updated, though I didn't understand the reason).

cahrens:
I wonder if we should add a third test point that does a staff override after a staff grade has been assigned (combining full staff grading and staff override).

test/acceptance/tests.py
```diff
@@ -212,10 +215,32 @@ def do_staff_override(self, username, final_score=STAFF_AREA_SCORE.format(STAFF_
...
+        self.staff_area_page.visit()
+        self.staff_area_page.expand_staff_grading_section()
+        start_numbers = self.staff_area_page.available_checked_out_numbers
+        assessed = 0
+        while number_to_assess == 0 or assessed < number_to_assess:
+            continue_after = False if number_to_assess-1 == assessed else True
+            self.staff_area_page.staff_assess(options_selected, continue_after)
+            assessed += 1
+            new_numbers = (start_numbers[0]-assessed, start_numbers[1])
+            self.assertEqual(new_numbers, self.staff_area_page.available_checked_out_numbers)
```
efischer19: TODO: this isn't waiting long enough - I can see the correct numbers appear while the test is running, but it's failing with not-yet-updated numbers being reported.

test/acceptance/accessibility.py
```diff
@@ -98,10 +98,10 @@ class StaffAreaA11yTest(OpenAssessmentA11yTest):
     """
     Test the accessibility of the staff area.
 
-    This is testing a problem with "self assessment only".
+    This is testing a problem with "staff assessment only".
```
efischer19: As I was writing the a11y test, I realized that this flow becomes a lot simpler if we use `staff-only` instead of `self-only` with override.

test/acceptance/tests.py
```diff
@@ -35,7 +35,7 @@ def _inner(*args, **kwargs):
                     return func(*args, **kwargs)
                 except (BrokenPromise, AssertionError) as ex:
                     if attempt_num >= (tries - 1):
-                        raise ex
+                        raise
```
efischer19: This will preserve the stack trace on failed test runs, which was extra helpful in writing these

test/acceptance/tests.py
```diff
@@ -212,10 +215,33 @@ def do_staff_override(self, username, final_score=STAFF_AREA_SCORE.format(STAFF_
...
+                will grade all available submissions.
+        """
+        self.staff_area_page.visit()
+        self.staff_area_page.expand_staff_grading_section()
+        start_numbers = self.staff_area_page.available_checked_out_numbers
+        assessed = 0
+        while number_to_assess == 0 or assessed < number_to_assess:
+            continue_after = False if number_to_assess-1 == assessed else True
+            self.staff_area_page.staff_assess(options_selected, continue_after)
+            self.staff_area_page.wait_for_ajax()
```
efischer19: Without this `wait_for_ajax` call, the numbers aren't refreshed quickly enough and the assertion below would fail every time.

test/acceptance/tests.py
```diff
@@ -275,15 +301,19 @@ def test_staff_assessment(self):
...
+        # Verify staff grade section appears as expected
         self.staff_asmnt_page.visit()
-        self.staff_asmnt_page.verify_status_value("COMPLETE")
+        self._verify_staff_grade_section(self.STAFF_GRADE_EXISTS, None)
+        self.assertEqual(self.EXPECTED_SCORE, self.grade_page.wait_for_page().score)
 
+        # Verify that staff scores can be overriden
+        self.do_staff_override(username)
+        self.browser.refresh()
+        self.assertEqual(self.STAFF_OVERRIDE_SCORE, self.grade_page.wait_for_page().score)
```
efischer19: It seemed simplest to just add this check on as another assertion on the simple `test_staff_assessment`, but it could also get it's own test method if we think that's necessary.

test/acceptance/tests.py
efischer19: The other learner is grading this one in the `do_self_train_peer` helper method below.

I've added ddt booleans to this test, they will determine if the other learner actually grades this one or not. This will cover both of the assertions you mentioned:
* peer grading doesn't have to happen to me in order for me to see a score
* a final grade doesn't show until staff grade is done

I've also added some assertions to the `do_staff_assessment` method to make sure the ungraded and checked out numbers are correct, and that "submit and grade another" is working. This should take effect on this test, as we have 2 learners creating a submission.

efischer19:
The bok choy tests are all passing locally

The accessibility tests are failing with this error:
```
AccessibilityError: URL 'https://staff:edx@efischer19.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/61944efb38a349edb140c762c7419b50/415c3ee1b7d04b58a1887a6fe82b31d6/' has 2 errors:

Rule ID: duplicate-id
Help URL: https://dequeuniversity.com/rules/axe/1.1/duplicate-id

	Message: Document has multiple elements with the same id attribute: section-parent
	Html: <a href="#section-child" role="button" class="button-chapter chapter" id="section-parent"
		aria-controls="section-child" aria-expanded="false">
	Target: [u'#content > .container > .course-wrapper > .course-index > .accordion > .course-
		navigation > a:nth-of-type(8)']


	Message: Document has multiple elements with the same id attribute: section-child
	Html: <div class="chapter-content-container" id="section-child" tabindex="-1" role="group" aria-
		label="Section submenu">
	Target: [u'#content > .container > .course-wrapper > .course-index > .accordion > .course-
		navigation > div:nth-of-type(8)']
```

efischer19:
@cahrens @andy-armstrong This is ready for another round of review. I've addressed the comments from before.

I'm going to look into the a11y failure, I think it's an actual accessibility issue though and not the test itself.

efischer19:
sigh, I found out why the a11y tests are failing - there are 2 different sections named `Section`, so naturally there are multiple id's for `section-child` and `section-parent`. I'll work on getting that cleaned up and update the attached course file.

screen shot 2015-12-22 at 9 37 13 am

efischer19:
Both `Section` sections have been removed, and all tests (a11y, acceptance, unit, and jasmine) are passing.

Latest course is available at https://efischer19.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/, or attached to this PR

test/acceptance/accessibility.py
```diff
@@ -120,12 +120,26 @@ def test_staff_info_panel(self):
...
+    def test_staff_grading_panel(self):
+        """
+        Check the accessibility of the "Staff Grading" panel
+        """
+        self.submission_page.visit().submit_response(self.SUBMISSION)
+        self.assertTrue(self.submission_page.has_submitted)
+
+        self.staff_area_page.visit()
+        self.staff_area_page.click_staff_toolbar_button("staff-grading")
+        self._check_a11y(self.staff_area_page)
```
cahrens: Do you also have a test where you click "Staff Assessment" to expand the form?

efischer19: Done, it will now expand that section as part of this test and check the whole thing in one fell swoop

test/acceptance/tests.py
cahrens: Instead of doing a wait_for_ajax, wait for  the numbers to be what you expect. wait_for_ajax often leads to flaky test failures.

cahrens:
:+1: after you change the wait_for_ajax.

efischer19:
@cahrens Got it, thanks for the tip.

dianakhuang:
:+1: from me.

efischer19:
Thanks @dianakhuang!

@cahrens - I'll leave this unmerged for now, so we don't clog up your in-flight PR. Could you merge this into your branch after you get thumbs but before merging into `ora-staff-grading`?

PR #823, opened 2015-12-17
Minor doc tweak

efischer19:
Per offline conversation with @catong, changing some wording in the
user visible string for editing ORA assessment steps. This new wording
is more accurate, and matches the error messages shown when violating
the condition.


efischer19:
For posterity: Note that the variable we check when enforcing this is name `is_released`, and that's the language used in the errors thrown here as well.

https://github.com/edx/edx-ora2/blob/efischer/doc_tweaks/openassessment/xblock/validation.py#L166

catong:
@efischer19 Awesome! Thanks for your help with this. :+1:

PR #827, opened 2015-12-24
Update bok choy tests for recent changes.

cahrens:
The refresh model in the staff grading form changed.
Also adds a "scrollToTop" call when continuing grading.

@andy-armstrong and @efischer19 review at your leisure.

Note that I still had to temporarily remove a test case due to https://openedx.atlassian.net/browse/TNL-3957 (I believe that the previously created https://openedx.atlassian.net/browse/TNL-3930 was a red herring).


openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -471,6 +471,9 @@
                 view.baseView.unsavedWarningEnabled(false, view.FULL_GRADE_UNSAVED_WARNING_KEY);
                 view.staffGradeFormLoaded = false;
                 view.loadStaffGradeForm(!continueGrading);
+                if (continueGrading) {
```
cahrens: Note to reviewers: without scrolling up, it's not at all obvious what happened when you press "Submit assessment and continue grading". The problem was introduced in my last PR, where I changed the refresh model so we aren't re-rendering the entire staff area.

test/acceptance/pages.py
```diff
@@ -512,8 +511,6 @@ def available_checked_out_numbers(self):
         Gets "N available and M checked out" information from staff grading sections.
         Returns tuple of (N, M)
         """
-        if not 'GRADE AVAILABLE RESPONSES' in self.selected_button_names:
```
cahrens: Note to reviewers: I wanted to be able to get the counts before we expand the grading form. Therefore the test points explicitly call self.click_staff_toolbar_button("staff-grading") when necessary.

test/acceptance/pages.py
```diff
@@ -531,17 +528,6 @@ def verify_available_checked_out_numbers(self, expected_value):
             "Expected avaiable and checked out values present"
         ).fulfill()
 
-    def submissions_available(self):
```
cahrens: Using this method was problematic because the text takes awhile to appear (and thus this method can incorrectly "pass").

test/acceptance/tests.py
```diff
@@ -228,17 +227,38 @@ def do_staff_assessment(self, number_to_assess=0, options_selected=OPTIONS_SELEC
                 will grade all available submissions.
         """
         self.staff_area_page.visit()
-        self.staff_area_page.expand_staff_grading_section()
+        self.staff_area_page.click_staff_toolbar_button("staff-grading")
+        # Get the counts before checking out a submission for assessment.
         start_numbers = self.staff_area_page.available_checked_out_numbers
+        # Check out a submission.
+        self.staff_area_page.expand_staff_grading_section()
+        # Checked out number should increase, ungraded decrease.
```
cahrens: As part of my last PR, the ungraded/checked out numbers update when you check out an assessment for grading.

test/acceptance/tests.py
```diff
@@ -228,17 +227,38 @@ def do_staff_assessment(self, number_to_assess=0, options_selected=OPTIONS_SELEC
...
                 break
+            else:
+                ungraded -=1
+                self.staff_area_page.verify_available_checked_out_numbers((ungraded, checked_out))
+
+    def refresh_page(self):
+        """
+        Helper method that waits for "unsaved changes" warnings to clear before refereshing the page.
+        """
+        EmptyPromise(
```
cahrens: More handling of "unsaved changes" was added in my last PR. The clearing of the warnings is asynchronous, and the bok choy tests previously attempted to reload the page immediately after pressing the submit button(s).

test/acceptance/tests.py
```diff
@@ -794,8 +814,15 @@ def verify_submission_has_peer_grade(self, learner, max_attempts=5):
...
             max_attempts: the maximum number of times an additional peer grading should be done
         """
+        def peer_grade_exists():
+            self.staff_area_page.visit()
+            self.staff_area_page.show_learner(learner)
+            self.staff_area_page.expand_learner_report_sections()
+            return "Peer Assessments for This Learner" in self.staff_area_page.learner_report_sections
+
         count = 0
-        while not self.peer_asmnt_page.is_complete and count < (max_attempts + 1):
```
cahrens: self.peer_asmnt_page.is_complete was the wrong thing to check here. It verifies that the learner has submitted required peer assessments, not that someone has assessed them.

Previously this code checked "self.grade_page.wait_for_page().score is not None", which worked well for the case of staff override because the learner would get a score as soon as peer assessments were complete (staff would later override that score). However, @efischer19 changed the check in a previous PR to re-use this method for the case of full staff grading (in which the learner does NOT get a grade immediately when peer grading is complete).

What this method really needs to do is verify that peer assessment of this learner is complete. This is tricky because there is no LMS indication of this (or at least due to TNL-3957 it is not clear what that indication would be). So instead I relied on using staff tools to see if peer assessments exist for the learner.

test/acceptance/tests.py
```diff
@@ -987,8 +1015,8 @@ def setUp(self):
 
     @retry()
     @attr('acceptance')
-    @ddt.data(True, False)
-    def test_train_self_peer_staff(self, peer_grades_me):
+    @ddt.data(False)
+    def test_train_self_peer_staff(self, peer_grades_me):  # TODO: can't run with "True" due to TNL-3957
```
cahrens: The "True" case wasn't actually covering what it was supposed to. The bug in verify_submission_has_peer_grade meant that the learner may not have been graded by peers. But due to TNL-3957, there is no way to tell if a learner has been graded by peers in the full grading workflow.

test/acceptance/tests.py
```diff
@@ -228,17 +227,36 @@ def do_staff_assessment(self, number_to_assess=0, options_selected=OPTIONS_SELEC
...
+        # Check out a submission.
+        self.staff_area_page.expand_staff_grading_section()
+        # Checked out number should increase, ungraded decrease.
+        ungraded = start_numbers[0]-1
+        checked_out = start_numbers[1]+1
+        self.staff_area_page.verify_available_checked_out_numbers((ungraded, checked_out))
         assessed = 0
         while number_to_assess == 0 or assessed < number_to_assess:
-            continue_after = False if number_to_assess-1 == assessed else True
+            continue_after = False if number_to_assess-1 == assessed else ungraded > 0
```
efischer19: This is easier to understand than before, I like it.

efischer19:
Looks good to me on first review, in the process of spinning up a sandbox to run the tests and be sure they pass locally.

test/acceptance/tests.py
```diff
@@ -794,16 +812,20 @@ def verify_submission_has_peer_grade(self, learner, max_attempts=5):
...
             max_attempts: the maximum number of times an additional peer grading should be done
         """
+        def peer_grade_exists():
+            self.staff_area_page.visit()
+            self.staff_area_page.show_learner(learner)
+            return "Peer Assessments for This Learner" in self.staff_area_page.learner_report_sections
+
         count = 0
-        while not self.peer_asmnt_page.is_complete and count < (max_attempts + 1):
+        while not peer_grade_exists() and count < max_attempts:
```
cahrens: "self.peer_asmnt_page.is_complete" is the wrong thing to check here. It verifies that the learner has submitted required peer assessments, not that someone has assessed them.

Previously this code checked "self.grade_page.wait_for_page().score is not None", which worked well for the case of staff override because the learner would get a score as soon as peer assessments were complete (staff would later override that score). However, @efischer19 changed the check in a previous PR to re-use this method for the case of full staff grading (in which the learner does NOT get a grade immediately when peer grading is complete).

What this method really needs to do is verify that peer assessment of this learner is complete. This is tricky because there is no LMS indication of this in the full grading case (or at least, due to TNL-3957, it is not clear what that indication should be). So instead I relied on using staff tools to see if peer assessments exist for the learner.

test/acceptance/tests.py
```diff
@@ -228,17 +227,36 @@ def do_staff_assessment(self, number_to_assess=0, options_selected=OPTIONS_SELEC
...
+            if not continue_after:
+                self.staff_area_page.verify_available_checked_out_numbers((ungraded, checked_out-1))
                 break
+            else:
+                ungraded -=1
+                self.staff_area_page.verify_available_checked_out_numbers((ungraded, checked_out))
+
+    def refresh_page(self):
+        """
+        Helper method that waits for "unsaved changes" warnings to clear before refereshing the page.
```
andy-armstrong: Typo: refereshing

andy-armstrong:
:+1: Looks great

efischer19:
:+1: confirmed passing locally.

I see some failures in the latest Jenkins run though, and I'm not sure they're all related to these changes. Does the test course on ora2.sandbox.edx.org need to be updated as well?

http://jenkins.edx.org:8080/job/ora2-acceptance-tests/606/

cahrens:
The Jenkins jobs are running the master branch, so they are not impacted by anything happening on ora-staff-grading.

PR #829, opened 2016-01-04
Staff Grading Events

dianakhuang:
## [TNL-3476](https://openedx.atlassian.net/browse/TNL-3476)

Add new staff grading events and new eventing helpers.

The event design can be found here: https://openedx.atlassian.net/wiki/display/AN/ORA+Staff+Grading+Event+Design

Note: we are only emitting `get_submission_for_staff_grading` in the case of full-grading.

### Testing
- [x] Analytics

### Reviewers

FYI: @catong 

### Post-review
- [x] Squash commits

### DIFF
--- a/openassessment/xblock/staff_area_mixin.py
+++ b/openassessment/xblock/staff_area_mixin.py
@@ -266,6 +266,13 @@ class StaffAreaMixin(object):
             submission_to_assess = staff_api.get_submission_to_assess(course_id, item_id, staff_id)
 
             if submission_to_assess is not None:
+                # This is posting a tracking event to the runtime.
+                self.runtime.publish(self, 'openassessmentblock.get_submission_for_staff_grading', {
+                    'type': 'full-grade',
+                    'requesting_staff_id': staff_id,
+                    'item_id': item_id,
+                    'submission_returned_uuid': submission_to_assess['uuid']
+                })
                 submission = submission_api.get_submission_and_student(submission_to_assess['uuid'])
                 if submission:
                     anonymous_student_id = submission['student_item']['student_id']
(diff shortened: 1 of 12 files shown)


efischer19:
Seems simple enough, :+1: from me.

I've added a comment to the event design page - the current design is slightly different than what's been implemented, we ought to ensure that documentation is consistent.

openassessment/xblock/staff_area_mixin.py
```diff
@@ -266,6 +266,12 @@ def render_staff_grade_form(self, data, suffix=''):  # pylint: disable=W0613
             submission_to_assess = staff_api.get_submission_to_assess(course_id, item_id, staff_id)
 
             if submission_to_assess is not None:
+                self.runtime.publish(self, 'openassessmentblock.get_submission_for_staff_grading', {
```
andy-armstrong: Nit: I think this is poor naming by the XBlock API but the way this reads it seems to be the XBlock itself is being published. Maybe just add a comment to say that this is publishing an event.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -452,7 +452,7 @@
                 // the staff override itself.
                 view.loadStudentInfo({expanded_view: 'final-grade'});
             };
-            this.callStaffAssess(submissionID, rubric, scope, successCallback, '.staff-override-error');
+            this.callStaffAssess(submissionID, rubric, scope, successCallback, '.staff-override-error', 'regrade');
```
andy-armstrong: I don't see any unit testing in Python or JS for 'regrade' (the four occurrences in these changes are all in non-test code). It would be good to have a Jasmine test that verifies that the parameter is passed correctly to the server, and then to have a server test that verifies that the correct event is published.

andy-armstrong:
@dianakhuang Looks great. I'd like to see some unit testing of the regrade submission logic, and then :+1:

dianakhuang:
@andy-armstrong I've added some testing for the submission logic. Let me know if this is sufficient.

andy-armstrong:
Thanks @dianakhuang, the tests look good. :+1:

PR #764, opened 2015-11-09
Staff waiting display updates (on hold)

efischer19:
Changes pulled out, they can wait to be made for a while.

Will require doc team review.

PR #830, opened 2016-01-04
Show peer scores in full grade problem

efischer19:
## [TNL-3957](https://openedx.atlassian.net/browse/TNL-3957)

There are 2 main issues being fixed here:
First, the peer scores that are not used to calculate score are always hidden.
This is because the peer api tries to hide these values by default.
Within the peer api, `scored_only` seems to only be used in a handful of tests, 
and is never expected implicitly by the ORA2 frontend. Because of this I've
changed the default to be `False`, and updated the tests that expected
the old default behavior to explicitly use `True`.

Secondly, the display strings for waiting values are not yet updated.
I've replayed the changes initially proposed in https://github.com/edx/edx-ora2/pull/764,
with a simplification made (thanks @catong for the suggestion - https://github.com/edx/edx-ora2/pull/764#discussion_r45105717). @explorerleslie, is this an okay simplification to
make, or is it important that the display string explicitly list each step requirement? I'll
add a screenshot below.

### Sandbox
- [x] https://efischer19.sandbox.edx.org/

### Testing
- [x] Unit tests passing locally
- [x] Acceptance tests passing locally

### Reviewers
- [x] Code review: @cahrens 
- [x] Display strings review: @catong 
- [ ] Product review: @explorerleslie

### DIFF
--- a/openassessment/assessment/api/peer.py
+++ b/openassessment/assessment/api/peer.py
@@ -483,7 +483,7 @@ def has_finished_required_evaluating(submission_uuid, required_assessments):
     return done, peers_graded
 
 
-def get_assessments(submission_uuid, scored_only=True, limit=None):
+def get_assessments(submission_uuid, limit=None):
     """Retrieve the assessments for a submission.
 
     Retrieves all the assessments for a submissions. This API returns related
@@ -495,9 +495,6 @@ def get_assessments(submission_uuid, scored_only=True, limit=None):
             associated with. Required.
 
     Keyword Arguments:
-        scored (boolean): Only retrieve the assessments used to generate a score
-            for this submission.
-
         limit (int): Limit the returned assessments. If None, returns all.
 
 
@@ -513,7 +510,7 @@ def get_assessments(submission_uuid, scored_only=True, limit=None):
             while retrieving the assessments associated with this submission.
 
     Examples:
-        >>> get_assessments("1", scored_only=True, limit=2)
+        >>> get_assessments("1", limit=2)
         [
             {
                 'points_earned': 6,
@@ -533,15 +530,10 @@ def get_assessments(submission_uuid, scored_only=True, limit=None):
 
     """
     try:
-        if scored_only:
-            assessments = PeerWorkflowItem.get_scored_assessments(
-                submission_uuid
-            )[:limit]
-        else:
-            assessments = Assessment.objects.filter(
-                submission_uuid=submission_uuid,
-                score_type=PEER_TYPE
-            )[:limit]
+        assessments = Assessment.objects.filter(
+            submission_uuid=submission_uuid,
+            score_type=PEER_TYPE
+        )[:limit]
         return serialize_assessments(assessments)
     except DatabaseError:
         error_message = (
@@ -551,7 +543,7 @@ def get_assessments(submission_uuid, scored_only=True, limit=None):
         raise PeerAssessmentInternalError(error_message)
 
 
-def get_submitted_assessments(submission_uuid, scored_only=True, limit=None):
+def get_submitted_assessments(submission_uuid, limit=None):
     """Retrieve the assessments created by the given submission's author.
 
     Retrieves all the assessments created by the given submission's author. This
@@ -564,8 +556,6 @@ def get_submitted_assessments(submission_uuid, scored_only=True, limit=None):
         we are requesting. Required.
 
     Keyword Arguments:
-        scored (boolean): Only retrieve the assessments used to generate a score
-            for this submission.
         limit (int): Limit the returned assessments. If None, returns all.
 
     Returns:
@@ -581,7 +571,7 @@ def get_submitted_assessments(submission_uuid, scored_only=True, limit=None):
             while retrieving the assessments associated with this submission.
 
     Examples:
-        >>> get_submitted_assessments("1", scored_only=True, limit=2)
+        >>> get_submitted_assessments("1", limit=2)
         [
             {
                 'points_earned': 6,
@@ -608,8 +598,6 @@ def get_submitted_assessments(submission_uuid, scored_only=True, limit=None):
             scorer=workflow,
             assessment__isnull=False
         )
-        if scored_only:
-            items = items.exclude(scored=False)
         assessments = Assessment.objects.filter(
             pk__in=[item.assessment.pk for item in items])[:limit]
         return serialize_assessments(assessments)
--- a/openassessment/assessment/test/test_peer.py
+++ b/openassessment/assessment/test/test_peer.py
@@ -104,7 +104,7 @@ ASSESSMENT_DICT_FAIL = {
     }
 }
 
-# Answers are against RUBRIC_DICT -- this is worth 12 points
+# Answers are against RUBRIC_DICT -- this is worth 14 points
 ASSESSMENT_DICT_PASS = {
     'overall_feedback': u"这是中国",
     'criterion_feedback': {},
@@ -116,7 +116,7 @@ ASSESSMENT_DICT_PASS = {
     }
 }
 
-# Answers are against RUBRIC_DICT -- this is worth 12 points
+# Answers are against RUBRIC_DICT -- this is worth 14 points
 # Feedback text is one character over the limit.
 LONG_FEEDBACK_TEXT = u"是" * Assessment.MAX_FEEDBACK_SIZE + "."
 ASSESSMENT_DICT_HUGE = {
@@ -322,7 +322,7 @@ class TestPeerApi(CacheResetTest):
             RUBRIC_DICT,
             REQUIRED_GRADED_BY,
         )
-        assessments = peer_api.get_assessments(sub["uuid"], scored_only=False)
+        assessments = peer_api.get_assessments(sub["uuid"])
         self.assertEqual(1, len(assessments))
 
     @file_data('data/valid_assessments.json')
@@ -340,7 +340,7 @@ class TestPeerApi(CacheResetTest):
             REQUIRED_GRADED_BY,
             MONDAY,
         )
-        assessments = peer_api.get_assessments(sub["uuid"], scored_only=False)
+        assessments = peer_api.get_assessments(sub["uuid"])
         self.assertEqual(1, len(assessments))
         self.assertEqual(assessments[0]["scored_at"], MONDAY)
 
@@ -859,14 +859,12 @@ class TestPeerApi(CacheResetTest):
         )
         self.assertEqual(assessment["points_earned"], 6)
         self.assertEqual(assessment["points_possible"], 14)
-        submitted_assessments = peer_api.get_submitted_assessments(bob_sub["uuid"], scored_only=True)
-        self.assertEqual(0, len(submitted_assessments))
 
-        submitted_assessments = peer_api.get_submitted_assessments(bob_sub["uuid"], scored_only=False)
+        submitted_assessments = peer_api.get_submitted_assessments(bob_sub["uuid"])
         self.assertEqual(1, len(submitted_assessments))
 
     def test_get_submitted_assessments_with_bad_submission(self):
-        submitted_assessments = peer_api.get_submitted_assessments("bad-uuid", scored_only=True)
+        submitted_assessments = peer_api.get_submitted_assessments("bad-uuid")
         self.assertEqual(0, len(submitted_assessments))
 
     def test_find_active_assessments(self):
@@ -1122,7 +1120,7 @@ class TestPeerApi(CacheResetTest):
         bob_sub, bob = self._create_student_and_submission("Bob", "Bob's answer")
         peer_api.get_submission_to_assess(bob_sub['uuid'], REQUIRED_GRADED_BY)
         mock_filter.side_effect = DatabaseError("Oh no.")
-        submitted_assessments = peer_api.get_submitted_assessments(bob_sub["uuid"], scored_only=False)
+        submitted_assessments = peer_api.get_submitted_assessments(bob_sub["uuid"])
         self.assertEqual(1, len(submitted_assessments))
 
     @patch.object(PeerWorkflow.objects, 'raw')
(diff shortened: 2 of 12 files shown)


openassessment/assessment/api/peer.py
```diff
@@ -565,7 +565,7 @@ def get_submitted_assessments(submission_uuid, scored_only=True, limit=None):
 
     Keyword Arguments:
         scored (boolean): Only retrieve the assessments used to generate a score
-            for this submission.
+            for their respective submission.
```
efischer19: This seemed to be incorrectly copied-but-never-updated documentation, let me know if this fix is incorrect.

openassessment/templates/openassessmentblock/grade/oa_grade_waiting.html
```diff
@@ -16,13 +16,7 @@ <h2 class="step__title">
...
             <div class="step__content">
                 <div class="grade__value__description">
-                    {% if waiting == 'peer' %}
-                        <p>{% trans "Your response is still undergoing peer assessment. After your peers have assessed your response, you'll see their comments and receive your final grade." %}</p>
-                    {% elif waiting == 'example-based' %}
-                        <p>{% trans "Your response is still undergoing example-based assessment. After your response has been assessed, you'll see the comments and receive your final grade." %}</p>
-                    {% elif waiting == 'all' %}
-                        <p>{% trans "Your response is still undergoing peer assessment and example-based assessment. After your peers have assessed your response and example-based assessment is complete, you'll see your peers' comments and receive your final grade." %}</p>
-                    {% endif %}
+                    <p>{% trans "Your response still needs to undergo some assessment steps. After all assessment steps are complete, you will see comments from peers or course staff and receive your final grade." %}</p>
```
efischer19: @explorerleslie This is the suggestion that @catong made in https://github.com/edx/edx-ora2/pull/764. Here is a screenshot, the user has finished training, self assessment, and their peer assessment requirements, but no peers or staff have graded the user.

screen shot 2016-01-04 at 4 24 40 pm

Because we're trying to avoid a crazy amount of strings to enumerate, the display does not change at all after peer grades for the user are made, since we're still waiting on staff assessment.

openassessment/xblock/openassessmentblock.py
```diff
@@ -809,16 +810,17 @@ def get_waiting_details(self, status_details):
...
+        modules = [('peer', 'peer'), ('ai', 'example-based'), ('staff', 'staff')]
+        for module in modules:
+            if module[0] in status_details and not status_details[module[0]]['graded']:
+                waiting.append(module[1])
+        num_details = len(waiting)
+        if num_details == 0:
+            return None
+        elif num_details == len(modules):
+            return "all"
+        return ','.join(waiting)
```
efischer19: This block is code I wrote in the previous PR. If we're no longer differentiating between the different waiting states in the html template, do we really need to do all of this, or just return a bool indicating whether we're in a waiting state?

openassessment/xblock/test/test_message.py
```diff
@@ -38,6 +38,10 @@ class TestMessageRender(XBlockHandlerTestCase):
             'completed': TODAY,
             'graded': TODAY,
         },
+        'staff': {
+            'completed': TODAY,
+            'graded': TODAY,
+        },
```
efischer19: I ought to add a test case to this file to verify that waiting for staff yields the correct message as well.

efischer19:
Confirming the other issue mentioned in the TNL ticket - the learner's peer assessment scores do now appear in the `Manage Individual Learners` dialog, Here's a screenshot:

screen shot 2016-01-04 at 4 50 44 pm

cahrens:
@efischer19 you should update the acceptance tests. There are at least two places that need to be fixed--

https://github.com/edx/edx-ora2/blob/ora-staff-grading/test/acceptance/tests.py#L1014

https://github.com/edx/edx-ora2/blob/ora-staff-grading/test/acceptance/tests.py#L1044

openassessment/assessment/api/peer.py
cahrens: Argument name is "scored_only". I'm not sure about the doc string-- is this method used? Is "scored_only" used? The whole thing is confusing since the doc says "this API returns related feedback without making any assumption about grading". 

I feel like there is much sketchiness with this method, and I'm not sure it is worth trying to fix the doc without understanding the method.

openassessment/assessment/api/peer.py
```diff
@@ -483,7 +483,7 @@ def has_finished_required_evaluating(submission_uuid, required_assessments):
     return done, peers_graded
 
 
-def get_assessments(submission_uuid, scored_only=True, limit=None):
+def get_assessments(submission_uuid, scored_only=False, limit=None):
```
cahrens: Similar to my comment below, is get_assessments ever used with scored_only=True outside of tests? If not, should we just delete the argument from the signature?

openassessment/templates/openassessmentblock/grade/oa_grade_waiting.html
cahrens: I like the general idea, but do not like the wording. How about

"All assessment steps have not been completed for your response. After all assessment steps are complete, you will see feedback from everyone who assessed your response, and you will receive your final grade."

Note that it is possible that it is possible that neither peers nor course staff are required (it could be self or AI grading). And I don't think we should call out course staff as a grading possibility.

openassessment/xblock/openassessmentblock.py
cahrens: If we don't need the complexity, definitely simplify.

openassessment/xblock/test/test_message.py
cahrens: Yes, sounds like a good idea.

openassessment/assessment/test/test_peer.py
```diff
@@ -104,7 +104,7 @@
     }
 }
 
-# Answers are against RUBRIC_DICT -- this is worth 12 points
+# Answers are against RUBRIC_DICT -- this is worth 14 points
```
efischer19: I have no idea why this was incorrectly documented initially, but comparing `options_selected` to `RUBRIC_DICT` above yields 1+1+10+2=14 points, so I updated the documentation here to match.

openassessment/assessment/test/test_peer.py
```diff
@@ -859,14 +859,12 @@ def test_get_submitted_assessments(self):
         )
         self.assertEqual(assessment["points_earned"], 6)
         self.assertEqual(assessment["points_possible"], 14)
-        submitted_assessments = peer_api.get_submitted_assessments(bob_sub["uuid"], scored_only=True)
-        self.assertEqual(0, len(submitted_assessments))
```
efischer19: Since the only purpose of this assertion seemed to be testing the functionality of the not-used-anywhere-else `scored_only` argument, I removed it.

openassessment/assessment/test/test_peer.py
```diff
@@ -1253,11 +1251,11 @@ def test_median_score_db_error(self, mock_filter):
         tim, _ = self._create_student_and_submission("Tim", "Tim's answer")
         peer_api.get_assessment_median_scores(tim["uuid"])
 
-    @patch.object(PeerWorkflowItem, 'get_scored_assessments')
+    @patch.object(Assessment.objects, 'filter')
     @raises(peer_api.PeerAssessmentInternalError)
     def test_get_assessments_db_error(self, mock_filter):
-        mock_filter.side_effect = DatabaseError("Bad things happened")
         tim, _ = self._create_student_and_submission("Tim", "Tim's answer")
+        mock_filter.side_effect = DatabaseError("Bad things happened")
```
efischer19: `side_effect` has to be defined here, otherwise it will blow up in the above line instead of the one we want.

openassessment/assessment/test/test_peer.py
```diff
@@ -1506,27 +1504,24 @@ def test_too_many_assessments_counted_in_score_bug(self):
...
-        peer_api.get_score(bob_sub['uuid'], requirements)
+        score = peer_api.get_score(bob_sub['uuid'], requirements)
 
-        # Get the assessments used to generate the score
-        # Only the first assessment should be used
-        scored_assessments = peer_api.get_assessments(bob_sub['uuid'], scored_only=True)
-        self.assertEqual(len(scored_assessments), 1)
-        self.assertEqual(scored_assessments[0]['scorer_id'], tim['student_id'])
+        # Verify that only the first assessment was used to generate the score
+        self.assertEqual(score['points_earned'], 14)
```
efischer19: Instead of "get the assessment used for scoring, and make sure it was Tim's", I gave Tim and Sue's assessments of Bob different point values so we can ensure that the score is calculated correctly.

test/acceptance/tests.py
```diff
@@ -1041,10 +1041,10 @@ def test_train_self_peer_staff(self, peer_grades_me):  # TODO: can't run with "T
         self._verify_staff_grade_section(self.STAFF_GRADE_EXISTS, None)
         self.assertEqual(self.STAFF_OVERRIDE_SCORE, self.grade_page.wait_for_page().score)
 
-        # Note that PEER ASSESSMENT isn't shown here - it gets cut off the page in this case
-        # See TNL-3930 for details and possible fix.
+        #TODO - readding PEER ASSESSMENT, verify actual score values before merging
```
efischer19: in-flight TODO

openassessment/templates/openassessmentblock/grade/oa_grade_waiting.html
catong: @cahrens good point. I'd suggest a minor tweak to your first sentence; the rest sounds good.
"Not all assessment steps have been completed for your response."
@explorerleslie your thoughts?

openassessment/xblock/grade_mixin.py
```diff
@@ -108,6 +108,7 @@ def render_grade_complete(self, workflow):
         has_submitted_feedback = False
 
         if "peer-assessment" in assessment_steps:
+            peer_api.get_score(submission_uuid, self.workflow_requirements()["peer"])
```
efischer19: Adding this line allows for the relevant peer assessments to be marked as "scored" so that they show up in the final grade report.

They are not actually used in calculating the final score, but it is necessary to differentiate which assessments would have been used to calculate the final score when getting the median for the final grade report.

efischer19: See https://github.com/edx/edx-ora2/blob/efischer/show_peer_scores/openassessment/assessment/api/peer.py#L434 for where this filtering takes place.

efischer19:
`message_mixin.py`, along with all the templates under `openassessment/templates/openassessmentblock/message/` were only ever being used in `test_message.py` (See https://github.com/edx/edx-ora2/search?utf8=%E2%9C%93&q=render_message, a search for the primary method call).

I went ahead and removed all of that code, as it's handled elsewhere. For example, compare https://github.com/edx/edx-ora2/blob/master/openassessment/templates/openassessmentblock/grade/oa_grade_waiting.html#L19 with https://github.com/edx/edx-ora2/blob/master/openassessment/templates/openassessmentblock/message/oa_message_complete.html#L7.

Tests are still passing, I'm going to do some manual poking on my sandbox to make sure everything looks okay. Otherwise, this is ready for review again @cahrens @andy-armstrong.

andy-armstrong:
As we discussed together, the message_mixin is used in a subtle way from the client-side. It renders the message at the top of the problem:

https://github.com/edx/edx-ora2/pull/340

efischer19:
I've re-added all the `message_mixin` code that was previously deleted, and fixed it and its tests up as well. Ready for another look at this point, sandbox is updated and all tests are passing locally.

openassessment/templates/openassessmentblock/grade/oa_grade_waiting.html
```diff
@@ -16,13 +16,7 @@ <h2 class="step__title">
...
                 <div class="grade__value__description">
-                    {% if waiting == 'peer' %}
-                        <p>{% trans "Your response is still undergoing peer assessment. After your peers have assessed your response, you'll see their comments and receive your final grade." %}</p>
-                    {% elif waiting == 'example-based' %}
-                        <p>{% trans "Your response is still undergoing example-based assessment. After your response has been assessed, you'll see the comments and receive your final grade." %}</p>
-                    {% elif waiting == 'all' %}
-                        <p>{% trans "Your response is still undergoing peer assessment and example-based assessment. After your peers have assessed your response and example-based assessment is complete, you'll see your peers' comments and receive your final grade." %}</p>
-                    {% endif %}
+                    <p>{% trans "Not all assessment steps have been completed for your response. After all assessment steps are complete, you will see feedback from everyone who assessed your response, and you will receive your final grade." %}</p>
```
efischer19: @catong @explorerleslie @cahrens I copied the already-discussed waiting string here as well, should this be different based on context?

Screenshot, this particular line is defining the top instance of the string shown, the already-discussed change is defining the bottom one:
screen shot 2016-01-05 at 5 37 52 pm

openassessment/xblock/test/test_message.py
efischer19: Added via a ddt test case in `test_waiting_on_steps` below.

andy-armstrong:
:+1: Nice clean up

openassessment/xblock/openassessmentblock.py
```diff
@@ -778,21 +778,16 @@ def is_closed(self, step=None, course_staff=None):
     def get_waiting_details(self, status_details):
         """
         Returns the specific waiting status based on the given status_details.
```
cahrens: It no longer returns a "specific waiting status", but instead a boolean.

openassessment/templates/openassessmentblock/message/oa_message_complete.html
```diff
@@ -4,12 +4,8 @@
...
     <div class="message__content">
         <p>
-            {% if waiting == 'peer' %}
-                <p>{% trans "Your grade will be available when your peers have completed their assessments of your response." %}</p>
-            {% elif waiting == 'example-based' %}
-                <p>{% trans "Your grade will be available when the example-based assessment of your response is complete." %}</p>
-            {% elif waiting == 'all' %}
-                <p>{% trans "Your grade will be available when your peers have completed their assessments of your response and the example-based assessment of your response is complete." %}</p>
+            {% if waiting %}
+                <p>{% trans "Not all assessment steps have been completed for your response. After all assessment steps are complete, you will see feedback from everyone who assessed your response, and you will receive your final grade." %}</p>
```
cahrens: Duplicating this message seems a bit like an overkill. What about making this one simpler, like "Check back later to see your final grade"? Not a big deal, but I'm not sure why we need to show the same message twice.

FYI @catong 

image

cahrens:
:+1: though I think it is overkill to show the message twice, as I noted above (and actually, you noted it too!). I'd advocate drastically reducing the message in the oa_message_complete template.

openassessment/templates/openassessmentblock/message/oa_message_complete.html
catong: @cahrens How about: "Your final grade will be available when all assessment steps for your response have been completed."

cahrens: Like it, I do. :+1:

efischer19:
Thanks for the notes @cahrens and @catong!

I've updated my sandbox at https://efischer19.sandbox.edx.org, here's the latest screenshot:

screen shot 2016-01-06 at 11 51 42 am

@catong, I'll let you give the final thumbs on this before I merge.

catong:
Looks great! :+1:

PR #835, opened 2016-01-07
Staff can view uploaded files

efischer19:
Not sure if this was a bug or an oversight, but staff were not able to see uploaded files before.

Verified fix on my sandbox. @andy-armstrong @dianakhuang Can I get quick thumbs on this?

screen shot 2016-01-07 at 12 39 04 pm

### DIFF
--- a/openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
+++ b/openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
@@ -36,12 +36,8 @@
                         {% trans "The learner's response to the question above:" as translated_label %}
                         {% include "openassessmentblock/oa_submission_answer.html" with answer=submission.answer answer_text_label=translated_label %}
 
-                        {% if submission.file_url %}
-                            <a href="{{ submission.file_url }}" class="submission--file">
-                                {% trans "The file associated with this response." %}
-                            </a>
-                            <span>{%  trans "Caution: This file was uploaded by another course learner and has not been verified, screened, approved, reviewed, or endorsed by edX. If you decide to access it, you do so at your own risk." %}</span>
-                        {% endif %}
+                        {% trans "Associated File" as translated_header %}
+                        {% include "openassessmentblock/oa_uploaded_file.html" with file_upload_type=file_upload_type file_url=staff_file_url header=translated_header class_prefix="staff-assessment" show_warning="true" %}
                     </div>
                 {% endif %}
             </div>
--- a/openassessment/xblock/staff_area_mixin.py
+++ b/openassessment/xblock/staff_area_mixin.py
@@ -323,25 +323,15 @@ class StaffAreaMixin(object):
         Returns:
             A context dict for rendering a student submission and associated rubric (for staff grading).
         """
-        if submission and 'file_key' in submission.get('answer', {}):
-            file_key = submission['answer']['file_key']
-
-            try:
-                submission['file_url'] = file_api.get_download_url(file_key)
-            except file_exceptions.FileUploadError:
-                # Log the error, but do not prevent the rest of the student info
-                # from being displayed.
-                msg = (
-                    u"Could not retrieve image URL for staff debug page.  "
-                    u"The learner username is '{student_username}', and the file key is {file_key}"
-                ).format(student_username=student_username, file_key=file_key)
-                logger.exception(msg)
-
         context = {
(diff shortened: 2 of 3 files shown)


openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -35,12 +35,7 @@ <h2 class="staff-info__title ui-staff__subtitle ui-toggle-visibility__control">
                     <div class="wrapper--content">
                         {% include "openassessmentblock/oa_submission_answer.html" with answer=submission.answer answer_text_label="The learner's response to the question above:" %}
 
-                        {% if submission.file_url %}
-                            <a href="{{ submission.file_url }}" class="submission--file">
-                                {% trans "The file associated with this response." %}
-                            </a>
-                            <span>{%  trans "Caution: This file was uploaded by another course learner and has not been verified, screened, approved, reviewed, or endorsed by edX. If you decide to access it, you do so at your own risk." %}</span>
-                        {% endif %}
+                        {% include "openassessmentblock/oa_uploaded_file.html" with file_upload_type=file_upload_type file_url=staff_file_url header="staff" class_prefix="staff-assessment" show_warning="true" %}
```
efischer19: Ack, this is using the new functionality in https://github.com/edx/edx-ora2/pull/833, so the string here will read "STAFF" unless both are merged together. I can fix this to use the old style if needed, but I'd rather just merge both PRs in at around the same time.

openassessment/xblock/staff_area_mixin.py
```diff
@@ -343,6 +343,10 @@ def get_student_submission_context(self, student_username, submission):
             'student_username': student_username,
         }
 
+        if submission['file_url'] is not None:
+            context["staff_file_url"] = submission["file_url"]
```
andy-armstrong: Could this logic be inside the ``try`` block as it wouldn't then need the check? Could there be times when the submission comes in with the ``file_url`` already populated?

openassessment/xblock/staff_area_mixin.py
```diff
@@ -343,6 +343,10 @@ def get_student_submission_context(self, student_username, submission):
             'student_username': student_username,
         }
 
+        if submission['file_url'] is not None:
```
andy-armstrong: Nit: it seems a shame that every mixin has to know to add the ``file_upload_type``. This seems like a brittle way to have added the feature (I know that wasn't you). I wonder if there's some way to share this logic with a helper method that knows how to add file upload context to a submission.

efischer19: Definitely. This actually caught me when making the fix - I forgot to include `file_upload_type` initially and it took a few minutes to realize why it wasn't appearing.

I have a few ideas to clean this section up a bit, hang on.

openassessment/xblock/staff_area_mixin.py
```diff
@@ -323,26 +323,16 @@ def get_student_submission_context(self, student_username, submission):
...
-
         context = {
             'submission': create_submission_dict(submission, self.prompts) if submission else None,
             'rubric_criteria': copy.deepcopy(self.rubric_criteria_with_labels),
             'student_username': student_username,
         }
 
+        if submission:
+            context["file_upload_type"] = self.file_upload_type
+            context["staff_file_url"] = self.get_download_url_from_submission(submission)
```
efischer19: Using `self.get_download_url_from_submission(submission)` in place of the custom try block avoids duplicating that logic, at the cost of losing the specialized `student_username` logging. My thinking here is that the `for staff debug page.` line above indicates that this custom logging was more for debugging than usefulness.

The changed code now reflects more closely what the `peer_assessment_mixin` does: https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/peer_assessment_mixin.py#L238

dianakhuang:
Minus the test failure, this looks good. :+1:

efischer19:
Test has been fixed up, and `oa_student_info,html` now uses the translation approach decided on in https://github.com/edx/edx-ora2/pull/833. Ready for re-review at some point.

cahrens:
:+1:

openassessment/xblock/staff_area_mixin.py
andy-armstrong: Did you decide against doing anything with this?

andy-armstrong:
:+1:

PR #836, opened 2016-01-07
Updating reponse_submitted wording

efischer19:
I wish I could think of an easy way to simplify this and keep the links present, but easily sorting through a step being present vs. user-completed vs. graded by others gets hairy very quickly, so I just went with the string below. Thoughts @cahrens @catong?

Here's how it looked before, note that the staff step would never impact the message so it still looked like this after peer grades were made and all user requirements were met.
screen shot 2016-01-07 at 1 08 29 pm

### DIFF
--- a/openassessment/templates/openassessmentblock/response/oa_response_submitted.html
+++ b/openassessment/templates/openassessmentblock/response/oa_response_submitted.html
@@ -22,19 +22,7 @@
         <div class="step__message message message--complete">
             <h3 class="message__title">{% trans "Your Response Has Been Submitted" %}</h3>
             <div class="message__content">
-                {% if has_peer and has_self %}
-                    {% blocktrans with peer_start_tag='<a data-behavior="ui-scroll" href="#openassessment__peer-assessment">'|safe self_start_tag='<a data-behavior="ui-scroll" href="#openassessment__self-assessment">'|safe end_tag='</a>'|safe %}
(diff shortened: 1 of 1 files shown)


cahrens:
We could keep the steps that you nee

openassessment/templates/openassessmentblock/response/oa_response_submitted.html
```diff
@@ -22,19 +22,7 @@
...
-                {% elif has_peer %}
-                    {% blocktrans with start_tag='<a data-behavior="ui-scroll" href="#openassessment__peer-assessment">'|safe end_tag='</a>'|safe %}
-                        You'll receive your grade after some of your peers have assessed your response and you complete the {{ start_tag }}peer assessment{{ end_tag }} step.
-                    {% endblocktrans %}
-                {% elif has_self %}
-                    {% blocktrans with start_tag='<a data-behavior="ui-scroll" href="#openassessment__self-assessment">'|safe end_tag='</a>'|safe %}
-                        You'll receive your grade after you complete the {{ start_tag }}self assessment{{ end_tag }} step.
-                    {% endblocktrans %}
-                {% endif %}
+                {%trans You'll receive your grade after all assessment steps have been completed. %}
```
cahrens: Let's keep the part about what YOU have to do, and make the beginning part more generic.

efischer19:
PR was accidentally closed and then I force pushed, so now moving over to https://github.com/edx/edx-ora2/pull/837

PR #837, opened 2016-01-07
Fixing up oa_response

efischer19:
With staff now being a possibility, we need to update oa_response_complete.
Now keeping "Needs to be graded" and "You need to do things" separate.

Supersedes https://github.com/edx/edx-ora2/pull/836

@cahrens @catong

### DIFF
--- a/openassessment/templates/openassessmentblock/response/oa_response_submitted.html
+++ b/openassessment/templates/openassessmentblock/response/oa_response_submitted.html
@@ -22,17 +22,18 @@
         <div class="step__message message message--complete">
             <h3 class="message__title">{% trans "Your Response Has Been Submitted" %}</h3>
             <div class="message__content">
-                {% if has_peer and has_self %}
+                {% trans "You will receive your grade after all steps are complete and your response is fully assessed." %}
+                {% if peer_incomplete and self_incomplete %}
                     {% blocktrans with peer_start_tag='<a data-behavior="ui-scroll" href="#openassessment__peer-assessment">'|safe self_start_tag='<a data-behavior="ui-scroll" href="#openassessment__self-assessment">'|safe end_tag='</a>'|safe %}
-                        You'll receive your grade after some of your peers have assessed your response and you complete the {{ peer_start_tag }}peer assessment{{ end_tag }} and {{ self_start_tag }}self assessment{{ end_tag }} steps.
+                        You still need to complete the {{ peer_start_tag }}peer assessment{{ end_tag }} and {{ self_start_tag }}self assessment{{ end_tag }} steps.
                     {% endblocktrans %}
-                {% elif has_peer %}
+                {% elif peer_incomplete %}
                     {% blocktrans with start_tag='<a data-behavior="ui-scroll" href="#openassessment__peer-assessment">'|safe end_tag='</a>'|safe %}
-                        You'll receive your grade after some of your peers have assessed your response and you complete the {{ start_tag }}peer assessment{{ end_tag }} step.
+                        You still need to complete the {{ start_tag }}peer assessment{{ end_tag }} step.
                     {% endblocktrans %}
-                {% elif has_self %}
+                {% elif self_incomplete %}
                     {% blocktrans with start_tag='<a data-behavior="ui-scroll" href="#openassessment__self-assessment">'|safe end_tag='</a>'|safe %}
-                        You'll receive your grade after you complete the {{ start_tag }}self assessment{{ end_tag }} step.
+                        You still need to complete the {{ start_tag }}self assessment{{ end_tag }} step.
                     {% endblocktrans %}
                 {% endif %}
             </div>
--- a/openassessment/xblock/test/test_grade.py
+++ b/openassessment/xblock/test/test_grade.py
@@ -322,25 +322,38 @@ class TestGrade(XBlockHandlerTestCase, SubmitAssessmentsMixin):
 
     @scenario('data/grade_incomplete_scenario.xml', user_id='Bunk')
     def test_grade_incomplete_missing_self(self, xblock):
-        # Graded peers, but haven't completed self assessment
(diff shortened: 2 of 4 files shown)


openassessment/templates/openassessmentblock/response/oa_response_submitted.html
```diff
@@ -22,17 +22,18 @@
         <div class="step__message message message--complete">
             <h3 class="message__title">{% trans "Your Response Has Been Submitted" %}</h3>
             <div class="message__content">
-                {% if has_peer and has_self %}
+                {% trans "You'll receive your grade after all steps are complete and your response is fully assessed." %}
```
catong: "You'll" -> " You will"

catong:
@efischer19 Just one small tweak, otherwise :+1:  from me.

efischer19:
Here's an updated screenshot. Note that I've only fixed the green portion here, the other string above it will need to be addressed separately. I also still need to address test changes with regards to this change.

screen shot 2016-01-07 at 2 49 44 pm

catong:
Thanks @efischer19. FYI, I'd like to replace all the contractions such as "You'll" with the full forms.

efischer19:
Feedback addressed, and tests are now passing. This is ready for review now as well, @cahrens and @dianakhuang could you give it a look at some point? I'm still working on https://github.com/edx/edx-ora2/pull/838, if you want to wait for that to be ready and review everything at once that'd work as well.

cahrens:
Would it be straightforward to add unit test coverage of the expected strings (in the different cases)? If so, it seems worth it to me.

When all the PRs are done, please set up a sandbox where I can try them out.

efischer19:
That's a good idea, I'll work on getting those tests added. Sandbox will be created eventually as well.

openassessment/xblock/test/test_grade.py
```diff
@@ -320,28 +320,38 @@ def test_grade_waiting(self, xblock, data):
...
-    def test_grade_incomplete_missing_self(self, xblock):
-        # Graded peers, but haven't completed self assessment
-        self.create_submission_and_assessments(
-            xblock, self.SUBMISSION, [self.PEERS[0]], [PEER_ASSESSMENTS[0]], None
-        )
-        resp = self.request(xblock, 'render_grade', json.dumps(dict()))
-
-        # Verify that we're on the right template
-        self.assertIn(u'not completed', resp.decode('utf-8').lower())
+    def test_grade_incomplete_missing(self, xblock, missing_step):
```
efischer19: This method was already doing all the setup for "incomplete" scenarios, so I elected to just add the new message tests here instead of duplicating all that setup in `test_submission.py`. Any strong objections?

openassessment/xblock/test/test_grade.py
```diff
@@ -320,28 +320,38 @@ def test_grade_waiting(self, xblock, data):
...
+        self_assessment = None
+        peers = []
+        peer_assessments = []
+        if missing_step == 'self':
+            # Graded peers, but haven't completed self assessment
+            peers = [self.PEERS[0]]
+            peer_assessments = [PEER_ASSESSMENTS[0]]
```
dianakhuang: If we have to do specialized logic, I think it would make more sense to make a separate test function for each one with shared set up and/or assertions instead of trying to work around `ddt`.

efischer19: done

dianakhuang:
:+1: from me.

openassessment/xblock/test/test_grade.py
```diff
@@ -322,26 +322,35 @@ def test_grade_waiting(self, xblock, data):
 
     @scenario('data/grade_incomplete_scenario.xml', user_id='Bunk')
     def test_grade_incomplete_missing_self(self, xblock):
-        # Graded peers, but haven't completed self assessment
-        self.create_submission_and_assessments(
-            xblock, self.SUBMISSION, [self.PEERS[0]], [PEER_ASSESSMENTS[0]], None
-        )
-        resp = self.request(xblock, 'render_grade', json.dumps(dict()))
+        resp = self._test_incomplete_helper(xblock, [self.PEERS[0]], None)
+        self.assertNotIn('peer assessment', resp)
```
cahrens: If I'm reading this correctly, it is verifying that the learner is not told to do a peer assessment. However, I don't see anything verifying that they are told to complete their self assessment.

openassessment/xblock/test/test_grade.py
```diff
@@ -322,26 +322,35 @@ def test_grade_waiting(self, xblock, data):
...
 
-        # Verify that we're on the right template
-        self.assertIn(u'not completed', resp.decode('utf-8').lower())
-
-    @scenario('data/grade_incomplete_scenario.xml', user_id='Daniels')
+    @scenario('data/grade_incomplete_scenario.xml', user_id='Bunk')
     def test_grade_incomplete_missing_peer(self, xblock):
-        # Have not yet completed peer assessment
+        resp = self._test_incomplete_helper(xblock, [], SELF_ASSESSMENT)
+        self.assertNotIn('self assessment', resp)
```
cahrens: Same comment-- check "peer assessment" is present.

openassessment/xblock/test/test_grade.py
```diff
@@ -322,26 +322,35 @@ def test_grade_waiting(self, xblock, data):
...
+
+    @scenario('data/grade_incomplete_scenario.xml', user_id='Bunk')
+    def test_grade_incomplete_missing_both(self, xblock):
+        self._test_incomplete_helper(xblock, [], None)
```
cahrens: Check both are present.

cahrens:
:+1: after additional checks added in unit test.

efischer19:
Extra checks added, tests passing locally.

andy-armstrong:
:+1:

PR #840, opened 2016-01-11
Make checkboxes accessible.

cahrens:
TNL-3999

This changes the checkboxes in Studio ORA problem settings to use the native control instead of hiding it and showing font awesome icons. Using the native control allows keyboard accessibility to work (can tab into the control and toggle it).

image

@efischer19 and @andy-armstrong please review


cahrens:
FYI @cptvitamin this will fix the checkboxes in the Studio ORA settings. I've already updated my sandbox with the change for testing purposes.

andy-armstrong:
@talbs Do you remember why ORA didn't use the native controls originally? Are you comfortable with switching over?

cahrens:
@talbs has no idea, and he helped me with the CSS removal.

efischer19:
:+1: from me

andy-armstrong:
FYI here's the PR where Grady introduced this code. I don't see any comments as to why he did it, or review feedback on those lines:

https://github.com/edx/edx-ora2/pull/445

The only thing I could think of is that web fonts will scale up with font size, but that seems like a minor benefit compared to the loss of accessibility.

andy-armstrong:
:+1: given that nobody remembers why we did this. :-)

PR #841, opened 2016-01-11
Move focus within staff panels.

cahrens:
TNL-3932

When one of the staff area buttons is clicked, focus needs to move into the panel, to the first focusable component. This is the close button (which was OK'd by @cptvitamin as the right thing to do).

Sandbox: https://cahrens.sandbox.edx.org

@efischer19 and @andy-armstrong please review


efischer19:
:+1: Looks good to me

andy-armstrong:
:+1: Looks great. As we just discussed, it would be good to decide where the focus should go when you close the panel too.

cahrens:
Excellent point, @andy-armstrong. I have updated the PR (and sandbox) to have focus go back to the "tab" on close. Sorry, but I squashed the commit at the same time.

cahrens:
@efischer19 you should re-review as well.

andy-armstrong:
:+1:

efischer19:
:+1: Tested on sandbox, closing behavior looks good as well.

PR #842, opened 2016-01-11
Editor Assessments Order improvements

efischer19:
## [TNL-4005](https://openedx.atlassian.net/browse/TNL-4005)

Moves `editor_assessments_order` to a property in `studio_mixin`, allowing for validation on every get and set, ensuring invalid orders are never viewed by the user.

### DIFF
--- a/openassessment/xblock/studio_mixin.py
+++ b/openassessment/xblock/studio_mixin.py
@@ -109,7 +109,7 @@ class StudioMixin(object):
 
         submission_start, submission_due = date_ranges[0]
         assessments = self._assessments_editor_context(date_ranges[1:])
-        editor_assessments_order = self._editor_assessments_order_context()
+        self.editor_assessments_order = self._editor_assessments_order_context()
 
         # Every rubric requires one criterion. If there is no criteria
         # configured for the XBlock, return one empty default criterion, with
@@ -139,7 +139,7 @@ class StudioMixin(object):
             'leaderboard_show': self.leaderboard_show,
             'editor_assessments_order': [
                 make_django_template_key(asmnt)
-                for asmnt in editor_assessments_order
+                for asmnt in self.editor_assessments_order
             ],
             'is_released': self.is_released(),
         }
@@ -349,30 +349,38 @@ class StudioMixin(object):
             list of assessment names
 
         """
-        order = copy.deepcopy(self.editor_assessments_order)
-        used_assessments = [asmnt['name'] for asmnt in self.valid_assessments]
-        default_editor_order = copy.deepcopy(DEFAULT_EDITOR_ASSESSMENTS_ORDER)
+        # Start with the default order, to pick up any assessment types that have been added
+        # since the user last saved their ordering.
+        effective_order = copy.deepcopy(DEFAULT_EDITOR_ASSESSMENTS_ORDER)
 
-        # Backwards compatibility:
         # If the problem already contains example-based assessment
-        # then allow the editor to display example-based assessments.
-        if 'example-based-assessment' in used_assessments:
-            default_editor_order.insert(0, 'example-based-assessment')
+        # then allow the editor to display example-based assessments,
+        # which is not included in the default
+        enabled_assessments = [asmnt['name'] for asmnt in self.valid_assessments]
+        if 'example-based-assessment' in enabled_assessments:
+            effective_order.insert(0, 'example-based-assessment')
 
-        # Backwards compatibility:
-        # If the editor assessments order doesn't match the problem order,
-        # fall back to the problem order.
-        # This handles the migration of problems created pre-authoring,
-        # which will have the default editor order.
-        problem_order_indices = [
-            order.index(asmnt_name) for asmnt_name in used_assessments
-            if asmnt_name in order
+        # Account for changes the user has made to the default order
+        user_order = copy.deepcopy(self.editor_assessments_order)
+        effective_order = self._subset_in_relative_order(effective_order, user_order)
+
+        # Account for inconsistencies between the user's order and the problems
+        # that are currently enabled in the problem (These cannot be changed)
+        enabled_ordered_assessments = [
+            assessment for assessment in enabled_assessments if assessment in user_order
         ]
-        if problem_order_indices != sorted(problem_order_indices):
-            unused_assessments = list(set(default_editor_order) - set(used_assessments))
-            return sorted(unused_assessments) + used_assessments
+        effective_order = self._subset_in_relative_order(effective_order, enabled_ordered_assessments)
 
-        # Forwards compatibility:
-        # Include any additional assessments that may have been added since the problem was created.
-        else:
-            return order + list(set(default_editor_order) - set(order))
+        return effective_order
+
+    def _subset_in_relative_order(self, superset, subset):
+        """
+        Returns a copy of superset, with entries that appear in subset being reordered to match
+        their relative ordering in subset.
+        """
+        superset_indices = [superset.index(item) for item in subset]
+        sorted_superset_indices = sorted(superset_indices)
+        if superset_indices != sorted_superset_indices:
+            for i in range(len(sorted_superset_indices)):
+                superset[sorted_superset_indices[i]] = subset[i]
+        return superset
--- a/openassessment/xblock/test/test_studio.py
+++ b/openassessment/xblock/test/test_studio.py
@@ -106,7 +106,8 @@ class StudioViewTest(XBlockHandlerTestCase):
         "example-based-assessment": "oa_ai_assessment_editor",
         "peer-assessment": "oa_peer_assessment_editor",
         "self-assessment": "oa_self_assessment_editor",
-        "student-training": "oa_student_training_editor"
+        "student-training": "oa_student_training_editor",
+        "staff-assessment": "oa_staff_assessment_editor",
     }
 
     @scenario('data/basic_scenario.xml')
@@ -246,22 +247,17 @@ class StudioViewTest(XBlockHandlerTestCase):
 
     @scenario('data/self_then_peer.xml')
     def test_render_editor_assessment_order(self, xblock):
-        # Initially, the editor assessment order should be the default
-        # (because we haven't set it yet by saving in Studio)
-        # However, the assessment order IS set when we import the problem from XML,
-        # and it differs from the default order (self->peer instead of peer->self)
         # Expect that the editor uses the order defined by the problem.
         self._assert_rendered_editor_order(xblock, [
-            'example-based-assessment',
             'student-training',
             'self-assessment',
             'peer-assessment',
+            'staff-assessment'
         ])
 
         # Change the order (simulates what would happen when the author saves).
         xblock.editor_assessments_order = [
             'student-training',
-            'example-based-assessment',
             'peer-assessment',
             'self-assessment',
         ]
@@ -273,9 +269,9 @@ class StudioViewTest(XBlockHandlerTestCase):
         # Expect that the rendered view reflects the new order
         self._assert_rendered_editor_order(xblock, [
             'student-training',
-            'example-based-assessment',
             'peer-assessment',
             'self-assessment',
+            'staff-assessment',
         ])
 
     def _assert_rendered_editor_order(self, xblock, expected_assessment_order):
@@ -304,6 +300,7 @@ class StudioViewTest(XBlockHandlerTestCase):
         actual_assessment_order = [
             index_dict['name']
             for index_dict in sorted(assessment_indices, key=lambda d: d['index'])
+            if index_dict['index'] > 0
         ]
         self.assertEqual(actual_assessment_order, expected_assessment_order)
 


efischer19:
Actually, nevermind about the review just yet, I'll get a unit test up here too.

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,70 @@ class StudioMixin(object):
...
+            order_to_validate.append(u'staff-assessment')
+            logger.info('Backwards compatibility: editor_assessments_order now contains staff-assessment')
+
+        # Backwards compatibility:
+        # If the problem already contains example-based assessment
+        # then allow the editor to display example-based assessments.
+        enabled_assessments = [asmnt['name'] for asmnt in self.valid_assessments]
+        if u'example-based-assessment' in enabled_assessments and u'example-based-assessment' not in order_to_validate:
+            order_to_validate.insert(0, u'example-based-assessment')
+
```
efischer19: add logging here

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,70 @@ class StudioMixin(object):
...
+        # Forwards compatibility:
+        # Include any additional assessments that may have been added since the problem was created.
+        new_assessments = list(set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) - set(order_to_validate))
+        if len(new_assessments) > 0:
+            order_to_validate = new_assessments + order_to_validate
+
```
efischer19: logging

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,70 @@ class StudioMixin(object):
...
+            asmnt_name for asmnt_name in enabled_assessments
+            if asmnt_name in order_to_validate
+        ]
+        enabled_assessment_indices = [
+            order_to_validate.index(asmnt_name) for asmnt_name in enabled_ordered_assessments
+        ]
+        if enabled_assessment_indices != sorted(enabled_assessment_indices):
+            for enabled_index, new_index in enumerate(enabled_assessment_indices):
+                order_to_validate[new_index] = enabled_ordered_assessments[enabled_index]
+
```
efischer19: log that mess

openassessment/xblock/studio_mixin.py
```diff
@@ -81,14 +138,7 @@ def studio_view(self, context=None):
...
         """
-        Update the XBlock's XML.
-
-        Args:
-            data (dict): Data from the request; should have a value for the key 'xml'
-                containing the XML for this XBlock.
-
-        Keyword Arguments:
-            suffix (str): Not used
+        Gets the editor context, ready for rendering.
```
efischer19: This docstring was wildly out-of-date, should now more accurately reflect what's happening here

openassessment/xblock/studio_mixin.py
```diff
@@ -172,17 +222,10 @@ def update_editor_context(self, data, suffix=''):
...
-            if set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) == (
-                # Check the given set, minus example-based, plus staff
-                given_without_example_based | {'staff-assessment'}
-            ):
-                data['editor_assessments_order'].append('staff-assessment')
-                logger.info('Backwards compatibility: editor_assessments_order now contains staff-assessment')
-            else:
-                logger.exception('editor_assessments_order does not contain all expected assessment types')
-                return {'success': False, 'msg': self._('Error updating XBlock configuration')}
+            logger.exception('editor_assessments_order does not contain all expected assessment types')
```
efischer19: Since `staff-assessment` is now added at validation, I've elected to remove the ability to skip it here. Is this an acceptable decision? It means that `update_editor_context` must always be called with an `editor_assessments_order` that includes `staff-assessment`.

openassessment/xblock/test/data/update_xblock.json
```diff
@@ -56,59 +56,6 @@
...
-                "due": "4014-03-10T00:00"
-            },
-            {
-                "name": "self-assessment",
-                "start": null,
-                "due": null
-            }
-        ],
-        "editor_assessments_order": ["student-training", "peer-assessment", "self-assessment"]
-    },
```
efischer19: See note above. No longer is it possible to update the xblock to an order that does not include `staff`. My reasoning is that staff will always be present, so let's force the user to acknowledge that they know that and have included it in their order.

openassessment/xblock/test/test_studio.py
```diff
@@ -154,6 +155,25 @@ def test_update_editor_context_saves_assessment_order(self, xblock):
         self.assertEqual(xblock.editor_assessments_order, data['editor_assessments_order'])
 
     @scenario('data/basic_scenario.xml')
+    def test_editor_context_adds_staff_in_order(self, xblock):
+        # Mock loading an old problem that has saved an editor_assessments_order without staff-assessment
+        # Note self, then peer - this deliberately does *not* match the order given in basic_scenario
+        xblock.editor_assessments_order = ['student-training', 'self-assessment', 'peer-assessment']
```
efischer19: This is the test that fails on current `ora-staff-grading` code (checkout this branch, then `HEAD~1` to verify). It simulates loading an old xblock from xml with an `editor_assessments_order` that does not match the assessments as they are defined on the problem.

openassessment/xblock/test/test_studio.py
```diff
@@ -154,6 +155,25 @@ def test_update_editor_context_saves_assessment_order(self, xblock):
...
+
+        # Update the editor context
+        xblock.runtime.modulestore = MagicMock()
+        response_dict = xblock.editor_context()
```
efischer19: Note that the bug was *not* present if going through `update_editor_context`, only `editor_context`. This was the impetus behind moving all the validation logic to a property as close to the actual value as possible.

openassessment/xblock/test/test_studio.py
```diff
@@ -254,8 +274,8 @@ def test_render_editor_assessment_order(self, xblock):
         self._assert_rendered_editor_order(xblock, [
             'example-based-assessment',
             'student-training',
-            'self-assessment',
             'peer-assessment',
+            'self-assessment',
```
efischer19: New behavior - will force the `editor_assessments_order` to match the assessments defined in the problem. In this case, `data/basic_scenario.xml` defines peer before self.

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,69 @@ class StudioMixin(object):
...
+            logger.info('Backwards compatibility: editor_assessments_order now contains example-based-assessment')
+
+        # Forwards compatibility:
+        # Include any additional assessments that may have been added since the problem was created.
+        new_assessments = list(set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) - set(order_to_validate))
+        if len(new_assessments) > 0:
+            order_to_validate = new_assessments + order_to_validate
+            logger.info('Forwards compatibility: editor_assessments_order now includes {}'.format(new_assessments)
+
+        # Consistency
```
efischer19: This block contains the new logic. Previously, we had been calculating all of the unused assessment types, then just slapping them onto the front of the returned list without any regard for order. Now, I just change the ordering of the enabled assessments to match the order in which they appear in the problem, while leaving the unused assessments untouched in `order_to_validate`.

efischer19:
@cahrens @andy-armstrong This is ready for review again, let me know if I can help explain anything.

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,69 @@ class StudioMixin(object):
...
+    _editor_assessments_order = List(
         default=DEFAULT_EDITOR_ASSESSMENTS_ORDER,
         scope=Scope.content,
         help="The order to display assessments in the editor."
     )
+    _editor_assessments_order_validated = False
+
+    def _get_editor_assessments_order(self):
+        if not self._editor_assessments_order_validated:
+            self._set_editor_assessments_order(self._editor_assessments_order)
```
cahrens: Just call validate every time, it's easier to understand.

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,69 @@ class StudioMixin(object):
...
+        # then allow the editor to display example-based assessments.
+        enabled_assessments = [asmnt['name'] for asmnt in self.valid_assessments]
+        if u'example-based-assessment' in enabled_assessments and u'example-based-assessment' not in order_to_validate:
+            order_to_validate.insert(0, u'example-based-assessment')
+            logger.info('Backwards compatibility: editor_assessments_order now contains example-based-assessment')
+
+        # Forwards compatibility:
+        # Include any additional assessments that may have been added since the problem was created.
+        new_assessments = list(set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) - set(order_to_validate))
+        if len(new_assessments) > 0:
```
cahrens: Move this to the top-- it should be done before example-based.

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,69 @@ class StudioMixin(object):
...
+            order_to_validate = new_assessments + order_to_validate
+            logger.info('Forwards compatibility: editor_assessments_order now includes {}'.format(new_assessments))
+
+        # Consistency
+        # If the editor assessments order doesn't match the problem order,
```
cahrens: Won't this prevent you from being able to change the order of the steps in Studio?

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,69 @@ class StudioMixin(object):
...
+            order_to_validate.index(asmnt_name) for asmnt_name in enabled_ordered_assessments
+        ]
+        if enabled_assessment_indices != sorted(enabled_assessment_indices):
+            for enabled_index, new_index in enumerate(enabled_assessment_indices):
+                order_to_validate[new_index] = enabled_ordered_assessments[enabled_index]
+            logger.info("Consistency: editor_assessments_order updated to reflect order of enabled assessments")
+
+        # We're ready to set the new value - let's run one last check to ensure validity
+        if not is_valid_assessment_sequence(order_to_validate):
+            logger.info('Invalid editor_assessments_order! Could not validate the list {}'.format(order_to_validate))
```
cahrens: This is just logging and then returning order_to_validate.

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,93 @@ class StudioMixin(object):
...
-    # that the user specified.
-    editor_assessments_order = List(
+    # that the user specified. We use a property in order to add validation
+    # logic, and update old problems with new validation requirements on load
+    _editor_assessments_order = List(
         default=DEFAULT_EDITOR_ASSESSMENTS_ORDER,
         scope=Scope.content,
         help="The order to display assessments in the editor."
     )
+    _editor_assessments_order_validated = False
```
efischer19: unused variable

openassessment/xblock/studio_mixin.py
```diff
@@ -45,12 +45,93 @@ class StudioMixin(object):
...
+        """
+        # Forwards compatibility:
+        # Include any additional assessments that may have been added since the problem was created.
+        new_assessments = list(set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) - set(order_to_validate))
+        if len(new_assessments) > 0:
+            if fix:
+                order_to_validate = new_assessments + order_to_validate
+                logger.info('Forwards compatibility: editor_assessments_order now includes {}'.format(new_assessments))
+            else:
+                logger.info('Invalid editor_assessments_order! Returning old value, no changes made.')
```
efischer19: Should I make these log values more descriptive?

openassessment/xblock/studio_mixin.py
```diff
@@ -45,13 +45,103 @@ class StudioMixin(object):
...
+            sorted_enabled_indices = sorted(enabled_assessment_indices)
+            if enabled_assessment_indices != sorted_enabled_indices:
+                # But if they don't match, we have:
+                    # The list of assessments whose order we care about (enabled_ordered_assessments)
+                    # The indices of those assessments in order_to_validate (enabled_assessment_indices)
+                    # The *desired* indices of those assessments in order_to_validate (sorted_enabled_indices)
+                # So we can loop over the data we have and "fix" order_to_validate to match the problem order
+                for i in range(len(sorted_enabled_indices)):
+                    order_to_validate[sorted_enabled_indices[i]] = enabled_ordered_assessments[i]
+                logger.info("Consistency: editor_assessments_order updated to reflect order of enabled assessments")
```
efischer19: Here's some logged data showing an example of this fix occurring, if it helps:
```
Before fixing anything:
order_to_validate: ['student-training', 'peer-assessment', 'self-assessment', 'staff-assessment']
enabled_ordered_assessments: [u'self-assessment', u'peer-assessment']
enabled_assessment_indices: [2, 1]
sorted_enabled_indices: [1, 2]

Fix occurs:
Consistency: editor_assessments_order updated to reflect order of enabled assessments
order_to_validate: ['student-training', u'self-assessment', u'peer-assessment', 'staff-assessment']
return from get: ['student-training', u'self-assessment', u'peer-assessment', 'staff-assessment']
```

openassessment/xblock/studio_mixin.py
efischer19: Needs to be after staff, or else it will prepend staff as part of its logic.

openassessment/xblock/studio_mixin.py
```diff
@@ -169,20 +252,10 @@ def update_editor_context(self, data, suffix=''):
...
-            # If that is the only problem with this data, just add it manually and continue.
-            if set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) == (
-                # Check the given set, minus example-based, plus staff
-                given_without_example_based | {'staff-assessment'}
-            ):
-                data['editor_assessments_order'].append('staff-assessment')
-                logger.info('Backwards compatibility: editor_assessments_order now contains staff-assessment')
-            else:
-                logger.exception('editor_assessments_order does not contain all expected assessment types')
-                return {'success': False, 'msg': self._('Error updating XBlock configuration')}
```
efischer19: This block is now part of the validation logic, see `# Include any additional assessments that may have been added since the problem was created.` above

openassessment/xblock/test/data/invalid_update_xblock.json
```diff
@@ -1403,7 +1403,7 @@
                 "due": null
             }
         ],
-        "editor_assessments_order": ["student-training", "peer-assessment", "self-assessment", "example-based-assessment", "staff-assessment"],
+        "editor_assessments_order": ["example-based-assessment", "student-training", "peer-assessment", "self-assessment", "staff-assessment"],
```
efischer19: The changes to validation have made "Verify that the new `editor_assessments_order` is valid" be checked before the non-unique option names this test is supposed to be checking, so I updated the order to a valid one.

openassessment/xblock/test/test_studio.py
```diff
@@ -106,7 +107,8 @@ class StudioViewTest(XBlockHandlerTestCase):
         "example-based-assessment": "oa_ai_assessment_editor",
         "peer-assessment": "oa_peer_assessment_editor",
         "self-assessment": "oa_self_assessment_editor",
-        "student-training": "oa_student_training_editor"
+        "student-training": "oa_student_training_editor",
+        "staff-assessment": "oa_staff_assessment_editor",
```
efischer19: Turns out, adding staff to this line and `test_render_editor_assessment_order` below will cause a failure before this fix due to the bug this ticket was filed against. I've removed the test I initially added since it is now superfluous.

openassessment/xblock/test/test_studio.py
```diff
@@ -256,14 +256,16 @@ def test_render_editor_assessment_order(self, xblock):
...
             'self-assessment',
             'peer-assessment',
+            'staff-assessment',
         ])
 
         # Change the order (simulates what would happen when the author saves).
         xblock.editor_assessments_order = [
-            'student-training',
             'example-based-assessment',
+            'student-training',
```
efischer19: This was previously an invalid order to be setting, but with example-based not enabled it didn't matter. Now the validation checks are stricter, so I moved it around to make the list valid.

efischer19:
@andy-armstrong @cahrens This is ready to be reviewed

openassessment/xblock/studio_mixin.py
```diff
@@ -45,13 +45,103 @@ class StudioMixin(object):
...
+        Returns:
+            list of string, a valid value for self._editor_assessments_order
+        """
+        # Backwards compatibility - may be loading a module without staff-assessment
+        if u'staff-assessment' not in order_to_validate:
+            if fix:
+                order_to_validate.append(u'staff-assessment')
+                logger.info('Backwards compatibility: editor_assessments_order now contains staff-assessment')
+            else:
+                logger.info('Invalid editor_assessments_order, staff-assessment not present. Returning old value, no changes made.')
```
cahrens: This should be warning or error level instead of info.

openassessment/xblock/studio_mixin.py
```diff
@@ -45,13 +45,103 @@ class StudioMixin(object):
...
+
+        # Forwards compatibility:
+        # Include any additional assessments that may have been added since the problem was created.
+        new_assessments = list(set(DEFAULT_EDITOR_ASSESSMENTS_ORDER) - set(order_to_validate))
+        if len(new_assessments) > 0:
+            if fix:
+                order_to_validate = new_assessments + order_to_validate
+                logger.info('Forwards compatibility: editor_assessments_order now includes {}'.format(new_assessments))
+            else:
+                logger.info('Invalid editor_assessments_order, not all expected types present. Returning old value, no changes made.')
```
cahrens: Same issue here.

openassessment/xblock/studio_mixin.py
```diff
@@ -349,30 +349,38 @@ def _editor_assessments_order_context(self):
...
+        return effective_order
+
+    def _subset_in_relative_order(self, superset, subset):
+        """
+        Returns a copy of superset, with entries that appear in subset being reordered to match
+        their relative ordering in subset.
+        """
+        superset_indices = [superset.index(item) for item in subset]
+        sorted_superset_indices = sorted(superset_indices)
+        if superset_indices != sorted_superset_indices:
```
cahrens: Nit: I find it confusing that you use len(sorted_superset_indices), since the code in the loop is using superset_indices. I know they have the same length, but I would prefer len(superset_indices).

efischer19: I actually got this wrong the other way - it should be `sorted_superset_indices` in the loop as well.

openassessment/xblock/test/test_studio.py
```diff
@@ -304,6 +300,7 @@ def _assert_rendered_editor_order(self, xblock, expected_assessment_order):
         actual_assessment_order = [
             index_dict['name']
             for index_dict in sorted(assessment_indices, key=lambda d: d['index'])
+            if index_dict['index'] > 0
```
efischer19: Turns out, `find()` returns -1 if a value is not found, which makes that value always appear at the front of `actual_assessment_order` even if it is not present at all.

openassessment/xblock/test/test_studio.py
```diff
@@ -246,22 +247,17 @@ def test_check_released(self, xblock):
 
     @scenario('data/self_then_peer.xml')
     def test_render_editor_assessment_order(self, xblock):
-        # Initially, the editor assessment order should be the default
-        # (because we haven't set it yet by saving in Studio)
-        # However, the assessment order IS set when we import the problem from XML,
-        # and it differs from the default order (self->peer instead of peer->self)
         # Expect that the editor uses the order defined by the problem.
         self._assert_rendered_editor_order(xblock, [
-            'example-based-assessment',
```
efischer19: See below for fix - `example-based-assessment` should have never been appearing here, since it is not defined in the problem.

openassessment/xblock/test/test_studio.py
```diff
@@ -246,22 +247,17 @@ def test_check_released(self, xblock):
...
             'student-training',
             'self-assessment',
             'peer-assessment',
+            'staff-assessment'
         ])
 
         # Change the order (simulates what would happen when the author saves).
         xblock.editor_assessments_order = [
             'student-training',
-            'example-based-assessment',
```
efischer19: Since `example-based` does not exist for this problem, I've removed it from this test case.

efischer19:
https://efischer19.sandbox.edx.org/ is updated as of https://github.com/edx/edx-ora2/commit/c0af3fc87832f46ac62e694645b5f3ac8e748824. I've manually verified the bug for TNL-4005 to be fixed, and run both acceptance and accessibility tests (passed with flying colors).

openassessment/xblock/studio_mixin.py
```diff
@@ -349,30 +349,38 @@ def _editor_assessments_order_context(self):
...
+            effective_order.insert(0, 'example-based-assessment')
+
+        # Account for changes the user has made to the default order
+        user_order = copy.deepcopy(self.editor_assessments_order)
+        effective_order = self._subset_in_relative_order(effective_order, user_order)
+
+        # Account for inconsistencies between the user's order and the problems
+        # that are currently enabled in the problem (These cannot be changed)
+        enabled_ordered_assessments = [
+            assessment for assessment in enabled_assessments if assessment in enabled_assessments
```
efischer19: This should be `assessment for assessment in enabled_assessments if assessment in user_order`. In all our tests, `enabled_assessments` is a subset of `user_order`, so this logic still worked, but for the wrong reasons.

andy-armstrong:
:+1: I really like where this landed.

openassessment/xblock/test/test_studio.py
cahrens: Ah, got it now!

openassessment/xblock/studio_mixin.py
```diff
@@ -349,30 +349,38 @@ def _editor_assessments_order_context(self):
...
-            order.index(asmnt_name) for asmnt_name in used_assessments
-            if asmnt_name in order
+        # then allow the editor to display example-based assessments,
+        # which is not included in the default
+        enabled_assessments = [asmnt['name'] for asmnt in self.valid_assessments]
+        if 'example-based-assessment' in enabled_assessments:
+            effective_order.insert(0, 'example-based-assessment')
+
+        # Account for changes the user has made to the default order
+        user_order = copy.deepcopy(self.editor_assessments_order)
```
cahrens: I don't think the copy is necessary-- I don't see that self.editor_assessments_order is changed. Perhaps it is safer though in case the helper method changed user_order.

cahrens:
:+1: Note that you can put staff someplace besides last place and save the change if the staff step is not enabled. I verified that this is consistent with existing behavior for training (training can be saved after the first position, as long as it is not enabled). If staff is enabled, it must be last.

PR #846, opened 2016-01-13
Changing wording of "waiting for peers" message.

cahrens:
TNL-4019

@andy-armstrong and @efischer19 please review

This is how it looks now:

image


andy-armstrong:
As we discussed, I would like to also change the message at the top of the problem that is similarly confusing.

cahrens:
@andy-armstrong I updated that second message as well.

FYI, @catong

efischer19:
:+1: Looks good to me, I'd wait for @catong to give final approval on the exact wording but this is certainly better.

catong:
Much better! :+1:

andy-armstrong:
:+1: looks great!

PR #847, opened 2016-01-14
New i18n strings for release

efischer19:
FYI @cahrens @andy-armstrong 

Looks like a few late breaking string updates were not included on `ora-staff-grading`, so I'm making sure these are as up-to-date as possible for the new ORA release.


efischer19:
Green build, only translation updates, merging.

PR #852, opened 2016-01-20
Add feedback text to self assessments

efischer19:
## [TNL-3882](https://openedx.atlassian.net/browse/TNL-3882)

Also includes fix to default text, and removal of a seemingly unused,
hardcoded 'estimated_time' context variable.

### Sandbox
- [x] https://efischer19.sandbox.edx.org/

### Testing
- [x] manually tested to verify that previously un-used template variable is now filled in
- [x] old a11y test now able to pass successfully

### Reviewers
- [x] Code review: @cahrens 
- [ ] UI strings/error msgs review: @catong 

### Post-review
- [x] Squash commits


efischer19:
@catong I've implemented the changes you suggested on the TNL ticket

screen shot 2016-01-20 at 3 27 40 pm

efischer19:
Wait, hang on @catong, the text should be updated, but my screenshot clearly shows it isn't... I'll investigate.

openassessment/xblock/peer_assessment_mixin.py
```diff
@@ -165,7 +165,6 @@ def peer_path_and_context(self, continue_grading):
 
         context_dict = {
             "rubric_criteria": self.rubric_criteria_with_labels,
-            "estimated_time": "20 minutes",  # TODO: Need to configure this.
```
efischer19: This was never being used, so by the Boy Scout rule, I'm removing it.

Confirmation that it doesn't appear in any templates anywhere (run after making these changes):
```
ericfischer@Eric-Fischers-MacBook-Pro:~/devstack/src/edx-ora2$ grep -rnw . -e "estimated_time"
grep: ./edx-ora2/local/bin: No such file or directory
grep: ./edx-ora2/local/include: No such file or directory
grep: ./edx-ora2/local/lib: No such file or directory

ericfischer@Eric-Fischers-MacBook-Pro:~/devstack/src/edx-ora2$ 
```

openassessment/xblock/defaults.py
```diff
@@ -63,13 +63,13 @@
 # The rubric's feedback prompt is a set of instructions letting the student
 # know they can provide additional free form feedback in their assessment.
 DEFAULT_RUBRIC_FEEDBACK_PROMPT = """
-(Optional) What aspects of this response stood out to you? What did it do well? How could it improve?
+(Optional) What aspects of this response stood out to you? What did it do well? How could it be improved?
```
catong: @efischer19 Thanks! The screenshot showing the old text also shows an odd indent for this line. Does that issue still exist and if yes does this PR fix it?

efischer19: I'm looking into the indentation now, that struck me as odd also.

As far as the old/new text, changing the defaults will only affect *new* problems, as old ones have already been saved with the previous default value. Is this a huge problem?

screen shot 2016-01-20 at 3 43 29 pm

catong: @efischer19 I think not updating existing problems is fine.

openassessment/xblock/static/sass/oa/views/_oa-base.scss
```diff
@@ -719,6 +719,15 @@
       // genereal feedback question
       .assessment__rubric__question--feedback {
 
+        .wrapper--input {
+          margin-top: $baseline-v;
+        }
+
+        .question__title__copy {
+          @include margin-left(0);
```
efischer19: This is what was causing the odd indentation for self, it was already present for peer: https://github.com/edx/edx-ora2/blob/eff9961015d1164e2b119c7eb9a91e6580740c8d/openassessment/xblock/static/sass/oa/views/_oa-base.scss#L658

efischer19:
I can confirm that the above sass change fixes the indentation issue (on devstack), but I can't get it to update the effective style sheets on my sandbox for some reason.

screen shot 2016-01-20 at 4 48 25 pm

efischer19:
@dianakhuang @cahrens @catong I believe this PR is ready for final review.

The odd indentation issue Carol noted is fixed, as seen in my latest screenshot (using devstack), but I can't get the styles to update on the sandbox and I'm not sure what voodoo I need to utter to make it so. Everything else (default text change for new problems, feedback prompt/example for self assessment feedback) is present, so check that out if you'd like.

I've also run the a11y tests locally and confirmed they are passing.

efischer19:
AHA! I was forgetting to git-submit the updated css files after `make sass`, the sandbox should be good-to-go now.

openassessment/xblock/static/sass/oa/views/_oa-base.scss
cahrens: Instead of adding the override, is it possible to remove the thing that caused the indent? Right now I see various places where question__title__copy is associated with the left margin, and then overwritten with a margin of 0.

When is the left margin desirable? Are all the places where question__title__copy appears in SASS actually necessary?

efischer19: Looks like it is used, in some places. I tracked down this instance (https://github.com/edx/edx-ora2/blob/master/openassessment/templates/openassessmentblock/grade/oa_grade_complete.html#L47) on my sandbox, and removing the `margin-left: 10px;` is a bad idea there.

screen shot 2016-01-21 at 9 54 57 am

Whether we should be using the same class for that text and the `default_feedback_prompt` is probably a discussion we should have at some point, but for now I'm okay with copying the approach used by peer and staff assessments.
https://github.com/edx/edx-ora2/blob/efischer/add_feedback_to_self/openassessment/xblock/static/sass/oa/views/_oa-base.scss#L658
https://github.com/edx/edx-ora2/blob/eff9961015d1164e2b119c7eb9a91e6580740c8d/openassessment/xblock/static/sass/oa/elements/_staff.scss#L230

cahrens:
:+1:

dianakhuang:
:+1:

catong:
Me too! :+1:

PR #853, opened 2016-01-20
Change how staff tool tabs work.

cahrens:
## [TNL-4002](https://openedx.atlassian.net/browse/TNL-4002)

Fixes keyboard/voice over accessibility issues in the staff tools sections.

Also contains fixes for TNL-4023 and TNL-4021.

### Sandbox
- [x] https://finalgrade.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb/

### Testing
- [x] Unit, integration, acceptance tests as appropriate. Bok choy acceptance and a11y tests are passing.

### Reviewers
- [ ] Accessibility review: @clrux / @cptvitamin 

FYI: @robrap, @andy-armstrong, @scottrish 

### Post-review
- [ ] Squash commits

(original comments for Mark)
@cptvitamin Can you do an accessibility review of this? It fixes all the known issues with keyboard navigation (and hopefully screenreaders) in the ORA staff tools. Specifically:

1) Keyboard focus moves to the "Close button" when a tab is selected. If the tab is closed, focus moves back to the button (this change was already checked in, but I don't think you have re-reviewed yet).

2) Collapsible/expandable sections are using JQuery slideUp/slideDown so that the hidden elements are not keyboard accessible anymore (and hopefully the screenreader doesn't pick them up). This includes the tabs themselves, the various collapsible sections under "Manage Individual Learners", and the staff grading tool ("Grade Available Responses").

Sandbox: https://finalgrade.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb/


openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -13,12 +13,12 @@ <h2 class="title">
     </h2>
 
     <div class="staff-info__status ui-staff__content__section wrapper--ui--collapse staff-info__student__response">
-        <div class="ui-staff ui-toggle-visibility is--collapsed">
-            <h2 class="staff-info__title ui-staff__subtitle ui-toggle-visibility__control">
-                <i class="icon fa fa-caret-right" aria-hidden="true"></i>
-                <span>{% trans "Learner's Response" %}</span>
+        <div class="ui-staff ui-slidable" role="button" aria-pressed="false" tabindex="0">
```
cahrens: @cptvitamin for these collapsible sections, I left them as divs because changing to a button caused too many styling issues. However, I made them focusable, added the role of button, and am toggling "aria-pressed".

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -125,27 +163,28 @@
...
-                // Collapse the editor and update the counts.
-                $staffGradeTab.find('.staff__grade__control').toggleClass('is--collapsed', true);
-                $staffGradeTab.find('.staff__grade__form').replaceWith('<div class="staff__grade__form"></div>');
-                view.updateStaffGradeCounts();
+            $staffGradeControl.toggleClass('is--showing', true);
+            if (this.staffGradeFormLoaded) {
+                $staffGradeControl.find('.ui-slidable__content').slideDown(
+                    0,  // If transition is visible, the shrinking/growing of the form is seen.
+                    function() {  // This is to work around a JQuery bug (fixed in recent versions of JQuery).
+                        $('.ui-slidable__content').css('overflow', 'visible');
```
cahrens: @talbs There is a bug in "older versions of JQuery" that overflow:hidden is added on slideUp, and not removed on slideDown. I added this workaround because I was seeing the panel appearing at about half-width after the first slide-up.

However, I am not experiencing sizing issues in the other places where I am using slideUp/slideDown (despite the "bad" overflow value existing there), which makes me think there is probably something I could do to a parent container that would cause overflow:hidden to not be an issue. Do you have an idea of what I could do?

This UI is the "Grade Available Responses" staff tools tab at the end of the ORA problem on https://finalgrade.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb/
The slideUp/slideDown calls happen when you press the "Staff Assessment" button.

openassessment/templates/openassessmentblock/staff_area/oa_staff_grade_learners.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n %}
 
-<div class="staff__grade__control ui-toggle-visibility is--collapsed">
-    <header class="staff__grade__header ui-toggle-visibility__control">
+<div class="staff__grade__control ui-slidable">
+    <header class="staff__grade__header ui-slidable__control">
         <h3 class="staff__grade__title">
             <span class="wrapper--copy">
                 <button class="staff__grade__show-form">{% trans "Staff Assessment" %}</button>
```
cahrens: @cptvitamin note that I chose (in a previous PR) to make this control a button so that keyboard focus is handled correctly.

cahrens:
@clrux This is the PR I was talking about-- thanks for taking a look! This is specifically around the "staff tools" that appear at the end of the ORA problem here:

https://finalgrade.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb/

They look like this (log in as staff):

image

image

Note that I haven't done anything with the "View Assignment Statistics" tab. I'm not sure how a screenreader accesses the content there (which has no actionable, and thus focusable, components). My work has really been on the "Manage Individual Learners" and "Grade Available Responses" tabs.

clrux:
@cahrens some findings on my end:

* After entering the username/email and submitting the form, let's send focus to the new content that appears. At present there's no audible indication that new content has appeared or that the form submitted. You could send focus to "Viewing learner: staff".
* The collapsible options are `div role="button"` but I'm wondering if you could just use `button` for this? Buttons are natively focusable and would fall naturally within the tab order.
* Furthermore, VoiceOver controls to press a button (Ctrl + Option + Shift + Spacebar) should be the same as pressing Enter/Return or clicking. It toggles the state of the button but it doesn't toggle the collapsible panel.
* On a PC, even pressing Enter when focusing on a collapsible toggle doesn't open the container.
* All of the content contained in each collapsible container - because the container has the role of a toggle button - is announced as a toggle button. This includes the tabular data and any other content that is _not_ a toggle button. For example, the data in Peer Assessments table isn't navigable. I think converting the collapsible toggles to `buttons` as in bullet two above would help with this. We've recently made changes to the accordion navigation in Courseware so maybe that could be a good example to follow?

In short, the two showstopper issues that enclose the others are the collapsible containers aren't wholly functional based on their markup (`div` that acts as a `button` yet contains all the content, versus a `button` that toggles a sibling content `div`), and keyboard listeners to interact with the toggles.

Happy to preview this with you on both Mac and PC if you need visual help.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
clrux: Can you use CSS/Sass to handle presentation rather than jQuery? For example, when expanding and collapsing, let's add classes (i.e., `is-expanded`) which would then contain styles for opened containers. You could have in this class `overflow: visible;` for example, among others. When collapse, the class is removed.

cahrens: @clrux We could, but I thought you recommended slideUp and slideDown as the best way to do this. There are advantages to slideUp and slideDown in that you can get the animation effect (though I'm not getting it in this particular case due to the bug). You can see the animation when selecting the tabs "Manage Individual Learners" and "Grade Available Responses".

Honestly, I don't know what was causing the accessibility issues with the CSS that was there before. Perhaps an easier fix would be to make that CSS correct, but I was under the impression that slideUp and slideDown was the better way to go.

cahrens: This is the css that was being used for is--hidden. Why was this not sufficient for screen readers/keyboard navigation?

````
// --------------------
// UI: element rendering/visibility
// --------------------
%state-hidden {
  @extend %state-disabled;
  display: none;
  visibility: hidden;
}

%state-disabled {
  pointer-events: none;
  outline: none;
}
````

Comment (clrux) [inline openassessment/xblock/static/js/src/lms/oa_staff_area.js]: My understanding was that the animation of sliding up/down needed a fixed opened height in order to animate. Since the height was variable I suggested using `$.slideUp()` and `$.slideDown()` which does basically the animation but doesn't require the fixed heights.

Not sure how important the animations are, but the above mixins would hide content visually and from screen readers, but would not have the sliding animation.

If the animations are important then I'd do something like the following:
```
$(el).slideUp().removeClass('is-expanded');
...
$(el).slideDown().addClass('is-expanded');
```
And then have all other styles - unrelated to height and display, which are handled by jQuery - part of the `is-expanded` class. This may be padding, margins, font styling, etc.

Comment (cahrens) [inline openassessment/xblock/static/js/src/lms/oa_staff_area.js]: @clrux Sorry, I provided the CSS for the wrong class (that was for "is--hidden" instead of "is--collapsed"). This is the CSS that was being used in the "collapsed" state:

```
  // STATE: is collapsed
  &.is--collapsed {

    >.ui-toggle-visibility__control > .icon {
      @include transform(rotate(bidi-rotate-angle(0deg)));
      @include transform-origin(50% 50%);
    }

    // item being toggled
    .ui-toggle-visibility__content {
      @include transition(opacity $tmg-f2 ease-in 0, max-height $tmg-f1 ease-in 0);
      @extend %trans-opacity;
      max-height: 0;
      overflow: hidden;
      opacity: 0.0;
    }
```

So display was not being set to none, and opacity was being used, which I guess is the accessibility issue?

Comment (clrux) [inline openassessment/xblock/static/js/src/lms/oa_staff_area.js]: Yes, that would mean that the screenreader would read everything, even "hidden" stuff, making the content very lengthy. Using jQuery to `slideUp()` and `slideDown()` will animate the container open and closed and set it's `display` to `none` when closed.

**Comment by cahrens** [comment]
@clrux Can you take another pass at the sandbox for this PR? I have converted to buttons, moved the containing HTML to peer level, moved focus as you suggested, etc. I also worked through my overflow hidden issues, so now animation is in full-force...

Thanks!

**Comment by clrux** [comment]
@cahrens Working much better! A few additional things:

* try sending focus to the expanded container once it opens so the flow is continuous
* opening the section by pressing the button works, but closing it (pressing the button again) submits the form above
* when the sections are expanded, add `aria-expanded="true"` to the button. Toggle that state to `false` when the section is closed. This will allow screenreaders who are navigating around to know which section is open and which are closed. Example: `button aria-expanded="true|false"`
* for even more robustness, give the buttons `id`'s and add `aria-labelledby="[id]"` on its respective section. This would be especially helpful for the tables, which don't have (but need) `captions`. You may add `class="sr"` to the captions. Example: `button id="button_01"` ... `<div class="ui-slidable__content" aria-labelledby="button_01"> ...`

```
<table>
    <caption class="sr">Briefly describe the data of the table</caption>
    ...
</table>
```

**Comment by cahrens** [comment]
@clrux for this one-- "try sending focus to the expanded container once it opens so the flow is continuous", the non-editable (data only) sections have nothing focusable. Do you suggest making the container of the data focusable?

Which button are you referring to in this comment? "opening the section by pressing the button works, but closing it (pressing the button again) submits the form above"
If it is the submit button shown below, that is not intended to close the section below-- it should always submit the form.

image

**Comment by cahrens** [comment]
@clrux My sandbox is updated with the latest changes:

* Added aria-expanded on the top-level buttons that control which tab the staff tools display, the collapsible/expandable buttons on the first panel, and the "Staff Assessment" button on the third tab ("Grade Available Responses").

* Added aria-labelled by to the content areas related to the collapsible/expandable buttons on the first panel. I didn't think it was necessary to add it elsewhere, and the context seems clear.

* Made sure that focus moves to any warnings/errors that are displayed in the UI (using tabindex="-1").

* Added table captions and removed the summaries (from what I read, summary was being used incorrectly).

I wasn't able to reproduce the "opening the section by pressing the button works, but closing it (pressing the button again) submits the form above" issue. I looked at the HTML, and the form element is properly closed. Let me know if you can reproduce it.

Thanks again for all the help!

**Comment by clrux** [comment]
@cahrens Very close.

* There's a "blank" element after each accordion button but before the content container
* I'm still randomly (accidentally) clicking the button in the form when closing buttons (Chrome on Mac only). I think it may be a focus struggle with VoiceOver and the OS focus.
* When VoiceOver is on, I can't press Enter to open the sections. I can when VoiceOver is off.

**Comment by cahrens** [comment]
@clrux and I had some quality time with voice tools. We found the following:

* There's a "blank" element after each accordion button but before the content container-- this is only on Windows and Firefox with NVDA. We don't experience it on Windows with IE/NVDA. We don't see anything in the HTML that could be causing this.
* I'm still randomly (accidentally) clicking the button in the form when closing buttons (Chrome on Mac only). I think it may be a focus struggle with VoiceOver and the OS focus--this appears to be a Chrome/Mac/VoiceOver interaction, and I don't see anything in the code that would cause it. As long as Safari and VoiceOver is working, I'm not inclined to continue struggling with it.
* When VoiceOver is on, I can't press Enter to open the sections. I can when VoiceOver is off-- @clrux said that Enter is not the way a VoiceOver user would select items. The more typical keystrokes do work. Again, there is nothing I am doing with a keylistener or HTML that should cause this. It may be related ot the last item.

File: openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html [hunk 4]
```diff
@@ -22,7 +22,7 @@ <h2 class="staff-info__title ui-staff__title">
                 <div class="staff-info__student">
                     <div class="wrapper--input" class="staff-info__student__form">
                         <form class="openassessment_student_info_form">
-                            <div class="form--error"></div>
+                            <div class="form--error" tabindex="-1"></div>
```
cahrens: Note to reviewers: -1 allows focus to be moved to item programmatically, but keeps the item out of the normal tab ordering.

openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html
```diff
@@ -3,14 +3,14 @@
 <div class="openassessment__staff-area wrapper--staff-area">
 
     <div class="wrapper--staff-toolbar">
-        <button class="ui-staff__button button-staff-tools" data-panel="openassessment__staff-tools">{% trans "Manage Individual Learners" %}</button>
-        <button class="ui-staff__button button-staff-info" data-panel="openassessment__staff-info">{% trans "View Assignment Statistics" %}</button>
+        <button class="ui-staff__button button-staff-tools" aria-expanded="false" data-panel="openassessment__staff-tools">{% trans "Manage Individual Learners" %}</button>
```
cahrens: Note to reviewers: for buttons that are triggering display of an item, aria-expanded helps screen reader users to understand that the button is behaving like a toggle.

openassessment/templates/openassessmentblock/staff_area/oa_staff_area.html
```diff
@@ -3,14 +3,14 @@
...
+        <button class="ui-staff__button button-staff-info" aria-expanded="false" data-panel="openassessment__staff-info">{% trans "View Assignment Statistics" %}</button>
         {% if staff_assessment_required %}
-            <button class="ui-staff__button button-staff-grading" data-panel="openassessment__staff-grading">{% trans "Grade Available Responses" %}</button>
+            <button class="ui-staff__button button-staff-grading" aria-expanded="false" data-panel="openassessment__staff-grading">{% trans "Grade Available Responses" %}</button>
         {% endif %}
     </div>
 
-    <div class="openassessment__staff-tools wrapper--staff-tools wrapper--ui-staff is--hidden">
```
cahrens: Note to reviewers: removal of "is--hidden" because collapsing/expanding is being done with JQuery slideUp/slideDown.

openassessment/templates/openassessmentblock/staff_area/oa_staff_grade_learners.html
```diff
@@ -1,19 +1,19 @@
 {% load i18n %}
 
-<div class="staff__grade__control ui-toggle-visibility is--collapsed">
```
cahrens: Note to reviewers: no more usage of "ui-toggle-visibility" in the staff grading tools (we will purge this from LMS portion in a later story). Instead we are using JQuery slideUp/slideDown which properly marks collapsed content as display none (while still supporting animation).

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -13,40 +13,40 @@ <h2 class="title">
     </h2>
 
     <div class="staff-info__status ui-staff__content__section wrapper--ui--collapse staff-info__student__response">
-        <div class="ui-staff ui-toggle-visibility is--collapsed">
-            <h2 class="staff-info__title ui-staff__subtitle ui-toggle-visibility__control">
+        <button class="ui-staff ui-slidable" aria-expanded="false" id="learner_response_{{ submission.uuid }}">
```
cahrens: Note to reviewers: just always use a button! The id is being used to link the button with its associated content, since it is less clear here than in the other places that "toggle" buttons are used. I'm using the submission.uuid to ensure unique IDs should there ever be 2 instances on the page at once.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -76,124 +76,123 @@ <h2 class="staff-info__title ui-staff__subtitle ui-toggle-visibility__control">
...
+            {% if workflow_status == "done" %}
+                <div class="staff-info__final__grade__score">
+                {% with points_earned_string=score.points_earned|stringformat:"s" points_possible_string=score.points_possible|stringformat:"s" %}
+                    {% blocktrans with points_earned='<span class="grade__value__earned">'|safe|add:points_earned_string|add:'</span>'|safe points_possible='<span class="grade__value__potential">'|safe|add:points_possible_string|add:'</span>'|safe %}
+                        Final grade: {{ points_earned }} out of {{ points_possible }}
+                    {% endblocktrans %}
+                {% endwith %}
+                </div>
+                <table class="staff-info__status__table staff-info__final__grade__table">
+                    <caption class="sr">{% trans "Final Grade Details" %}</caption>
```
cahrens: Note to reviewers: caption is needed for screen readers. And summary wasn't being used properly before (you don't want to have both repeating the same information).

openassessment/xblock/staff_area_mixin.py
```diff
@@ -236,8 +236,6 @@ def render_student_info(self, data, suffix=''):  # pylint: disable=W0613
         try:
             student_username = data.params.get('student_username', '')
             path, context = self.get_student_info_path_and_context(student_username)
-            expanded_view = data.params.get('expanded_view', [])
```
cahrens: Note to reviewers: I moved this functionality directly into the JavaScript instead of sending it to the server, setting classes in the HTML, and then having to deal with them in either CSS/JS.

openassessment/xblock/static/js/spec/lms/oa_staff_area.js
```diff
@@ -310,23 +315,23 @@ describe('OpenAssessment.StaffAreaView', function() {
...
 
-        it('shows the correct buttons with no panels initially', function() {
+        it('shows the correct buttons when full grading is not enabled', function() {
             var view = createStaffArea(),
                 $buttons = $('.ui-staff__button', view.element);
             expect($buttons.length).toBe(2);
+            expect($buttons).toHaveAttr('aria-expanded', 'false');
             expect($($buttons[0]).text().trim()).toEqual('Manage Individual Learners');
             expect($($buttons[1]).text().trim()).toEqual('View Assignment Statistics');
-            expect(getVisibleStaffPanels(view).length).toBe(0);
```
cahrens: Note to reviewers: the initial visibility is governed by the CSS, and for a Jasmine test, the CSS is not in play. Therefore we can't test this anymore.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -109,6 +114,33 @@
...
                         );
                     }
+
+                    $manageLearnersTab.find('.' + view.SLIDABLE_CLASS).click(
+                        function(event) {
+                            var $toggle = $(event.currentTarget),
+                                $content = $toggle.next('.' + view.SLIDABLE_CONTENT_CLASS);
+
+                            if ($toggle.hasClass(view.IS_SHOWING_CLASS)) {
+                                $toggle.removeClass(view.IS_SHOWING_CLASS).attr('aria-expanded', 'false');
```
cahrens: Note to reviewers: the "is--showing" class is necessary for the collapse/expand icon to update.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -125,27 +157,25 @@
          * Upon request, loads the staff grade/assessment section of the staff area.
          * This allows staff grading when staff assessment is a required step.
          *
-         * @param {boolean} clearAndCollapse if true, clear the staff grade form and collapse it. Otherwise
-         *     render the staff grade form if is not already loaded.
          * @returns {promise} A promise representing the successful loading
          * of the staff grade (assessment) section.
          */
-        loadStaffGradeForm: function(clearAndCollapse) {
+        loadStaffGradeForm: function() {
```
cahrens: Note to reviewers: I created a separate closeStaffGradeForm method, which simplified the code.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -214,32 +281,36 @@
...
-            var $staffTools = $('.openassessment__staff-tools', $staffArea);
-            var $staffInfo =  $('.openassessment__student-info', $staffArea);
+            var $manageLearnersTab = $('.openassessment__staff-tools', $staffArea);
             var $staffGradeTool = $('.openassessment__staff-grading', $staffArea);
 
             if ($staffArea.length <= 0) {
                 return;
             }
 
-            this.baseView.setUpCollapseExpand($staffTools, function() {});
```
cahrens: Note to reviewers: this was related to the old ui-toggle-visibility code. The method is still be used elsewhere, so it can't be deleted yet.

openassessment/xblock/static/sass/oa/elements/_staff.scss
```diff
@@ -17,34 +17,31 @@
     border-top: ($baseline-v/4) solid $color-decorative-staff;
     padding: $baseline-v ($baseline-h/2);
     background: $staff-bg;
+    display: none;
```
cahrens: Note to reviewers: for collapsible items, the display is initially set to none in the CSS (since they are initially collapsed), and then JQuery slideUp/slideDown changes the display as necessary.

openassessment/xblock/static/sass/oa/elements/_staff.scss
```diff
@@ -17,34 +17,31 @@
...
 
   .ui-staff {
 
-    // CASE: area is collapse/expand friendly
-    &.ui-toggle-visibility {
+    .ui-slidable {
 
-      .staff-info__title__copy {
-        @include margin-left(($baseline-h/4));
+      .icon {
```
cahrens: Note to reviewers: we used to inherit this icon CSS from the ui-toggle-visibility stuff. When we get to changing the student views, we may be able to promote some of this CSS out of _staff.scss.

test/acceptance/tests.py
```diff
@@ -500,10 +500,9 @@ def test_staff_area_panel(self, panel_name, button_label):
         # Click on the button and verify that the panel has opened
         self.staff_area_page.click_staff_toolbar_button(panel_name)
         self.assertEqual(self.staff_area_page.selected_button_names, [button_label])
-        self.assertIn(
-            u'openassessment__{button_name}'.format(button_name=panel_name),
-            self.staff_area_page.visible_staff_panels[0]
-        )
+        visible_panels = self.staff_area_page.visible_staff_panels
```
cahrens: Note to reviewers: no change in functionality, just checking length of visible panels.

openassessment/xblock/static/sass/oa/utilities/_developer.scss
```diff
@@ -101,6 +101,8 @@ $link-hover: $edx-blue-l1 !default; // from our Pattern Library http://ux.edx.or
 
     input {
       display: block;
+      padding-top: 0;
```
cahrens: Note to reviewers: this very targeted fix is to address TNL-4021 (a special gift for @robrap). The cause of the issue is some general styling in the "xb" directory for all inputs. The padding is set to 10 px (in all dimensions), causing the text to be cut off slightly on Firefox.

I don't want to spend long on this, and I was afraid of introducing other unknown issues if I changed the general rule. However, if we don't feel comfortable with this workaround, I'm happy to remove the commit from the PR.

As far as the "xb" directory, this is what @talbs says: "the xb stuff was a set of styling that I abstracted in case it was useful to other xblocks that would use Sass to compile their CSS
so its mostly general UI and base stlying
I don't think anyone ever documented that or leveraged it, so feel free to merge/combine".

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
efischer19: :+1: Thanks for the tip about `?w=1`, that made the diff way easier to see at a glance.

I also found this chrome plugin useful for wide files like the HTML in this PR, in case anyone else gets annoyed at left-right scrolling repeatedly.
https://chrome.google.com/webstore/detail/wide-github/kaalofacklcidaampbokdplbklpeldpj

openassessment/templates/openassessmentblock/staff_area/oa_student_info_assessment_detail.html
```diff
@@ -1,55 +1,56 @@
...
+            {% with assessment_num=forloop.counter %}
+                {% if assessments|length > 1 %}
+                    <h4 class="title--sub">
+                        {% blocktrans %}
+                            Assessment {{ assessment_num }}:
+                        {% endblocktrans %}
+                    </h4>
+                {% endif %}
+                <table class="staff-info__status__table">
+                    <caption class="sr">{% trans "Assessment Details" %}</caption>
```
efischer19: Looks like this display string has been modified ("Assessment" to "Assessment Details").

FYI @catong, in case there are screenshots or other documentation relying on this particular value.

cahrens: That's text that is only there for screen readers (see the "sr" class). Neither the caption, nor the summary (what was previously in the HTML) is visible to non-screen reader users, so there's nothing Carol needs to update.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -191,6 +230,34 @@
         },
 
         /**
+         * Closes the staff grade/assessment section of the staff area.
+         *
+         * @param {boolean} clear if true, remove the staff grade form and collapse it. Otherwise
```
efischer19: Tiny nit: Can we capitalize this first `if` here, or have some other way of indicating separation between the variable name and the following sentence? Reading `clear if true` tripped me up the first time I read it.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -290,7 +361,7 @@
             );
 
             // Install a click handler for rescheduling unfinished AI tasks for this problem
-            $staffTools.find('.action--submit-unfinished-tasks').click(
+            $manageLearnersTab.find('.action--submit-unfinished-tasks').click(
```
efischer19: Good renaming here

openassessment/xblock/static/sass/oa/utilities/_developer.scss
efischer19: :ship:

I like it, I think fixing the general rule here is more likely to accidentally fix this exact situation in other places than break them.

openassessment/templates/openassessmentblock/staff_area/oa_student_info_assessment_detail.html
efischer19: Ah, so it is. Thanks for the clarification!

efischer19:
This looks great, :+1: from me. One very small inline documentation nit if you feel it's worth fixing.

efischer19:
Also, thanks a bunch for all the inline "Notes for reviewers", they made the review much easier than it otherwise may have been.

openassessment/templates/openassessmentblock/staff_area/oa_student_info_assessment_detail.html
catong: Thanks! And for screen readers, "Assessment Details" is an improvement.

openassessment/xblock/static/js/src/lms/oa_staff_area.js
```diff
@@ -109,6 +114,33 @@
                             }
                         );
                     }
+
+                    $manageLearnersTab.find('.' + view.SLIDABLE_CLASS).click(
```
dianakhuang: I guess this is the best way to handle finding classes, but I'm not super happy with it..

dianakhuang:
:+1: Yeah, the notes were great.

cahrens:
@cptvitamin As we discussed, I have gone ahead and merged this PR. I am happy to follow up on any feedback as a separate PR.

My sandbox is still available, and the default ORA2 sandbox should pick up these changes tomorrow.

cptvitamin:
@cahrens you did a great job building this very complicated interface.  I have a few small bits of feedback for future work:

* It's kinda neat how @clrux recommended you reference the button as the label for the div.  However, regular old divs don't have accessibility properties (like labels).  On a mac, the presence of an accessibility attribute, like aria-labelledby, kicks off an heuristic designed to fix bad code, in a "I think I know what the dev was trying to do here" kinda way.  This only happens on MAC right now, so that label doesn't do anything on Windows. What we want to do here is make that a full accessible object, so that it works cross platform.  We do that by also adding ```role="group"``` to that div as well.  
* What we really wanted to add to button the expands and collapses those sections is add an ```aria-controls``` attribute that references the div, as in "this button controls the display of that div." 
* On the mac, with voiceover, I had an issue with the close buttons (which focus was moved to for the horizontal controls/tabs).  This is great practice.  The only problem was that the close button is the child of the Heading.  Headings with focusable content in them become "group"s roles on Mac.  As a user, I was confused to be thrown into the last item in a group like that.  When I tried to advance past the close button, I heard a "bonk" from the browser.  I had to reverse out of the group, to proceed.  Would it be possible to move the close button (ideally) before the Heading in the DOM, as its sibling?
* For the buttons that expand the vertical nav, I think there is a race condition happening. As a VO user, I could not access the content that was expanded unless I moved my focus elsewhere on the page and then tried to go back.  This can probably be fixed if we add a ```tabindex="-1"``` to those divs and then move focus to them after they expand.

The regular keyboard accessibility here worked great.  Nice job.  And thanks for adding all those comments for reviewers that may not understand accessibility.  It's a great way to continually teach/share/reinforce the concepts.

PR #855, opened 2016-01-21
Christina/fix pep8

cahrens:
@efischer19 and @dianakhuang please review when you have time-- no rush. I just couldn't resist reducing the number of violations somewhat....

Also, I found a bug with the pep8 script (see note below).

### DIFF
--- a/openassessment/xblock/submission_mixin.py
+++ b/openassessment/xblock/submission_mixin.py
@@ -435,8 +435,10 @@ class SubmissionMixin(object):
             student_submission = self.get_user_submission(
                 workflow["submission_uuid"]
             )
-            context["peer_incomplete"] = "peer" in workflow["status_details"] and not workflow["status_details"]["peer"]["complete"]
-            context["self_incomplete"] = "self" in workflow["status_details"] and not workflow["status_details"]["self"]["complete"]
+            peer_in_workflow = "peer" in workflow["status_details"]
+            self_in_workflow = "self" in workflow["status_details"]
+            context["peer_incomplete"] = peer_in_workflow and not workflow["status_details"]["peer"]["complete"]
+            context["self_incomplete"] = self_in_workflow and not workflow["status_details"]["self"]["complete"]
             context["student_submission"] = create_submission_dict(student_submission, self.prompts)
             path = 'openassessmentblock/response/oa_response_submitted.html'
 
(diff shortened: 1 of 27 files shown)


scripts/run-pep8.sh
```diff
@@ -10,7 +10,7 @@ pep8 --config=.pep8 openassessment > $PEP8_VIOLATIONS
 NUM_PEP8_VIOLATIONS=$(cat $PEP8_VIOLATIONS | wc -l)
 
 echo "Found" $NUM_PEP8_VIOLATIONS "pep8 violations, threshold is" $MAX_PEP8_VIOLATIONS
-if [[ $NUM_PEP8_VIOLATIONS > $MAX_PEP8_VIOLATIONS ]]; then
+if [[ $NUM_PEP8_VIOLATIONS -gt $MAX_PEP8_VIOLATIONS ]]; then
```
cahrens: Note to reviewers: apparently ">" does something different than greater than. It appeared to work with the number I tried, but then I hit a number that didn't work. 

I don't have experience writing bash scripts...

http://www.tldp.org/LDP/abs/html/comparison-ops.html

dianakhuang:
:+1:

openassessment/xblock/submission_mixin.py
```diff
@@ -435,8 +435,10 @@ def submission_path_and_context(self):
             student_submission = self.get_user_submission(
                 workflow["submission_uuid"]
             )
-            context["peer_incomplete"] = "peer" in workflow["status_details"] and not workflow["status_details"]["peer"]["complete"]
-            context["self_incomplete"] = "self" in workflow["status_details"] and not workflow["status_details"]["self"]["complete"]
+            peer_in_workflow = "peer" in workflow["status_details"]
+            self_in_workflow = "self" in workflow["status_details"]
+            context["peer_incomplete"] = peer_in_workflow and not workflow["status_details"]["peer"]["complete"]
+            context["self_incomplete"] = self_in_workflow and not workflow["status_details"]["self"]["complete"]
```
efischer19: Would these be "cleaner" with `get()` and default values?

`context["peer_incomplete"] = workflow["status_details"].get("peer"), {}).get("complete", False)`

I'm okay with this as it is here, just throwing a suggestion out there.

efischer19:
:+1: One minor suggestion, feel free to ignore if desired.

Hooray for code cleanup!

PR #857, opened 2016-01-22
Scope the submission response to the correct form elements.

dianakhuang:
A fix for: https://openedx.atlassian.net/browse/TNL-4025

The bug was caused by having multiple selectors on the page that matched ".submission__answer__part__text__value". The JS assumed that there's only one (the student answer textarea). When instructors fetched student submissions, the rendered submission also included a div with the class ".submission__answer__part__text__value". The auto-save code would then try to save the response, which would not be a valid response format, and the error would appear.

This fix will scope the response selector to only match the student area textarea.

Sandbox: https://dianakhuang.sandbox.edx.org

I attempted to write a Bok Choy test to test this fix, but I wasn't able to get one to fail appropriately on the older version of the code.

Reviews: @efischer19 @cahrens


cahrens:
@dianakhuang Can you add a description of what was causing the bug?

dianakhuang:
@cahrens Added it.

efischer19:
:+1: Looks good to me, let's make sure to manually verify when this goes to stage

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -247,7 +247,7 @@ OpenAssessment.ResponseView.prototype = {
      array of strings: The current response texts.
      **/
     response: function(texts) {
-        var sel = $('.submission__answer__part__text__value', this.element);
+        var sel = $('form.response__submission .submission__answer__part__text__value', this.element);
```
cahrens: Instead of relying on the form HTML component (which could appear anywhere) to provide the scoping, I'd feel better if the scoping was done via a class that is specific to the response HTML. For instance, could openassessment__response be used to scope this?

cahrens: Sorry, I didn't see response__submission. If you just get rid of the form part of the selector, :+1:

PR #860, opened 2016-01-27
Always calculate peer median

efischer19:
FYI @cahrens, this is super WIP currently.

I'm installing this on my sandbox now so we can test things out.

### DIFF
--- a/openassessment/xblock/grade_mixin.py
+++ b/openassessment/xblock/grade_mixin.py
@@ -333,7 +333,7 @@ class GradeMixin(object):
         # Fetch all the unique assessment parts
         criterion_name = criterion['name']
         staff_assessment_part = _get_assessment_part(_('Staff Grade'), criterion_name, staff_assessment)
-        if len(peer_assessments) > 0:
+        if "peer-assessment" in assessment_steps:
             peer_assessment_part = {
                 'title': _('Peer Median Grade'),
                 'criterion': criterion,
@@ -346,11 +346,6 @@ class GradeMixin(object):
                     )
                     for index, peer_assessment in enumerate(peer_assessments)
                 ],
-            }
-        elif "peer-assessment" in assessment_steps:
-            peer_assessment_part = {
-                'title': _('Peer Median Grade'),
-                'option': {'label': _('Waiting for peer reviews')}
             }
         else:
             peer_assessment_part = None
@@ -449,7 +444,12 @@ class GradeMixin(object):
         #  - the median score, and no explanation (it is too verbose to show an aggregate).
         options = median_options()
         if len(options) == 0:
-            return None
+            # If we weren't able to get a median option when there should be one, show the following message
+            # This happens when there are less than must_be_graded_by assessments made for the user
+            if len(criterion['options']) > 0:
+                return {'label': _('Waiting for peer reviews')}
+            else:
+                return None
         if len(options) == 1:
             return options[0]
         return {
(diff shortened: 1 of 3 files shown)


efischer19:
This seems to be working, I've verified on my sandbox. These screenshots are for a learner who:
1. Has finished their peer grading requirements by grading 2 other responses
2. Has been graded by 1 other learner, not the 2 required by must_be_graded_by
3. Has been given a staff grade

One question - should we suppress the 'additional comments for your response` section in these cases?
screen shot 2016-01-27 at 1 39 00 pm

Previous, incorrect behavior was:
screen shot 2016-01-27 at 11 45 08 am

Also, there is a discrepancy between what the learner sees (above screenshots) and what the staff tools for the learner show, should we reconcile this? New behavior:
screen shot 2016-01-27 at 1 44 02 pm

Previously in this scenario we showed this:
screen shot 2016-01-27 at 11 52 58 am

efischer19:
The reason staff tools were showing the incomplete values previously when the learner view wasn't is that the staff tools view was checking `individual_assessments`, not `median_score`.

Staff tools template: https://github.com/edx/edx-ora2/blob/master/openassessment/templates/openassessmentblock/staff_area/oa_student_info.html#L122

Learner grade complete template: https://github.com/edx/edx-ora2/blob/master/openassessment/templates/openassessmentblock/grade/oa_grade_complete.html#L51

efischer19:
@cahrens and @dianakhuang, could I get your input on the above?

I'm leaning towards suppressing the "additional comments for your response" section in the learner view if there's no grade to show.

For the staff tools, I could see either showing the "waiting for responses" message for a median score, or showing the individual assessments as before. I'd at least suggest that in the latter case we update the title of that column to "peer summary" or something, since it's not really a median score being shown.

efischer19:
@cahrens @dianakhuang This is working as far as I can tell, can I get a review on this PR if you get a chance?

cahrens:
It's looking good, but I will wait until the tests are written.

efischer19:
@cahrens @dianakhuang Test has been added, this is ready for final review.

Running the new test without the code fix fails:
```
File "/edx/src/edx-ora2/openassessment/xblock/test/test_grade.py", line 340, in test_peer_update_after_override
    self.assertTrue(all([option['label'] == u'Waiting for peer reviews' for option in peer_scores]))
TypeError: 'NoneType' object has no attribute '__getitem__'
```
As part of the code fix, it is now impossible to return an option without a `label`, as I've closed the codepath that was returning 'None'.

openassessment/xblock/grade_mixin.py
```diff
@@ -347,11 +347,10 @@ def _get_assessment_part(title, part_criterion_name, assessment):
                     for index, peer_assessment in enumerate(peer_assessments)
                 ],
             }
-        elif "peer-assessment" in assessment_steps:
-            peer_assessment_part = {
-                'title': _('Peer Median Grade'),
-                'option': {'label': _('Waiting for peer reviews')}
-            }
+            # If we weren't able to get a median option when there should be one, show the following message
```
cahrens: Nit: can this functionality move into _peer_median_option? It already has criterion, and then the contract will be option always returns a label.

cahrens:
:+1: I have one suggested refactoring, but otherwise looks great.

If you aren't in the habit already, check if the number of pep8/pylint violations decreased with your submission. If it did, please decrease the thresholds appropriately.

openassessment/xblock/grade_mixin.py
efischer19: Good idea! Changed.

cahrens:
The refactoring looks good to me-- :+1:

dianakhuang:
:+1: from me.

PR #862, opened 2016-02-02
Allow openassessment xblock to clear state

efischer19:
All the changes here are in support of one goal - openassessmentblock
should be able to provide a clear_state method which will:
* soft-delete any models that refer to a submission_uuid
* passthrough to the submissions api to delete the submission as well

Doing this will allow staff the ability to reset student state on ORA problems,
which is otherwise impossible. The changes in this commit are an entry point
in openassessmentblock, helpers in workflow_ and submission_mixin, and changes to
the workflow and assessment models and apis to allow them to be soft deleted.

Migrations are needed as a part of this change, they too are included.

Sibling change to https://github.com/edx/edx-platform/pull/11371 and https://github.com/edx/edx-submissions/pull/33

### DIFF
--- a/openassessment/assessment/api/ai.py
+++ b/openassessment/assessment/api/ai.py
@@ -15,7 +15,7 @@ from openassessment.assessment.errors import (
 from openassessment.assessment.models import (
     Assessment, AITrainingWorkflow, AIGradingWorkflow,
     InvalidRubricSelection, NoTrainingExamples,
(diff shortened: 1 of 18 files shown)


efischer19:
Will need to generate migrations, forgot that

efischer19:
Finally working on devstack, looks like this will fix things.

PR #865, opened 2016-02-05
Declare all packages.

cahrens:
I declared all packages except for "migrations" and "commands".


cahrens:
@efischer19 and @nedbat I'm going to go ahead and merge this so I can re-tag and hopefully still get the ORA update in before today's release candidate.

We can further refine the list of packages in the next release, if we wish.

PR #866, opened 2016-02-05
Update bok choy tests to have unique emails.

cahrens:
@efischer19 and @dianakhuang Please review. Note that I need the email in order to log back in as an existing user-- that's why I was using a specific email in the past. Now I am returning the generated unique email from the auto auth code.

All tests passed with these changes on my sandbox.

### DIFF
--- a/test/acceptance/auto_auth.py
+++ b/test/acceptance/auto_auth.py
@@ -82,11 +82,25 @@ class AutoAuthPage(PageObject):
 
     def get_username(self):
         """
-        Finds and returns the username
+        Finds and returns the username of the current user.
+        """
+        username, _ = self._get_username_and_email()
+        return username
+
+    def get_email(self):
+        """
+        Finds and returns the email address of the current user.
+        """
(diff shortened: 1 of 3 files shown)


test/acceptance/auto_auth.py
```diff
@@ -82,11 +82,24 @@ def get_user_id(self):
...
+        """
+        Finds and returns the email address of the current user.
+        """
+        username_and_email = self._get_username_and_email()
+        return username_and_email.split(' ')[1]
+
+    def _get_username_and_email(self):
+        """
+        Finds and returns a string containing username and email address of the current user.
         """
```
dianakhuang: I think it makes sense to return the username and email as a tuple instead of as a string.

cahrens: It was just an internal helper method, but I will change it.

dianakhuang:
Other than that one comment, :+1:

efischer19:
Looks great! :+1:

PR #869, opened 2016-02-10
Change package declaration so that templates load.

cahrens:
Changing to find_packages instead of the explicit listing of packages allows the templates to load (without -e). I tested this with https://github.com/edx/edx-platform/pull/11512, building a sandbox (orae.sandbox.edx.org) with the edx-platform branch and both manually testing and running the ORA acceptance tests against the sandbox.

I also found that I could leave the package listing as it was and include the HTML files via package_data. But given that most of our repos just use find_packages, I decided to go with the simpler approach.

@nedbat and @efischer19 please review.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,20 +1,5 @@
 #!/usr/bin/env python
-from setuptools import setup
-
-PACKAGES = [
-    'openassessment',
-    'openassessment.assessment',
-    'openassessment.assessment.api',
-    'openassessment.assessment.errors',
-    'openassessment.assessment.models',
-    'openassessment.assessment.serializers',
-    'openassessment.assessment.worker',
-    'openassessment.fileupload',
-    'openassessment.fileupload.backends',
-    'openassessment.workflow',
-    'openassessment.management',
-    'openassessment.xblock'
-]
+from setuptools import setup, find_packages
 
 def is_requirement(line):
     """
@@ -60,7 +45,7 @@ setup(
         'Operating System :: OS Independent',
         'Programming Language :: Python',
     ],
-    packages=PACKAGES,
+    packages=find_packages(exclude=["test", "tests"]),
     install_requires=load_requirements('requirements/base.txt', 'requirements/wheels.txt'),
     tests_require=load_requirements('requirements/test.txt'),
     entry_points={


nedbat:
I don't understand how this makes the templates loadable.  I thought the problem was that the .html files weren't included in the kit.  This change doesn't put the HTML files in the kit.  I tried building a tar file with ``$ python setup.py sdist``, and looking at the contents of the file, there are no HTML files.

Is it possible the sandbox already had the template files?

cahrens:
@nedbat I don't know. I built the sandbox from scratch so I thought that would be sufficient, but perhaps not? This is what I did:

1) I left the packages definition as it was and added package_data with .html as a file extension. I created a sandbox with this and tested-- everything worked.

2) I then decided to use find_packages to remove the explicit packages, and out of curiosity I decided to try first with package_data removed. I expected that to fail, but a new sandbox (spun up from scratch) worked. 

Do you think package_data is the right way to go?

cahrens:
BTW, this was the failure the last time around:

File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/xblock/runtime.py", line 806, in render
    frag = view_fn(context)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/openassessment/xblock/openassessmentblock.py", line 361, in student_view
    template = get_template("openassessmentblock/oa_base.html")
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/template/loader.py", line 46, in get_template
    raise TemplateDoesNotExist(template_name)
TemplateDoesNotExist: openassessmentblock/oa_base.html

nedbat:
(I don't know anything about sandboxes, somehow I have never used one here, so I don't know what happens when making one.)

My test is to use setup.py to make a tar file, and then examine that tar file:
```
$ cd blah/blah/edx-ora2
$ python setup.py sdist
$ tar tvfz sdist/ora2-0.2.7.tar.gz
drwxr-xr-x  0 ned    wheel       0 Feb 11 14:24 ora2-0.2.7/
-rw-r--r--  0 ned    wheel     867 Dec 30 13:37 ora2-0.2.7/AUTHORS
-rw-r--r--  0 ned    wheel   35135 Sep 12  2014 ora2-0.2.7/LICENSE
-rw-r--r--  0 ned    wheel     287 Sep 12  2014 ora2-0.2.7/MANIFEST.in
drwxr-xr-x  0 ned    wheel       0 Feb 11 14:24 ora2-0.2.7/openassessment/
-rw-r--r--  0 ned    wheel       0 Sep 12  2014 ora2-0.2.7/openassessment/__init__.py
drwxr-xr-x  0 ned    wheel       0 Feb 11 14:24 ora2-0.2.7/openassessment/assessment/
-rw-r--r--  0 ned    wheel       0 Sep 12  2014 ora2-0.2.7/openassessment/assessment/__init__.py
-rw-r--r--  0 ned    wheel    5328 Sep 12  2014 ora2-0.2.7/openassessment/assessment/admin.py
...
...
```

nedbat:
I believe "packages" is just for Python code.  "package_data" or "data_files" are for other kinds of files, like templates.

cahrens:
OK, I'll try package_data and do your test to make sure the HTML files appear.

(how do you review open source PRs without making a sandbox running the code?)

nedbat:
I triage them, and let the engineering team do the real review.

PR #870, opened 2016-02-11
more test info

benpatterson:


benpatterson:
@nasthagiri @cahrens @robrap 

This should give more granular test data. We can potentially add the junit publish plugin once this merges to master. (Note that this particular plugin will mean we can't run jobs concurrently, so we also do not have to use it. We use a different plugin on the public Jenkins.)

nasthagiri:
lgtm, but I'll defer to @cahrens on this.

cahrens:
:+1: Thanks for continuing to investigate, @benpatterson 

FYI @efischer19 and @dianakhuang

PR #871, opened 2016-02-12
Update to latest nose, bok-choy.

benpatterson:
@dianakhuang @efischer19 

Main motivation: the newer bok-choy will let us specify which version of firefox we want to use in tests. That is something that could help out with local troubleshooting.


dianakhuang:
:+1:

efischer19:
Looks great :+1: 

image

PR #875, opened 2016-02-18
Move ORA data download logic into ora2 repo.

dianakhuang:
Necessary for this ORA data download PR: https://github.com/edx/edx-platform/pull/11167

https://openedx.atlassian.net/browse/TNL-1402

### DIFF
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -3,10 +3,11 @@ Aggregate data for openassessment.
 """
 import csv
 import json
+
 from django.conf import settings
 from submissions import api as sub_api
 from openassessment.workflow.models import AssessmentWorkflow
-from openassessment.assessment.models import AssessmentPart, AssessmentFeedback
+from openassessment.assessment.models import Assessment, AssessmentPart, AssessmentFeedback
 
 
 class CsvWriter(object):
@@ -114,15 +115,15 @@ class CsvWriter(object):
             # so we select AssessmentPart and follow the foreign key to the Assessment.
             parts = self._use_read_replica(
                 AssessmentPart.objects.select_related('assessment', 'option', 'option__criterion')
-                    .filter(assessment__submission_uuid=submission_uuid)
-                    .order_by('assessment__pk')
+                .filter(assessment__submission_uuid=submission_uuid)
+                .order_by('assessment__pk')
             )
             self._write_assessment_to_csv(parts, rubric_points_cache)
 
             feedback_query = self._use_read_replica(
                 AssessmentFeedback.objects
-                    .filter(submission_uuid=submission_uuid)
-                    .prefetch_related('options')
+                .filter(submission_uuid=submission_uuid)
+                .prefetch_related('options')
             )
             for assessment_feedback in feedback_query:
                 self._write_assessment_feedback_to_csv(assessment_feedback)
@@ -164,8 +165,8 @@ class CsvWriter(object):
             end = start + self.QUERY_INTERVAL
             query = self._use_read_replica(
                 AssessmentWorkflow.objects
-                    .filter(course_id=course_id)
-                    .order_by('created')
+                .filter(course_id=course_id)
+                .order_by('created')
             ).values('submission_uuid')[start:end]
 
             for workflow_dict in query:
@@ -334,3 +335,163 @@ class CsvWriter(object):
             if "read_replica" in settings.DATABASES
             else queryset
         )
+
+
+class OraAggregateData(object):
+    """
+    Aggregate all the ORA data into a single table-like data structure.
+    """
+
+    @classmethod
+    def _use_read_replica(cls, queryset):
+        """
+        If there's a read replica that can be used, return a cursor to that.
+        Otherwise, return a cursor to the regular database.
+
+        Args:
+            queryset (QuerySet): The queryset that we would like to use the read replica for.
+        Returns:
+            QuerySet
+        """
+        return (
+            queryset.using("read_replica")
+            if "read_replica" in settings.DATABASES
+            else queryset
+        )
+
+    @classmethod
+    def _build_assessments_cell(cls, assessments):
+        """
+        Args:
+            assessments (QuerySet) - assessments that we would like to collate into one column.
+        Returns:
+            string that should be included in the 'assessments' column for this set of assessments' row
+        """
+        returned_string = u""
+        for assessment in assessments:
+            returned_string += u"Assessment #{}\n".format(assessment.id)
+            returned_string += u"-- scored_at: {}\n".format(assessment.scored_at)
+            returned_string += u"-- type: {}\n".format(assessment.score_type)
+            returned_string += u"-- scorer_id: {}\n".format(assessment.scorer_id)
+            if assessment.feedback != u"":
+                returned_string += u"-- overall_feedback: {}\n".format(assessment.feedback)
+        return returned_string
+
+    @classmethod
+    def _build_assessments_parts_cell(cls, assessments):
+        """
+        Args:
+            assessments (QuerySet) - assessments containing the parts that we would like to collate into one column.
+        Returns:
+            string that should be included in the relevant 'assessments_parts' column for this set of assessments' row
+        """
+        returned_string = u""
+        for assessment in assessments:
+            returned_string += u"Assessment #{}\n".format(assessment.id)
+            for part in assessment.parts.all():
+                returned_string += u"-- {}".format(part.criterion.label)
+                if part.option is not None and part.option.label is not None:
+                    option_label = part.option.label
+                    returned_string += u": {option_label} ({option_points})\n".format(
+                        option_label=option_label, option_points=part.option.points
+                    )
+                if part.feedback != u"":
+                    returned_string += u"-- feedback: {}\n".format(part.feedback)
+        return returned_string
+
+    @classmethod
+    def _build_feedback_options_cell(cls, assessments):
+        """
+        Args:
+            assessments (QuerySet) - assessment that we would like to use to fetch and read the feedback options.
+        Returns:
+            string that should be included in the relevant 'feedback_options' column for this set of assessments' row
+        """
+
+        returned_string = u""
+        for assessment in assessments:
+            for feedback in assessment.assessment_feedback.all():
+                for option in feedback.options.all():
+                    returned_string += option.text + u"\n"
+
+        return returned_string
+
+    @classmethod
+    def _build_feedback_cell(cls, submission_uuid):
+        """
+        Args:
+            submission_uuid (string) - the submission_uuid associated with this particular assessment feedback
+        Returns:
+            string that should be included in the relevant 'feedback' column for this set of assessments' row
+        """
+        try:
+            feedback = AssessmentFeedback.objects.get(submission_uuid=submission_uuid)
+        except AssessmentFeedback.DoesNotExist:
+            return u""
+        return feedback.feedback_text
+
+    @classmethod
+    def collect_ora2_data(cls, course_id):
+        """
+        Query database for aggregated ora2 response data.
+
+        Args:
+            course_id (string) - the course id of the course whose data we would like to return
+
+        Returns:
+            A tuple containing two lists: headers and data.
+
+            headers is a list containing strings corresponding to the column headers of the data.
+            data is a list of lists, where each sub-list corresponds to a row in the table of all the data
+                for this course.
+
+        """
+
+        all_submission_information = sub_api.get_all_course_submission_information(course_id, 'openassessment')
+
+        rows = []
+        for student_item, submission, score in all_submission_information:
+            row = []
+            assessments = cls._use_read_replica(
+                Assessment.objects.prefetch_related('parts').
+                prefetch_related('rubric').
+                filter(
+                    submission_uuid=submission['uuid']
+                )
+            )
+            assessments_cell = cls._build_assessments_cell(assessments)
+            assessments_parts_cell = cls._build_assessments_parts_cell(assessments)
+            feedback_options_cell = cls._build_feedback_options_cell(assessments)
+            feedback_cell = cls._build_feedback_cell(submission['uuid'])
+
+            row = [
+                submission['uuid'],
+                submission['student_item'],
+                student_item['student_id'],
+                submission['submitted_at'],
+                submission['answer'],
+                assessments_cell,
+                assessments_parts_cell,
+                score.get('created_at', ''),
+                score.get('points_earned', ''),
+                score.get('points_possible', ''),
+                feedback_options_cell,
+                feedback_cell
+            ]
+            rows.append(row)
+
+        header = [
+            'Submission ID',
+            'Item ID',
+            'Anonymized Student ID',
+            'Date/Time Response Submitted',
+            'Response',
+            'Assessment Details',
+            'Assessment Scores',
+            'Date/Time Final Score Given',
+            'Final Score Points Earned',
+            'Final Score Points Possible',
+            'Feedback Statements Selected',
+            'Feedback on Peer Assessments'
+        ]
+        return header, rows
--- a/openassessment/management/commands/collect_ora2_data.py
+++ b/openassessment/management/commands/collect_ora2_data.py
@@ -1,0 +1,73 @@
+"""
+Command to retrieve all ORA2 data for a course in a .csv.
+
+
+This command differs from upload_oa_data in that it places all the data into one file.
+
+Generates the same format as the instructor dashboard downloads.
+"""
+import csv
+from optparse import make_option
+import os
+
+from django.core.management.base import BaseCommand, CommandError
+
+from openassessment.data import OraAggregateData
+
+
+class Command(BaseCommand):
+    """
+    Query aggregated open assessment data, write to .csv
+    """
+
+    help = ("Usage: collect_ora2_data <course_id> --output-dir=<output_dir>")
+    args = "<course_id>"
+
+    option_list = BaseCommand.option_list + (
+        make_option('-o', '--output-dir',
+                    action='store', dest='output_dir', default=None,
+                    help="Write output to a directory rather than stdout"),
+        make_option('-n', '--file-name',
+                    action='store', dest='file_name', default=None,
+                    help="Write CSV file to the given name"),
+    )
+
+    def handle(self, *args, **options):
+        """
+        Run the command.
+        """
+        if not args:
+            raise CommandError("Course ID must be specified to fetch data")
+
+        course_id = args[0]
+
+        if options['file_name']:
+            file_name = options['file_name']
+        else:
+            file_name = ("%s-ora2.csv" % course_id).replace("/", "-")
+
+        if options['output_dir']:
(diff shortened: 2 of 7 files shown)


dianakhuang:
@efischer19 and @robrap I believe this is ready for review.

openassessment/data.py
```diff
@@ -333,4 +337,133 @@ def _use_read_replica(self, queryset):
...
+        """
+        Query MySQL database for aggregated ora2 response data.
+        """
+        cursor = cls._read_replica_cursor_if_available()
+
+        # Syntax unsupported by other vendors such as SQLite test db
+        if db.connection.vendor != 'mysql':
+            return '', ['']
+
+        raw_queries = cls.RAW_QUERY.split(';')
```
efischer19: I think a comment here explaining why we have 2 different queries would be good

openassessment/data.py
```diff
@@ -333,4 +337,133 @@ def _use_read_replica(self, queryset):
...
+
+        cursor.execute(raw_queries[0])
+        cursor.execute(raw_queries[1], [course_id])
+
+        header = [item[0] for item in cursor.description]
+        return header, cursor.fetchall()
+
+    # Raw MySQL query that this CsvWriter
```
efischer19: Incomplete comment?

openassessment/management/commands/collect_ora2_data.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    def handle(self, *args, **options):
+        """
+        Run the command.
+        """
+        if not args:
+            raise CommandError("Course ID must be specified to fetch data")
+
+        course_id = args[0]
```
efischer19: Is there a way we can validate this near the start of the management command, and error out if it's not valid?

efischer19:
A few small issues, looks great overall. I definitely think we can get this into today's ORA release.

openassessment/management/commands/collect_ora2_data.py
dianakhuang: The problem is that ORA2 doesn't really have a concept of what a valid or invalid course id is. It's just a string in the code right now. I wasn't sure I wanted to add opaque-keys as a dependency just for this one minor use case.

efischer19: Ah, I see. Makes sense then.

efischer19:
:+1:

openassessment/data.py
```diff
@@ -333,4 +337,135 @@ def _use_read_replica(self, queryset):
...
+
+class AggregateOraData(object):
+    """
+    Aggregate all the ORA data into a single table.
+
+    Only works on MySQL databases.
+    """
+
+    @classmethod
+    def _read_replica_cursor_if_available(cls):
```
robrap: Maybe call this `_read_replica_or_regular_cursor`?  The name as-is is slightly vague about what it returns in the case it is not available.

openassessment/data.py
```diff
@@ -333,4 +337,135 @@ def _use_read_replica(self, queryset):
...
+    @classmethod
+    def collect_ora2_data(cls, course_id):
+        """
+        Query MySQL database for aggregated ora2 response data.
+        """
+        cursor = cls._read_replica_cursor_if_available()
+
+        # Syntax unsupported by other vendors such as SQLite test db
+        if db.connection.vendor != 'mysql':
+            return '', ['']
```
robrap: Should this be an exception?

dianakhuang: I'm not sure. That was the behavior in the original PR and I wasn't sure if there was anything that was relying on it.

openassessment/data.py
```diff
@@ -333,4 +337,135 @@ def _use_read_replica(self, queryset):
...
+
+        # The raw query has two statements separated by a semi-colon.
+        # We would like to run them separately through the Django ORM.
+        raw_queries = cls.RAW_QUERY.split(';')
```
robrap: Any reason not to split RAW_QUERY into 2 variables with a reasonable name for each?

openassessment/data.py
```diff
@@ -333,4 +337,135 @@ def _use_read_replica(self, queryset):
...
+) AS `feedback_options`,
+(
+    SELECT `feedback_text`
+    FROM `assessment_assessmentfeedback` as `feedback`
+    WHERE `feedback`.`submission_uuid`=`sub`.`uuid`
+    LIMIT 1
+) AS `feedback`
+FROM `submissions_submission` AS `sub`
+JOIN `submissions_studentitem` AS `student` ON `sub`.`student_item_id`=`student`.`id`
+WHERE `student`.`item_type`="openassessment" AND `student`.`course_id`=%s
```
dianakhuang: @ormsbee Here's the end of the giant query. Enjoy?

openassessment/data.py
```diff
@@ -333,4 +337,135 @@ def _use_read_replica(self, queryset):
...
+    ON `feedback`.`id`=`feedback_join`.`assessmentfeedback_id`
+    WHERE `feedback`.`submission_uuid`=`sub`.`uuid`
+) AS `feedback_options`,
+(
+    SELECT `feedback_text`
+    FROM `assessment_assessmentfeedback` as `feedback`
+    WHERE `feedback`.`submission_uuid`=`sub`.`uuid`
+    LIMIT 1
+) AS `feedback`
+FROM `submissions_submission` AS `sub`
```
robrap: We have the open question about whether or not we are ok with the fact that this query ignores the boundary between ORA and the Submissions API.  We discussed performance testing it as is to see if it is a non-issue (i.e. fails performance testing).

openassessment/data.py
robrap: If we could determine and at least comment the mysql specific syntax, that might be helpful in the future.

openassessment/data.py
robrap: Do we need to filter out data with status deleted from the submissions api, or should that already be properly handled by relying on valid ORA data?  You may want to test Clear Student State with the data download to verify.

efischer19: This depends on the use case - the whole point of doing soft-deletes was to keep the data available for later use. Using SQL directly will not filter out the deleted entries unless `submissions.status=D` is added to this query.

dianakhuang:
@robrap could you take a look at this and let me know what you think of the refactor work?

openassessment/data.py
```diff
@@ -114,15 +115,15 @@ def write_to_csv(self, course_id):
             # so we select AssessmentPart and follow the foreign key to the Assessment.
             parts = self._use_read_replica(
                 AssessmentPart.objects.select_related('assessment', 'option', 'option__criterion')
-                    .filter(assessment__submission_uuid=submission_uuid)
-                    .order_by('assessment__pk')
+                .filter(assessment__submission_uuid=submission_uuid)
```
robrap: I personally prefer this as it was with the indent.  Do you mind showing me where we state to do otherwise, if we do?  Thanks.

openassessment/data.py
```diff
@@ -333,4 +334,151 @@ def _use_read_replica(self, queryset):
             queryset.using("read_replica")
             if "read_replica" in settings.DATABASES
             else queryset
-        )
\ No newline at end of file
+        )
+
+
+class AggregateOraData(object):
```
robrap: What do you think of ``OraAggregateData``?  The original tripped me up because it read like an "Or".  Also, it sounds a bit more like an action than a noun as it stands.

openassessment/data.py
dianakhuang: This is a pep8 rule that we don't have exceptions for. I was just cleaning up the file on my way through.

openassessment/data.py
```diff
@@ -333,4 +334,151 @@ def _use_read_replica(self, queryset):
...
+    """
+    Aggregate all the ORA data into a single table.
+
+    Only works on MySQL databases.
```
robrap: MySQL only?  Should this comment be deleted?

openassessment/data.py
```diff
@@ -333,4 +334,149 @@ def _use_read_replica(self, queryset):
...
             if "read_replica" in settings.DATABASES
             else queryset
-        )
\ No newline at end of file
+        )
+
+
+class OraAggregateData(object):
+    """
+    Aggregate all the ORA data into a single table.
```
dianakhuang: rename `table` to something clearer

openassessment/data.py
```diff
@@ -333,4 +334,149 @@ def _use_read_replica(self, queryset):
...
+            QuerySet
+        """
+        return (
+            queryset.using("read_replica")
+            if "read_replica" in settings.DATABASES
+            else queryset
+        )
+
+    @classmethod
+    def _build_assessments_column(cls, assessments):
```
dianakhuang: Rename column -> cell

dianakhuang:
@efischer19 This PR has been updated.

dianakhuang:
@robrap I updated those names.

openassessment/data.py
```diff
@@ -333,4 +334,151 @@ def _use_read_replica(self, queryset):
...
+            returned_string = assessments[0].assessment_feedback.all()[0].feedback_text
+        return returned_string
+
+    @classmethod
+    def collect_ora2_data(cls, course_id):
+        """
+        Query MySQL database for aggregated ora2 response data.
+        """
+
+        all_submission_information = sub_api.get_all_course_submission_information(course_id, 'openassessment')
```
efischer19: Note for myself so I can find the submissions integration point quickly when I need it again

openassessment/data.py
```diff
@@ -333,4 +334,151 @@ def _use_read_replica(self, queryset):
...
+            row = []
+            assessments = cls._use_read_replica(
+                Assessment.objects.prefetch_related('parts').
+                prefetch_related('rubric').
+                filter(
+                    submission_uuid=submission['uuid']
+                )
+            )
+            assessments_cell = cls._build_assessments_cell(assessments)
+            assessments_parts_Cell = cls._build_assessments_parts_cell(assessments)
```
efischer19: Inconsistent casing, can this be `assessments_parts_cell`?

openassessment/data.py
```diff
@@ -333,4 +334,151 @@ def _use_read_replica(self, queryset):
...
+        """
+        returned_string = u""
+        for assessment in assessments:
+            returned_string += u"Assessment #{}\n".format(assessment.id)
+            returned_string += u"-- scored_at: {}\n".format(assessment.scored_at)
+            returned_string += u"-- type: {}\n".format(assessment.score_type)
+            returned_string += u"-- scorer_id: {}\n".format(assessment.scorer_id)
+            if assessment.feedback != u"":
+                returned_string += u"-- overall_feedback: {}\n".format(assessment.feedback)
+        return returned_string
```
efischer19: :+1: I'm a fan of this system of including multiple assessments' data in a single cell in the resulting table.

openassessment/tests/test_data.py
```diff
@@ -152,3 +225,108 @@ def _load_fixture(self, fixture_relpath):
...
+            'Assessment Details',
+            'Assessment Scores',
+            'Date/Time Final Score Given',
+            'Final Score Points Earned',
+            'Final Score Points Possible',
+            'Feedback Statements Selected',
+            'Feedback on Peer Assessments'
+        ])
+
+        self.assertEqual(data[1], [
```
efischer19: Maybe I'm confused but why are we using `data[1]` here? Isn't `submission` created before `scorer_submission`? Or is the returned ordering not FIFO with respect to when the submissions were made?

efischer19:
:+1: once we get the ordering issue sorted out, on to edx-submissions!

openassessment/data.py
```diff
@@ -333,4 +334,152 @@ def _use_read_replica(self, queryset):
...
+        try:
+            feedback = AssessmentFeedback.objects.get(submission_uuid=submission_uuid)
+        except AssessmentFeedback.DoesNotExist:
+            return u""
+        return feedback.feedback_text
+
+    @classmethod
+    def collect_ora2_data(cls, course_id):
+        """
+        Query MySQL database for aggregated ora2 response data.
```
robrap: Anything MySQL specific left, or can this just say "Query database..." ?

openassessment/data.py
```diff
@@ -333,4 +334,152 @@ def _use_read_replica(self, queryset):
...
+        """
```
robrap: Args and Returns?

openassessment/management/commands/collect_ora2_data.py
```diff
@@ -0,0 +1,67 @@
...
+
+This command differs from upload_oa_data in that it places all the data into one file.
+
+Generates the same format as the instructor dashboard downloads.
+"""
+import csv
+from optparse import make_option
+import os
+
+from openassessment.data import OraAggregateData
```
robrap: This import section (line) should be at the bottom.

openassessment/management/commands/collect_ora2_data.py
```diff
@@ -0,0 +1,67 @@
...
+    def handle(self, *args, **options):
+        """
+        Run the command.
+        """
+        if not args:
+            raise CommandError("Course ID must be specified to fetch data")
+
+        course_id = args[0]
+
+        file_name = ("%s-ora2.csv" % course_id).replace("/", "-")
```
robrap: It feels like someone should be able to supply a file name to override this default.

openassessment/management/commands/collect_ora2_data.py
```diff
@@ -0,0 +1,67 @@
...
+        writer = csv.writer(csv_file, dialect='excel', quotechar='"', quoting=csv.QUOTE_ALL)
+
+        header, rows = OraAggregateData.collect_ora2_data(course_id)
+
+        writer.writerow(header)
+        for row in rows:
+            writer.writerow(_preprocess(row))
+
+
+def _preprocess(data_list):
```
robrap: More specific:
```
def _encode_row(row):
```

openassessment/management/commands/collect_ora2_data.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Properly encode ora2 responses for transcription into a .csv
+    """
+    processed_data = []
```
robrap: processed_row (to go with above).

openassessment/management/tests/test_collect_ora2_data.py
```diff
@@ -0,0 +1,71 @@
...
+            "assessments",
+            "assessments_parts",
+            "final_score_given_at",
+            "final_score_points_earned",
+            "final_score_points_possible",
+            "feedback_options",
+            "feedback",
+        ]
+
+        self.test_row = [
```
robrap: Is this actually test_rows?  Also, I'd have 2 rows, otherwise same test would pass even if you didn't loop.

openassessment/management/tests/test_collect_ora2_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+            mock_writerow.assert_any_call(self.test_header)
+            mock_writerow.assert_called_with(self.test_row[0])
+
+            mock_writerow.mock_calls = []
+
+            call_command('collect_ora2_data', self.COURSE_ID)
+
+            mock_writerow.assert_any_call(self.test_header)
+            mock_writerow.assert_called_with(self.test_row[0])
```
robrap: Seems like this is implicitly, but not explicitly testing the `unicode(item).encode('utf-8')`.  Should you add an assertion on the type of one of the items in the row before and after?  Does it require any special unicode characters in the test data as well?

openassessment/tests/test_data.py
```diff
@@ -152,3 +225,124 @@ def _load_fixture(self, fixture_relpath):
         )
         print "Loading database fixtures from {}".format(fixture_path)
         call_command('loaddata', fixture_path)
+
+
+class TestOraAggregateData(TransactionCacheResetTest):
+    """
+    Test that the component parts of OraAggregateData behave as expected.
+    """
```
robrap: Maybe mention that this is an integration test, testing against the real submissions api.

openassessment/tests/test_data.py
```diff
@@ -152,3 +225,124 @@ def _load_fixture(self, fixture_relpath):
...
+            u'',
+            u'',
+            u'',
+            u'',
+            u'',
+            u'',
+            u'',
+        ])
+
+        self.assertEqual(data[1], [
```
robrap: Seems like this doesn't give full code coverage.  Some examples:
- multiple assessments,
- without feedback
- without option label
There may be other cases.

robrap:
@dianakhuang Overall nice work.  Thank you.  It is clean, and above all, it separates the submissions api!  Just minor naming, commenting, testing feedback.

dianakhuang:
@robrap @efischer19 The last two commits should have the new tests and more addressing of code review feedback. When you get a chance, could you guys take another look?

openassessment/management/tests/test_collect_ora2_data.py
```diff
@@ -0,0 +1,101 @@
...
+        with patch('openassessment.management.commands.collect_ora2_data.csv') as mock_write:
+            call_command('collect_ora2_data', self.COURSE_ID)
+
+            mock_writerow = mock_write.writer.return_value.writerow
+
+            mock_writerow.assert_any_call(self.test_header)
+            mock_writerow.assert_any_call(self.test_rows[0])
+            mock_writerow.assert_any_call(self.unicode_encoded_row)
+
+            mock_writerow.mock_calls = []
```
robrap: 1. I'm not following this.  It looks like the same test, with a different mock_writerow, and the same assertions for both tests.  
2. I was imagining a mock implementation of writerow which performs an assertion on the input to verify the list contains objects of type str.

robrap: 1. I'm not following this.  It looks like the same test, with a different mock_writerow, and the same assertions for both tests.  
2. I was imagining a mock implementation of writerow which performs an assertion on the input to verify the list contains objects of type str.

robrap:
Thanks for all the work @dianakhuang.  I didn't realize the mock code would be so extensive.  I appreciate the effort, and hopefully it could also get used for other tests in the future.

openassessment/management/tests/test_collect_ora2_data.py
dianakhuang: It is the same test, but now it's using two rows instead of one, and one of the rows contains unicode that we are checking is encoded properly. I don't know why the original test insisted on running the command twice, but I left it there in case there was a specific error condition they were checking for. I don't think it's there to check a different case.

robrap: I forgot you didn't write the original test.  I'm almost thinking we should remove the second call_command if we don't see any good reason to have it, and no way to comment why it exists.

I saw the unicode encoded row.  I wasn't clear on how that would work with this assertion, but I think I see now.  Did you verify that if you stop the encoding inside encode_row() this test fails?

Thanks for explaining.

dianakhuang: I just double-checked that pulling out `.encode('utf-8')` caused a failure in that line, and it does. I can remove the second `call_command` in the test because it really does seem redundant.

efischer19:
Still :+1: from me

robrap:
:+1: Nice job.  Thank you.

PR #878, opened 2016-02-19
Hide screen reader value

efischer19:
This is a fix for [TNL-4009](https://openedx.atlassian.net/browse/TNL-4009)

~~The text was appearing oddly here, and it looks as if a very similar value close by in the template is being used as screen reader only (`.submission__answer__part__prompt__title` at line 482 in `_oa-base.scss`, versus `.submission__answer__part__text__title` which I just added at line 503). See https://github.com/edx/edx-ora2/blob/master/openassessment/templates/openassessmentblock/oa_submission_answer.html for the template that defines these classes. No other style references this `..._text_title` class, so I think it just got overlooked. The cleanest fix seems to be to apply the `%text-sr` style here as well.~~

~~The `oa_submission_answer.html` template is used elsewhere, so this change will hide a bit more text than just in the leaderboard. That text looked fairly unstyled though, so I still think it was always intended to be hidden as screen-reader text. I'll add some screenshots shortly.~~


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

dianakhuang:
:+1:

efischer19:
Here are some screenshots of the "not just in the leaderboard" text being changed.

Before, note how "Your response to the question above" seems unstyled:
screen shot 2016-02-19 at 2 42 53 pm

After, no text. Note that "Your Response" is still present (and styled!) above the prompt:
screen shot 2016-02-19 at 2 45 11 pm

robrap:
@efischer19 I'll need to review the original ticket, but your images make me question this change.  It doesn't seem to be an improvement to me.

cahrens:
I agree with Robert. The text adds clarity for all users, so we should style it correctly and always show it (in my opinion).

efischer19:
Tagging @chris-mike, this is the PR being discussed.

openassessment/templates/openassessmentblock/oa_submission_answer.html
```diff
@@ -4,13 +4,13 @@
...
             <article class="submission__answer__part__prompt">
-                <h2 class="submission__answer__part__prompt__title">{% trans "The question for this section." %}</h2>
+                <h3 class="submission__answer__part__prompt__title">{% trans "The question for this section." %}</h3>
                 <div class="submission__answer__part__prompt__value">
                     {{ part.prompt.description|linebreaks }}
                 </div>
             </article>
             <div class="submission__answer__part__text">
-                <h2 class="submission__answer__part__text__title">{{ answer_text_label }}</h2>
+                <h3 class="submission__answer__part__text__title">{{ answer_text_label }}</h3>
```
efischer19: I kept running into an overriding style coming from https://github.com/edx/edx-platform/blob/master/lms/static/sass/course/courseware/_courseware.scss#L453 for the `h2` element here.

Is changing it to an `h3` an acceptable workaround? Is there some other way to say "don't apply that style in this case" in CSS?

efischer19:
New screenshots:

Non-leaderboard:
screen shot 2016-02-22 at 12 59 03 pm

Leaderboard:
screen shot 2016-02-22 at 12 59 25 pm

FYI @chris-mike @robrap @dianakhuang

openassessment/templates/openassessmentblock/oa_submission_answer.html
cahrens: Keep it as an h2, but override the rule with a more explicit rule.

openassessment/xblock/static/sass/oa/utilities/_shame.scss
```diff
@@ -71,7 +71,16 @@
...
+  // submission answer leaderboard fix (TNL-4009). 4 class selectors is more specific than
+  // lms/static/sass/course/courseware/_courseware.scss#L453 in edx-platform (3 class selectors)
+  .leaderboard__answer .submission__answer__display__content {
+    h2.submission__answer__part__text__title {
+      @include text-align(left);
+      @include float(none);
+      font-size: inherit;
+      margin: 0;
+    }
+  }
```
efischer19: If `_shame.scss` isn't the right place for this code, let me know. This case seems to fit "rules to overcome any poorly defined, scoped, or architected edX LMS-based CSS" as described above.

openassessment/xblock/static/sass/oa/views/_oa-base.scss
```diff
@@ -500,6 +500,10 @@
     }
   }
 
+  .submission__answer__part__text__title {
+    @extend %t-subheading;
```
efischer19: This is what's causing the bold labels - is that acceptable? I think it looks better than before.

efischer19:
Updated to avoid `s/h2/h3` from before. Screenshots are still the same (https://github.com/edx/edx-ora2/pull/878#issuecomment-187296560), but the style definition is now made in a targeted `shame.scss` rule.

openassessment/xblock/static/sass/oa/utilities/_shame.scss
cahrens: I don't see anything particularly shameful about this css. I would probably put it in _oa-base.scss, but I don't feel strongly about it.

efischer19:
Sandbox has been rebuilt, check https://efischer19.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/ee5a35f57dce4b9facd7c29ca06e4597/ for a live demo.

@andy-armstrong, would you mind looking over this fix? I chose `_shame.scss` due to the existence of this (https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/static/sass/oa/utilities/_shame.scss#L39), which seems similar.

cahrens:
:+1: @dianakhuang should also re-review since the code changed quite a bit since her initial thumbs.

dianakhuang:
:+1:

PR #877, opened 2016-02-19
Remove ruby sass compilation

efischer19:
Inspired by my desire to stop getting errors when re-installing ora on a fresh devstack. After ruby was removed from edx-platform, I was getting errors on `gem install sass` in `Makefile`, so I decided to try and copy the work done in https://github.com/edx/edx-platform/commit/c13e62306760b892e3b257fe042dc5b27a9c1d4b in this repo.

I've attempted to keep all the functionality the same, and only had a make a few minor changes (`s/Calc/calc`) to get compilation to succeed. Feel free to correct anything I may have done improperly!

@andy-armstrong @ormsbee Could you review this when you have a minute? No rush at all.

FYI @dianakhuang @robrap @benpatterson

PR #879, opened 2016-02-19
Fix for leaderboard image bug

efischer19:
Fix for https://openedx.atlassian.net/browse/TNL-3115. See https://efischer19.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/4f70056660ac490ab6ff253fe62a18b1/ for a live example.

This is a simple one line fix, can I get a quick review @robrap @dianakhuang? The `css` and `css.map` files attached to this PR may change, depending on which order this and https://github.com/edx/edx-ora2/pull/878 get merged, but the end result will be "make both `scss` changes and re-compile sass".

Screenshots:
screen shot 2016-02-19 at 3 36 35 pm

After the fix is applied:
screen shot 2016-02-19 at 4 35 21 pm


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

efischer19:
@robrap @dianakhuang I got some clarity from @andy-armstrong regarding the `css.map` files (https://github.com/edx/edx-ora2/pull/877#issuecomment-187249557), and have implemented his suggestion here.

This PR is ready and able to be merged as it currently stands, can I get thumbs on the diff?

robrap:
:+1:   

It would be nice to get a thumb from @andy-armstrong to confirm that what you did was what he intended.  Could/should the ignore be more globally applied for that file type?

andy-armstrong:
I was intrigued as to why this image was floated left, so I tracked down the PR that introduced it. It appears that it has always been this way:

https://github.com/edx/edx-ora2/pull/615

I'm somewhat nervous that there was a reason for it being floated, but clearly it doesn't wrap well in this situation. Choosing not to float it might mean that the design doesn't look as good for some existing courses. @simultech would you be able to help us understand what the impact would be of this change?

andy-armstrong:
:+1: for the .gitignore change. I would also be fine with ignoring all .map files as we never want to check them in.

efischer19:
I've updated the `.gitignore` to ignore all map files.

@andy-armstrong I don't know that I'm as concerned about breaking existing courses with this change. We're changing a fairly narrow class style (`leaderboard__score__image`), which is only used in one location in ORA code (https://github.com/edx/edx-ora2/blob/master/openassessment/templates/openassessmentblock/leaderboard/oa_leaderboard_show.html#L21). What could course authors or learners have done that would be affected by this change, since we define the HTML that uses it?

andy-armstrong:
@efischer19 It just means that courses that use this feature (probably the University of Queensland since they added it) will find that their score image appears on its own line, rather than floated to the left. I'd like to understand if that's an issue before merging this.

efischer19:
Ah, okay I understand.

efischer19:
As @simultech seems to not be terribly active of late, I'll also tag @ankithkonda from https://github.com/UQ-UQx.

@ankithkonda, we're proposing the removal a `float: left` css rule from leaderboard images in openassessment problems. This was introduced in https://github.com/edx/edx-ora2/pull/615, but the layout is buggy and affecting active edx.org courses. Will removing this rule cause issues with anything UQx is doing?

simultech:
Sorry for the delay in replying, I am no longer with UQx (although still keep in contact with them).  The float left was part of the original feature, and if I remember correctly it was to responsible for text responses to float next to the image (before submission__answer__display__content had a background I would assume).  

As long as there are no regressions in the leaderboard layout from removing the rule (which there shouldn't be if its causing issues in running courses), then go for it (UQx doesn't do any special with the leaderboard setup so they will probably experience similar issues in the future).

efischer19:
Excellent, thanks for the reply @simultech!

PR #885, opened 2016-02-26
Revert package name change

efischer19:
This recent change broke requirements.txt installs


efischer19:
Merging without thumbs

PR #888, opened 2016-03-01
Fix leaderboard PDFs

efischer19:
As Shelby noted on TNL-3115, pdf files do not appear properly on the leaderboard, because the OSPR that added images to the leaderboard used raw `img` tags instead of our existing template.

This is a quick fix that seems to handle things.
Before screenshot:
screen shot 2016-03-01 at 5 12 13 pm

After screenshot:
screen shot 2016-03-01 at 5 12 45 pm

You'll notice I also moved the image to be below the response instead of above the prompt; it seems more logical to me that way. If no one else agrees, we can move it back quite easily.

Seen on my sandbox at https://efischer19.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/8725631f67fa4b1099136f1196782767/

@dianakhuang @cahrens Can you give this a quick review?

PR #890, opened 2016-03-02
Updates to account for a11y CSS changes

efischer19:
@cahrens @andy-armstrong @clrux  Making changes to account for platform-wide changes made recently.

I've removed `%t-subheading`, as the only difference between it and `%t-heading` is the use of `%t-titlecase`, which, according to @clrux, is not something we should be doing in headings. I also updated our tests to no longer be expecting ALL_CAPS.

The other question I wanted to potentially address in this PR is the size changing of many of our `h3` elements. `message__title`, for example (https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/static/sass/xb/elements/_system-feedback.scss#L18), extends `%t-heading`, which styles it as `%hd-4`. However, many/all of the locations where we use this class, it is being set on an `h3` element (https://github.com/edx/edx-ora2/search?p=2&q=h3&utf8=%E2%9C%93). The result is that our h3 (previously styled as an h4) is now properly styled as an h3, which makes it appear too large and doesn't look great.
screen shot 2016-03-02 at 10 08 31 am

How should we address this? It feels wrong to have logical h3 elements styled as h4, should we change the elements to match the style (using h4 everywhere), or change the style to match the elements (update `%t-heading` to use h3)?


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

cahrens:
I agree that styling a h3 as a h4 seems wrong. Would the HTML still be logical if we changed them to h4? Looking at the HTML, we would then jump from h2 to h4, which I believe is a no-no.

@clrux can you chime in on what we should do about h3 elements that are now being styled as "too large"?

@efischer19 can you attach a sandbox?

efischer19:
efischer19.sandbox.edx.org has a recent version of `edx-platform` with the new CSS installed, it's where I got the above screenshot. So if you want to see the h3->h4 issues, those are present.

I'll update its ORA version to match this PR after I get https://github.com/edx/edx-ora2/pull/888 fixed up and merged.

andy-armstrong:
The Pattern Library has a section on headings, but it doesn't seem to provide any guidance here. The examples are still using h1, h2 etc as well. We need to define this more clearly for everyone, and then we should update the Pattern Library documentation to match:

http://ux.edx.org/#headings

clrux:
@cahrens @efischer19 The beauty of the UXPL is that you can use any heading style that fits your needs while maintaining a semantic structure. If you need a smaller H3, you can try `<h3 class="hd hd-5">` which uses a smaller style.

efischer19:
@clrux Won't any style we define in the xblock be overridden by this line, due to the use of `!important`? https://github.com/edx/edx-platform/blob/master/lms/static/sass/xmodule/_headings.scss#L117

clrux:
@efischer19 Not any style, just one that affects font weight. Pulling out that `!important` may be good, but it may open up room for inconsistencies. Are you wanting to alter the weight of the heading?

efischer19:
I'm not sure, I guess so? I'm just trying to reconcile what we had before (h3 styled as h4, all caps) with the new changes that have been introduced (no more all caps transform, forced h3 font-weight). @cahrens and @andy-armstrong, what do you think? Or @chris-mike, can you weigh in? Are the (smaller) pre-existing UX styles something that should be overriding the new (larger) a11y styles?

Sandbox https://efischer19.sandbox.edx.org/ is updated with this branch if anyone wants to test live.

Before screenshot, taken from stage.edx.org:
screen shot 2016-03-02 at 2 18 13 pm

After screenshot, taken from my sandbox:
screen shot 2016-03-02 at 2 18 54 pm

chris-mike:
@efischer19 Comparing old to new: color is same, which is good. The size changed (we went bigger) and use of caps (no longer caps).  My understanding with the latter, is that all caps are not ideal for accessibility (per @clrux ).    As for the size, I'd suggest making it smaller (perhaps hd-4 or hd-5).  I don't think we need a 1:1 match up between old and new; visual consistency is more important (across and within pages).

efischer19:
@chris-mike Size-wise, we are currently (in ORA) styling these elements as `hd-4` (https://github.com/edx/edx-ora2/blob/efischer/css_fixes/openassessment/xblock/static/sass/oa/utilities/_extends.scss#L16).

The issue is that due to these new a11y rules, our `h3` stylings are being overridden. They are getting this style https://github.com/edx/edx-platform/blob/master/lms/static/sass/xmodule/_headings.scss#L115, which sets `font-size: 1.5em !important;` as part of the extended `hd-2` style (https://github.com/edx/edx-platform/blob/master/lms/static/sass/xmodule/_headings.scss#L34).

So if a11y styles say one thing, and ours say another, should we defer to theirs? Or try to override their `!important`?

efischer19:
Heads up @clrux, we're likely going to want to get this into the RC before the release goes live. I'll come over to the UX/a11y pod early Monday to make sure everyone is on board and ready to go with these changes.

openassessment/xblock/static/sass/oa/utilities/_shame.scss
```diff
@@ -3,6 +3,16 @@
 
 // NOTES: use for any bad-form/orphaned scss that knowingly violate edX FED architecture/standards (see - http://csswizardry.com/2013/04/shame-css)
 
+// Hacky a11y overrides
+// First, these elements, while logically h3, look really odd when styled in the large h3 style
+#seq_content .wrapper--xblock .openassessment h3.response__submission__status__title{
+  font-size: 12px !important;
```
clrux: Is the `!important` necessary?

openassessment/xblock/static/sass/oa/utilities/_shame.scss
```diff
@@ -3,6 +3,16 @@
...
+}
+// This one - we were getting <header><h3/></header> elements with 2 "font-size: 1.5em;" rules
+#seq_content .wrapper--xblock .openassessment header{
+  font-size: em(16) !important;
```
clrux: Is the `!important` necessary?

clrux:
@efischer19 If the `!importants` you're using are meant to override the `!important` I added, I'd prefer to remove all of them, including my own. If this needs to happen I can take care of it on Monday, first-thing.

efischer19:
@clrux The `!important` I'm overriding is for `font-size`, not `font-weight`. It's being inherited from `%hd-2` through this `extends`:  https://github.com/edx/edx-platform/blob/master/lms/static/sass/xmodule/_headings.scss#L116

openassessment/xblock/static/sass/oa/utilities/_shame.scss
efischer19: This one is different (`header` vs `h3`), let me see if removing `!important` works cleanly here.

efischer19:
Late-breaking update after speaking with @cptvitamin - it's very odd for that tiny status label to be an `h3` anyway, so I'm going to change it and that ought to remove the interplay. @clrux No action needed on your end right now, I'l touch base in a bit.

clrux:
@efischer19 I still favor removing any `!importants` even if it doesn't affect your work. I've created a [new PR](https://github.com/edx/edx-platform/pull/11760) that reflects this work.

efischer19:
@clrux I think I've got everything cleared up here, there was already a rule for `.self-` and `.peer-assessment__display__header` in `_shame.scss`, so I added `.staff-assessment__display__header` and that fixed the "h3 within a header" sizing issue I was seeing.

@cptvitamin @cahrens Could I bother you two for a quick review on this, so I can get this change into this week's RC? The 4 main issues I've addressed are:
- Removing h3 element `response__submission__status__title`, this is more properly just a div (thanks @cptvitamin!)
- Removing style `%t-subheading`, as `%t-titlecase` breaks a11y so `%t-heading` may now be used everywhere as that was the only differentiator
- Adding `staff-assessment__display__header` to _shame.css overrides
- Modifying acceptance tests to account for loss of `%t-titlecase`

openassessment/xblock/static/sass/oa/utilities/_shame.scss
```diff
@@ -11,7 +11,7 @@
   // rules to overcome any poorly defined, scoped, or architected edX LMS-based CSS. Ideally this file will be removed when the LMS Front End is re-visited.
 
   // headers
-  .self-assessment__display__header, .peer-assessment__display__header, .step__header {
+  .self-assessment__display__header, .peer-assessment__display__header, .step__header, .staff-assessment__display__header {
```
clrux: Is there any way to move these styles into the ORA2 repo? The `shame` partial was originally meant to be tentative and "shame" developers who used it :)

cahrens:
@efischer19 There are tests failing.

openassessment/xblock/static/sass/oa/utilities/_shame.scss
cahrens: @clrux They are already in the ORA repo. In my opinion, refactoring the ORA CSS files to get rid of _shame.scss is beyond the scope of this change.

clrux: Is there any way to add this CSS without adding it into the `shame` partial? Is there another partial it can go into instead?

cahrens: I think it belongs with these other header overrides because it was just an oversight when staff-assessment__display__header was added that it also wasn't added to this rule.

clrux: OK

efischer19:
Tests have been fixed up. All my sandboxes went down over the weekend, so I'm in the process of rebuilding them.

efischer19:
Sandbox is up at https://efischer19.sandbox.edx.org/

openassessment/templates/openassessmentblock/response/oa_response.html
```diff
@@ -99,22 +99,22 @@ <h3 class="message__title">{% trans "We could not save your progress" %}</h3>
...
                                 </div>
                             </li>
                             {% if allow_latex %}
                                 <li class="list--actions__item">
-                                    <button type="submit" id="submission__preview" class="action action--save submission__save is--disabled">{% trans "Preview in LaTeX"%}</button>
-                                    <div id="response__preview_explanation" class="response__submission__status">
-                                        <h3 class="response__submission__status__title">
-                                            {% trans "Click to preview your submission in LaTeX"%}
-                                        </h3>
+                                    <button type="submit" id="submission__preview" class="action action--save submission__save is--disabled" aria-describedby="respons__preview_explanation">
```
cahrens: Typo here.

cahrens:
This UI needs to be fixed--
image

cahrens:
In general, I'm really not happy with how the titles look now. They look overly large are seem very out of context with the surrounding elements (that are all much smaller). @chris-mike are you comfortable with how this looks? I'd really like to see smaller font sizes.

Examples:
image

image

image

image

efischer19:
My latest modification (`h3.staff__grade__title` in `_shame.scss`) fixes the staff grading bug @cahrens noted above.

screen shot 2016-03-07 at 12 23 56 pm

openassessment/xblock/static/sass/oa/utilities/_shame.scss
```diff
@@ -2,6 +2,10 @@
 // ====================
 
 // NOTES: use for any bad-form/orphaned scss that knowingly violate edX FED architecture/standards (see - http://csswizardry.com/2013/04/shame-css)
+#seq_content .wrapper--xblock .openassessment h3.staff__grade__title {
```
cahrens: @clrux Why did you use an ID on the original rule that Eric is trying to override? I thought we generally avoid IDs for CSS and just rely on classes (especially since IDs can force this sort of ugliness for overriding, and IDs also have uniqueness constraints).

clrux: My aim was to sync up heading styles across LMS and CMS for the main content area after removing H1s and H2s as options from the visual editor. Due to the overall messiness of the platform styles, I had to either use an !important to ensure things were consistent, or be more specific in targeting.

We should discuss this, perhaps with Christine. Without my specificity, styles aren't really cohesive. With the specificity then Eric must be even more specific. Not a win-win in either scenario.

efischer19:
Merging per verbal thumbs from @chris-mike and @clrux, going to try to get this into today's RC double-quick.

Sandbox is updated. I'll add a new ticket for the UXPL work @clrux and I were discussing.

efischer19:
Follow-up ticket added https://openedx.atlassian.net/browse/TNL-4208

PR #891, opened 2016-03-02
Use passed-in requesting_user_id

efischer19:
This pairs with the changes being made in https://github.com/edx/edx-platform/pull/11666

### DIFF
--- a/openassessment/xblock/staff_area_mixin.py
+++ b/openassessment/xblock/staff_area_mixin.py
@@ -492,7 +492,7 @@ class StaffAreaMixin(object):
                 'msg': self._(u"An error occurred while rescheduling tasks: {}".format(ex))
             }
 
-    def clear_student_state(self, user_id, course_id, item_id):
+    def clear_student_state(self, user_id, course_id, item_id, requesting_user_id):
         """
         This xblock method is called (from our LMS runtime, which defines this method signature) to clear student state
         for a given problem. It will cancel the workflow using traditional methods to remove it from the grading pools,
@@ -509,7 +509,7 @@ class StaffAreaMixin(object):
         submissions = submission_api.get_submissions(student_item)
         for sub in submissions:
             # Remove the submission from grading pools
-            self._cancel_workflow(sub['uuid'], "Student state cleared")
+            self._cancel_workflow(sub['uuid'], "Student state cleared", requesting_user_id=requesting_user_id)
 
             # Tell the submissions API to orphan the submission to prevent it from being accessed
             submission_api.reset_score(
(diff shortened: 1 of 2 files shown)


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

openassessment/xblock/staff_area_mixin.py
```diff
@@ -547,17 +547,20 @@ def cancel_submission(self, data, suffix=''):  # pylint: disable=W0613
...
         """
         Internal helper method to cancel a workflow using the workflow API.
+
+        If requesting_user is not provided, we will use the contextually-current staff user.
         """
         try:
             assessment_requirements = self.workflow_requirements()
-            student_item_dict = self.get_student_item_dict()
+            if requesting_user_id is None:
+                requesting_user_id = self.get_student_item_dict()["student_id"]
```
robrap: This doesn't seem like "the contextually-current staff user".

openassessment/xblock/test/test_staff_area.py
```diff
@@ -770,7 +770,7 @@ def test_staff_delete_student_state(self, xblock):
         self.assertIn("final grade", resp.body.lower())
 
         # Clear the submission
-        xblock.clear_student_state('Bob', 'test_course', xblock.scope_ids.usage_id)
+        xblock.clear_student_state('Bob', 'test_course', xblock.scope_ids.usage_id, STUDENT_ITEM['student_id'])
```
robrap: Shouldn't this be a staff user?

openassessment/xblock/staff_area_mixin.py
efischer19: `get_student_item_dict()` gets information about the contextually current user (the user bound to this xblock), which happens to be staff when this is called. I'll try to clarify the comment.

openassessment/xblock/test/test_staff_area.py
efischer19: Yes, in this instance the test is specifically checking "Can a staff user clear state on themselves".

Since `clear_student_state` has (after much debate) been allowed to proceed as an unbound xblock method call, the xblock has no way of verifying that the user passed in is staff; the assumption is that the runtime has secured any code that will call this method. Since our test is acting as the runtime, the simplest test is to use the staff user `Bob` we've already created.

dianakhuang:
:+1: from me.

openassessment/xblock/staff_area_mixin.py
robrap: Add comment:
The student_id is actually the bound user, which is the staff user in this context.

efischer19:
For posterity:  This PR will be merged using second set of :+1: from @robrap initially given on https://github.com/edx/edx-ora2/commit/378e44e

PR #894, opened 2016-03-10
Add support for multiple ORAs on a unit

mushtaqak:
If a unit has more than one ORA problems then students could not submit their peer assessments on the second ORA problem. 

**Issue**
The problem lied with the fetching of submission uuid which isn't ORA/Xblock specific. So whenever submission uuid is retrieved, we always get the uuid of the first ORA problem, therefore, we get an error message :

> We could not submit your assessment. This feedback has already been submitted or the submission has been cancelled.

**Solution**
Create a `getUsageID` method in `oa_base.js` that can be inherited by child views. And get usage ID on demand.

[TNL-4155](https://openedx.atlassian.net/browse/TNL-4155)


mushtaqak:
@cahrens Do we need an integration test for this? 

Also, with current implementation I am geeting it hard to write a jasmine test that has two ORA blocks as it is not supported in the fixtures and html templates. Any thoughts would be appreciated? Thanks

@adampalay FYI

mushtaqak:
By integration, I meant writing test in edx-platform.

adampalay:
@mushtaqak , are there other acceptance tests for ORA in edx-platform? Actually, writing a test for a problem with two xblocks on a page isn't a pattern I can seem to find anywhere in edx-platform, and was something I was running into in https://openedx.atlassian.net/browse/TNL-4217

cahrens:
You should add an integration test for 2 ORA components, but in the ORA repo, where all the other ORA Bok Choy tests live.

test/acceptance/tests.py
```diff
@@ -75,6 +75,9 @@ class OpenAssessmentTest(WebAppTest):
         'full_workflow_staff_required':
             u'courses/{test_course_id}/courseware/'
             u'8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb/'.format(test_course_id=TEST_COURSE_ID),
+        'multiple_ora':
+            u'courses/{test_course_id}/courseware/'
+            u'8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb'.format(test_course_id=TEST_COURSE_ID),
```
efischer19: Is there a reason for re-using the location of the `full_workflow_staff_required` problem? I feel that this would be better served as its own problem.

Also note that if you've changed the test course locally/on your sandbox for this test, your PR needs to include the changes to https://github.com/edx/edx-ora2/blob/master/scripts/data/test-course.tar.gz. (If you haven't changed the test course, ignore the rest of this comment.) This could be why you're having trouble with tests locally - if your local course is changed it could affect pre-existing tests.

Once you've got the course exactly as you like, export it and save to `scripts/data/test-course.tar.gz` in your ora2 git branch. When you push that change up, please note exactly what changed in an inline comment on the PR, as the file is compressed and it isn't easy to see what's changed by looking at the diff.

Before merging, please ensure that:
- [ ] The sandbox you use to run acceptance tests in jenkins has your course updates.
- [ ] A jenkins run of acceptance tests using this branch and pointed at that sandbox succeeds.
- [ ] This PR includes any changes you've made to the test course (such as adding a multiple assessment page)

mushtaqak: @efischer19 Yes you are right. I was having difficulties reproducing the the problems structure as it is currently in test course which seems outdated see [TNL-4371](https://openedx.atlassian.net/browse/TNL-4371)

But I got past through that somehow. I have added the new unit page containing multiple ora problems.

mushtaqak:
@efischer19 My test passed on local. I cleaned up all the ORA and the course and then started afresh. I have created the sandbox and added the latest copy of my course and updating the sandbox to point to my branch and run acceptance tests. 

Lets see how it goes.

scripts/test-acceptance.sh
```diff
@@ -53,4 +53,4 @@ if [[ "${test_name}" = "accessibility" ]]; then
 fi
 
 echo "Running acceptance tests from ${test_name}.py against the sandbox..."
-nosetests ../acceptance/${test_name}.py --with-xunit --xunit-file=../acceptance/xunit-${test_name}.xml
+nosetests ../acceptance/${test_name}.py:MultipleOpenAssessmentTest --with-xunit --xunit-file=../acceptance/xunit-${test_name}.xml
```
efischer19: @mushtaqak This is a good idea for debugging your new tests on jenkins quickly, but please be sure to have a full run of all tests (the changes to this file backed out) before merging.

The `No such test MultipleOpenAssessmentTest` failure you're seeing is happening because jenkins runs acceptance and accessibility tests, and there is no `MultipleOpenAssessmentTest` for accessibility.

mushtaqak: @efischer19 Yes, I know I had to drop the `scripts/test-acceptance.sh` change. I just looked at your PR and I thought this is faster. And I didn't knew that we have to write both acceptance and accessibility tests ?

For every acceptance, do we need accessibility test ?

test/acceptance/tests.py
```diff
@@ -75,6 +75,9 @@ class OpenAssessmentTest(WebAppTest):
         'full_workflow_staff_required':
             u'courses/{test_course_id}/courseware/'
             u'8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb/'.format(test_course_id=TEST_COURSE_ID),
+        'multiple_ora':
+            u'courses/{test_course_id}/courseware/'
+            u'2ff3b8d6b539411cacf3fa8a4a0acbe7/552253acd723478e96cdbea605c37469/'.format(test_course_id=TEST_COURSE_ID),
```
mushtaqak: Added this new unit containing multiple ORA blocks in the `test-course.tar.gz`

scripts/test-acceptance.sh
cahrens: We don't need an accessibility test for every acceptance test. But we should have an accessibility test for having 2 problems on the same page-- it will catch non-unique IDs, among other issues. It would be very helpful to know what issues such a test catches, but we may not be able to submit it until more issues are fixed.

Can you please try adding the accessibility test and let us know what fails?

efischer19: It's not a hard-and-fast rule, but if we're adding new functionality that isn't tested in accessibility tests already, we should be adding a test to verify it.

In this case, I'd recommend adding one similar to https://github.com/edx/edx-ora2/blob/master/test/acceptance/accessibility.py#L217, but pointed at your new problem.

mushtaqak: @efischer19 Thanks. I already did. I just wanted to know. 

@cahrens I couldn't agree more.

mushtaqak:
@cahrens I am getting following errors on output of accessibility test
http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/827/testReport/junit/acceptance.accessibility/MultipleOpenAssessmentA11yTest/test_multiple_ora_complete_flow/

So do I need to make these HTML elements' ids unique ? what is the best way to get this done ?

cahrens:
@mushtaqak thank you for getting that result. I think fixing all the IDs is out of the scope of your work, and there is already an open bug for it.  https://openedx.atlassian.net/browse/TNL-1593 I included the results of your test run in it.

In the meantime, please disable that specific rule, but only for your test. We don't want that rule disabled in general. You may be able to scope the amount of the page for which the rule is disabled as well-- please check with @clytwynec if you have questions.

mushtaqak:
@cahrens, @efischer19  May you please review.

Acceptance tests passing at [jenkins run](http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/837/)

However, I am waiting for @clytwynec to guide me to disable specific validation rule for my accessibility test.

mushtaqak:
To run the acceptance test on your local devstack, checkout my branch and run `make test-acceptance` command.

clytwynec:
@mushtaqak Can you point me toward which test you need help with?

mushtaqak:
@clytwynec [MultipleOpenAssessmentA11yTest](https://github.com/edx/edx-ora2/pull/894/files#diff-5af5cc11bcd9b5b162cfef228cbfcf0eR246) accessibility test to be specific :)

clytwynec:
I'm not sure I fully understand what the question is, but I'll try to answer.  The rules are typically disabled per page via `self.whatever_page.a11y_audit.set_rules`, not per test.  In this use case, they are being disabled on a per test level via the `_check_a11y` helper method.  So, if you want to disable the rule for all of the pages in the test, you will need to override the [`_check_a11y` method](https://github.com/edx/edx-ora2/pull/894/files#diff-5af5cc11bcd9b5b162cfef228cbfcf0eR18) in MultipleOpenAssessmentA11yTest.  If you are trying to disable the rule for only one of the pages being checked in this test, you'll have to directly interact with `self.whatever_page.a11y_audit` instead of using `_check_a11y()` for that page.   Does that answer your question?

mushtaqak:
@clytwynec I think so. So I have to override the `_check_a11y` to ignore the HTML elements that were giving error output like this?  :
```
page.a11y_audit.config.set_rules({
             "ignore": [
                 "color-contrast",  # TODO: AC-198
                 "empty-heading",  # TODO: AC-197
                 "link-href",  # TODO: AC-199
                 "link-name",  # TODO: AC-196
                 "skip-link",  # TODO: AC-179
             ]
         })
```

mushtaqak:
Like for this error, I have to add `openassessment` in the ignore list of the `page.a11y_audit.config.set_rules({ .. ` method. Right @clytwynec 
```
Rule ID: duplicate-id
Help URL: https://dequeuniversity.com/rules/axe/1.1/duplicate-id

	Message: Document has multiple elements with the same id attribute: openassessment
	Html: <div class="openassessment problem" id="openassessment">
	Target: [u'#seq_content > .xblock.xblock-student_view.xblock-student_view-vertical.xblock-
		initialized > .vert-mod > .vert.vert-0 > .xblock.xblock-student_view.xblock-student_view-
		openassessment.xblock-initialized > .wrapper.wrapper--xblock.wrapper--openassessment.theme--basic
		> .openassessment.problem']
```

So the fix would be like :

```
page.a11y_audit.config.set_rules({
             "ignore": [
                 "color-contrast",  # TODO: AC-198
                 "empty-heading",  # TODO: AC-197
                 "link-href",  # TODO: AC-199
                 "link-name",  # TODO: AC-196
                 "skip-link",  # TODO: AC-179,
                 "openassessment", # ignored my openassessment div error
             ]
         })
```

Please correct me if I am wrong.

clytwynec:
Close.  The `set_rules` method would take a rule id (in this case `duplicate-id`).  To ignore the div, you could use `set_scope` with `#openassessment` in the 'exclude' list.  That would ignore _all_ rules for that div though, meaning it wouldn't check any accessibility on it.

test/acceptance/tests.py
```diff
@@ -1102,6 +1105,43 @@ def test_train_self_peer_staff(self, peer_grades_me):
...
+        """
+        Scenario: complete workflow on a unit containing multiple ORA blocks.
+        """
+        self.set_vert_no(0)
+        # First Vertical - which is selected by default
+        learner = self.do_train_self_peer(peer_grades_me)
+
+        # Second Vertical
+        self.set_vert_no(1)
+        self.do_train_self_peer(peer_grades_me)
```
efischer19: @mushtaqak This test is only doing `train_self_peer`, but the problem you've added has a staff step enabled as well. It seems odd to only test a problem partway through.

Can you add functionality to this test to staff assess the user as well, and ensure:
- After the first assessment is complete, a grade appears for it but not for the second assessment
- After the second assessment is complete, grades appear for both
This will be similar to our full workflow tests: https://github.com/edx/edx-ora2/blob/master/test/acceptance/tests.py#L1078-L1102

efischer19:
Done with my first pass. The functionality being added looks good, I tried it out as both a staff user and a normal learner on your sandbox and everything worked as expected.

I am a bit concerned about the acceptance test not fully completing the problem though - we ought to complete the problem we're testing, either by adding a staff assessment or changing the test problem to have fewer steps.

mushtaqak:
@clytwynec Thanks :)

@cahrens Disabled the rules for my accessibility test. The test passed. Please have a look. Thanks.

test/acceptance/tests.py
mushtaqak: Ok - I will have a go tomorrow :) thanks for a quick review.

test/acceptance/accessibility.py
```diff
@@ -243,6 +243,78 @@ def test_multiple_rubrics(self):
...
+        self.staff_area_page = StaffAreaPage(self.browser, self.problem_loc)
+
+    # TODO: remove this method override. See TNL-1593
+    def _check_a11y(self, page):
+        page.a11y_audit.config.set_scope(
+            exclude=[
+                ".container-footer",
+                ".nav-skip",
+                "#global-navigation",
+                "#openassessment",
```
cahrens: We don't want to exclude all of these things. Please only exclude duplicate-id, and only for your new test. We want to minimize the number of things being ignored so that new issues are not introduced.

test/acceptance/accessibility.py
```diff
@@ -243,6 +243,78 @@ def test_multiple_rubrics(self):
...
+        Test accessibility when we have a unit containing multiple ORA blocks.
+        """
+        self.set_vert_no(0)
+        # First Vertical - which is selected by default
+        learner = self.do_train_self_peer()
+        self.check_a11y()
+
+        # Second Vertical
+        self.set_vert_no(1)
+        self.do_train_self_peer()
```
efischer19: The comment I had for acceptance tests applies here too - ensure that the problem is fully completed before checking a11y.

We probably ought to check it before starting as well, to ensure that the learner's "submit a response" view is a11y compliant.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -32,10 +32,13 @@ if (typeof OpenAssessment.unsavedChanges === 'undefined' || !OpenAssessment.unsa
     OpenAssessment.unsavedChanges = {};
 }
 
+var usageID;
```
cahrens: Please define usageID above in the constructor, instead of defining it outside of OpenAssessment.BaseView.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -32,10 +32,13 @@ if (typeof OpenAssessment.unsavedChanges === 'undefined' || !OpenAssessment.unsa
...
+
 // This is used by unit tests to reset state.
 OpenAssessment.clearUnsavedChanges = function() {
     OpenAssessment.unsavedChanges = {};
     window.onbeforeunload = null;
+    usageID = '';
```
cahrens: I don't think this should be needed.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -73,6 +76,25 @@ OpenAssessment.BaseView.prototype = {
...
+            usageID = $(event.target).closest('.xblock-student_view-openassessment').data('usage-id');
+        });
+    },
+
+    /**
+     * Get usage key of an XBlock.
+     */
+    getUsageID: function() {
+        if (!usageID) {
+            usageID = $(this.element).data('usage-id');
```
cahrens: You can likely assign this directly in the constructor (line 13).

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -73,6 +76,25 @@ OpenAssessment.BaseView.prototype = {
     },
 
     /**
+     * Install click handlers to set usage key of an XBlock.
+     */
+    setUpUsageID: function() {
+        $(".action--submit").on("click", function(event) {
```
cahrens: Can you also just do this in the constructor?

openassessment/xblock/static/js/src/lms/oa_peer.js
```diff
@@ -234,5 +237,16 @@ OpenAssessment.PeerView.prototype = {
...
+    },
+
+    /**
+    Get uuid of a peer assessment.
+
+    Args:
+        view: peer assessment view object.
+    **/
+    getUUID: function() {
+        var XBlockElement = $("div[data-usage-id='" + this.baseView.getUsageID() + "']");
```
cahrens: A variable name should not start with an upper-case letter.

openassessment/xblock/static/js/src/lms/oa_peer.js
```diff
@@ -234,5 +237,16 @@ OpenAssessment.PeerView.prototype = {
             view.baseView.toggleActionError('peer', errMsg);
             view.peerSubmitEnabled(true);
         });
+    },
+
+    /**
+    Get uuid of a peer assessment.
+
+    Args:
```
cahrens: There are no arguments to this method.

scripts/test-acceptance.sh
```diff
@@ -19,6 +19,7 @@
 cd `dirname $BASH_SOURCE` && cd ..
 
 # Note: support BASE_URL as a synonym for ORA_SANDBOX_URL for backward compatibility
+ORA_SANDBOX_URL = 'staff:edx@mushiora.sandbox.edx.org -o StrictHostKeyChecking=no'
```
cahrens: This should not be checked in.

test/acceptance/accessibility.py
```diff
@@ -243,6 +243,78 @@ def test_multiple_rubrics(self):
...
 
 
+class MultipleOpenAssessmentA11yTest(OpenAssessmentA11yTest, FullWorkflowMixin):
+    """
+    Test accessibility when we have a unit containing multiple ORA blocks.
+    """
+
+    def setUp(self):
+        super(MultipleOpenAssessmentA11yTest, self).setUp('multiple_ora', staff=True)
+        self.staff_area_page = StaffAreaPage(self.browser, self.problem_loc)
```
cahrens: This is interesting. self.staff_area_page, and all the other page objects, will only be for 1 of the 2 problems. That's fine in this case because the accessibility check actually verifies the entire page, not just the HTML within the page object whose _check_a11y  method is being called.

test/acceptance/accessibility.py
```diff
@@ -243,6 +243,78 @@ def test_multiple_rubrics(self):
...
+                "link-href",  # TODO: AC-199
+                "link-name",  # TODO: AC-196
+                "skip-link",  # TODO: AC-179
+            ]
+        })
+        page.a11y_audit.check_for_accessibility_errors()
+
+    def check_a11y(self):
+        """ check accessibility"""
+        self._check_a11y(self.submission_page)
```
cahrens: You don't actually need to call _check_a11y on all these different things. One call suffices, because the check does the whole HTML page at once. So this test point can be boiled down to just (note that it is important to actually visit the page you are testing):

````
self.submission_page.visit().submit_response(self.SUBMISSION)
self._check_a11y(self.submission_page)
````

A better test though would be going through the workflow and doing the accessibility check at different points, as different parts of the UI appear (take a look at the existing accessibility tests as an example).

File: test/acceptance/accessibility.py [hunk 16]
```diff
@@ -243,6 +243,78 @@ def test_multiple_rubrics(self):
...
+        self._check_a11y(self.self_asmnt_page)
+        self._check_a11y(self.peer_asmnt_page)
+        self._check_a11y(self.student_training_page)
+        self._check_a11y(self.staff_asmnt_page)
+        self._check_a11y(self.grade_page)
+        self._check_a11y(self.staff_area_page)
+
+    def set_vert_no(self, vert_no):
+        """ Set vertical number """
+        self.submission_page.vert_no = vert_no
```
cahrens: Don't use abbreviations like this. Either "vertical" or "vertical_number".

test/acceptance/tests.py
```diff
@@ -1102,6 +1105,43 @@ def test_train_self_peer_staff(self, peer_grades_me):
...
+class MultipleOpenAssessmentTest(OpenAssessmentTest, FullWorkflowMixin):
+    """
+    Test the multiple peer-assessment flow.
+    """
+
+    def setUp(self):
+        super(MultipleOpenAssessmentTest, self).setUp('multiple_ora')
+        self.staff_area_page = StaffAreaPage(self.browser, self.problem_loc)
+
+    def set_vert_no(self, vert_no):
```
cahrens: Don't duplicate this method in the a11y test file and here. Create a mixin that can be shared.

test/acceptance/pages.py
```diff
@@ -19,6 +19,7 @@ class OpenAssessmentPage(PageObject):
     """
     Base class for ORA page objects.
     """
+    vert_no = 0
```
cahrens: Document what this is (and as mentioned previously, don't use abbreviations).

test/acceptance/pages.py
```diff
@@ -38,7 +39,11 @@ def _bounded_selector(self, selector):
 
         The default implementation just returns the selector
         """
-        return selector
+        return "{vert_no_class} {selector}".format(vert_no_class=self.vert_no_class, selector=selector)
+
+    @property
+    def vert_no_class(self):
```
cahrens: A better name would be vertical_index_class

efischer19:
@mushtaqak I just merged my PR updating the test course into master; you'll need to rebase and re-apply your changes.

openassessment/xblock/static/js/src/lms/oa_base.js
mushtaqak: Actually, I could not move this to constructor because here we need the DOM to be loaded so as to apply click handler. That happens after the view has rendered, so I would need to call this in the views inheriting this.

openassessment/xblock/static/js/src/lms/oa_base.js
mushtaqak: Normally, I would not have need this, I could have got the usage id directly from the peer/self-assessment view as  `this.baseview.usageID` but sometimes (user has submitted some steps and refreshes or visits after browsing to the ORA problem) the usageID would not be set. So I had created this utility methods `getUsageId` that would always give us correct usage ID.

scripts/test-acceptance.sh
mushtaqak: Yes, you are right. I was testing this on the jenkins run. I would revert this file before merging :)

scripts/test-acceptance.sh
```diff
@@ -32,6 +32,7 @@ mkdir -p test/logs
 cd test/logs
 
 test_name="${1:-acceptance}"
+test=":MultipleOpenAssessmentTest"
```
mushtaqak: After getting thumbs-ups, I will un-check this file and squash my commits. This just for a fast jenkins run.

mushtaqak:
@efischer19 I fetched the latest course and then I have added my new unit to the `test-course.tar.gz`. I have have now checked for the `staff step` as well. Please have a look.

@cahrens I have created the Mixin that is shared between `tests.py` and `accessibility` tests. I have addressed your comments on javascript too. Please have a look.

FYI : Here is the [successful jenkins](http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/848/console) run for my acceptance tests.

mushtaqak:
@efischer19 , @cahrens ping

efischer19:
Hey @mushtaqak - we had a holiday yesterday, but I'll be getting to this shortly.

@cahrens is out all week, so I'll grab someone else here to be a second reviewer as well.

efischer19:
@andy-armstrong to be a second reviewer. Thanks Andy!

test/acceptance/pages.py
```diff
@@ -47,19 +57,21 @@ def url(self):
             loc=self._problem_location
         )
 
+
```
efischer19: Nit: extraneous newline

test/acceptance/tests.py
```diff
@@ -1201,6 +1263,30 @@ def test_feedback_only(self):
...
+    Test the multiple peer-assessment flow.
+    """
+
+    def setUp(self):
+        super(MultipleOpenAssessmentTest, self).setUp('multiple_ora')
+        self.staff_area_page = StaffAreaPage(self.browser, self.problem_loc)
+
+    @retry()
+    @attr('acceptance')
+    @ddt.data(True, False)
```
efischer19: While I appreciate the thoroughness here, I'm not sure we need to test both values of `peer_grades_me` in this test. `FullWorkflowRequiredTest:test_train_self_peer_staff` covers that behavior already, so doing it here as well adds an additional (long-running) test without really adding any additional coverage.

I'd recommend removing the `ddt.data`, and just use `assess_component` with `peer_grades_me` defaulting to `True`.

test/acceptance/tests.py
```diff
@@ -1201,6 +1263,30 @@ def test_feedback_only(self):
...
 
+@ddt.ddt
+class MultipleOpenAssessmentTest(OpenAssessmentTest, MultipeOpenAssessmentMixin):
+    """
+    Test the multiple peer-assessment flow.
+    """
+
+    def setUp(self):
+        super(MultipleOpenAssessmentTest, self).setUp('multiple_ora')
+        self.staff_area_page = StaffAreaPage(self.browser, self.problem_loc)
```
efischer19: Echoing @cahrens' comment here - how are the different staff area pages being differentiated? I'm going to run this locally and see what's happening

mushtaqak: @efischer19 Thanks for taking a look at this. And @andy-armstrong too :)

They are being differentiated using their vertical-index number. 

As I have mentioned the comment, each problem has a vertical-index number, so initial problem has `vert-0` class and second one has `vert-1` and so on. So when I move on to next problem, I reset the the vert-index number of the pages that test is using.

test/acceptance/tests.py
mushtaqak: OK - I will remove that. And that will also help faster run of acceptance tests :P

test/acceptance/pages.py
mushtaqak: Nice catch, I thought I removed extra lines in the last commits :P

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -73,6 +74,25 @@ OpenAssessment.BaseView.prototype = {
...
+            this.usageID = $(event.target).closest('.xblock-student_view-openassessment').data('usage-id');
+        });
+    },
+
+    /**
+     * Get usage key of an XBlock.
+     */
+    getUsageID: function() {
+        if (!this.usageID) {
+            this.usageID = $(this.element).data('usage-id');
```
andy-armstrong: I'm confused by the two different ways to set ``this.usageID``. Why do you need the click handler if this method can just determine the correct usage ID on demand?

efischer19:
@mushtaqak We're having some confusion looking at the TNL tickets for this bugfix. Here's what @andy-armstrong and I understand to be happening, can you confirm this, or correct anything that's wrong?

- https://openedx.atlassian.net/browse/TNL-1593 `Unable to submit assessments when multiple ORA problems in one unit`. This is saying that we need to properly fix all usages of id's in ORA for both accessibility and functionality reasons. The functionality is working as of this PR, but the a11y remains broken (as per your test comments)
- https://openedx.atlassian.net/browse/TNL-4155 `Specific Peer Assessment problem throws error message`. Someone made a unit with multiple ORA problems on the page, and we need to get that functionality working. This will be addressed with the landing of this PR.

If that's all correct, then we ought to make sure that 1593 gets properly updated and rescoped after this work lands.

test/acceptance/tests.py
efischer19: Ah okay, I understand now.

Would you mind adding a comment to this line in the constructor? Something like "set up staff area page for both problems, they will be differentiated by vertical index numbers".

andy-armstrong:
@mushtaqak This looks good to me. I'd like to understand the two different ways of determining the usage ID, and if they are both necessary have comments added that explain why. 👍  on the JavaScript once that's addressed.

efischer19:
That's all I've seen in this latest pass - it's getting close! Just a few small changes, then we can begin the squash-and-merge process.

openassessment/xblock/static/js/src/lms/oa_base.js
mushtaqak: Oh, nice question. You are right. I had to go in to deep to answer your question. 

I think, we can just go with `getUsageID` method. It is cleaner and better approach. Whenever we need `usageId` in any ORA step (self, response, peer, staff grade etc.), we can use `getUsageID` method to het the usageID on demand as I have done for peer step.

Initially, I was going with click handler option, then I was writing jasmine tests and I had to create `getUsageID` method. But your question made me to realise, do we need both. We can go with `getUsageID`.

mushtaqak:
@efischer19 You are right about both TNLs.  We would need to properly fix all usages of id's in ORA for both accessibility and functionality reasons in [TNL-1593](https://openedx.atlassian.net/browse/TNL-1593). My accessibility test has generated the possible places, where we can start of.

mushtaqak:
@efischer19 , @andy-armstrong Final review.

Here is the [latest jenkins run](http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/852/console) of all acceptance tests on my branch.

I will squash commits and merge after I get thumbs-ups.

test/acceptance/tests.py
```diff
@@ -910,6 +913,65 @@ def verify_grade_entries(self, expected_entries):
...
+        self.peer_asmnt_page.vertical_index = vertical_index
+        self.student_training_page.vertical_index = vertical_index
+        self.staff_asmnt_page.vertical_index = vertical_index
+        self.grade_page.vertical_index = vertical_index
+        self.staff_area_page.vertical_index = vertical_index
+
+    def staff_assessment(self, peer_grades_me=True):
+        """ Do staff assessment workflow """
+
+        # Ensure grade is not present, since staff assessment has not been made
```
efischer19: Since this is using the base `FullWorkflowMixin` class, could you refactor this logic to be shared and not duplicated?

https://github.com/edx/edx-ora2/blob/mushtaq/tnl-4155-fix-multiple-ora-issue/test/acceptance/tests.py#L939-L966 and https://github.com/edx/edx-ora2/blob/mushtaq/tnl-4155-fix-multiple-ora-issue/test/acceptance/tests.py#L1147-L1174 seem to be identical.

efischer19:
One last refactoring request, then 👍 from me.

openassessment/xblock/static/js/src/lms/oa_base.js
andy-armstrong: This is much cleaner. Thanks!

test/acceptance/accessibility.py
```diff
@@ -3,7 +3,7 @@
 """
 import os
 import unittest
-from tests import OpenAssessmentTest, StaffAreaPage, FullWorkflowMixin
+from tests import OpenAssessmentTest, StaffAreaPage, FullWorkflowMixin, MultipeOpenAssessmentMixin
```
andy-armstrong: Typo: MultpeOpenAssessmentMixin

test/acceptance/tests.py
```diff
@@ -910,6 +913,65 @@ def verify_grade_entries(self, expected_entries):
             self.assertEqual(expected_entry[1], self.grade_page.grade_entry(1, index))
 
 
+class MultipeOpenAssessmentMixin(FullWorkflowMixin):
```
andy-armstrong: Typo: Multipe...

andy-armstrong:
There's one typo, and then 👍  from me. Nice clean up.

mushtaqak:
@efischer19 I have squashed my commits. 

What else is required to get this PR merged. 

I have started the new [jenkins run](http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/854/console) here. If this passes, Are we good to merge ?

efischer19:
🚢  Yup, let's get that last jenkins run completed for safety, then merge away.

mushtaqak:
The [jenkins run](http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/856/console) just passed. I will be be merging in few minutes.

PR #896, opened 2016-03-22
Fix oa_block release state status

mushtaqak:
While checking for release state of the the ORA block, we are not checking if the XBlock has actually went live or it has release/start date in future thus making it's state to scheduled/ready.

**Fix**
Add a check to calculate XBlock's start date against current time to define if it is live or not.

[TNL-4268](https://openedx.atlassian.net/browse/TNL-4268)

### DIFF
--- a/openassessment/xblock/openassessmentblock.py
+++ b/openassessment/xblock/openassessmentblock.py
@@ -35,7 +35,7 @@ from openassessment.xblock.staff_assessment_mixin import StaffAssessmentMixin
 from openassessment.workflow.errors import AssessmentWorkflowError
 from openassessment.xblock.student_training_mixin import StudentTrainingMixin
 from openassessment.xblock.validation import validator
-from openassessment.xblock.resolve_dates import resolve_dates, DISTANT_PAST, DISTANT_FUTURE
+from openassessment.xblock.resolve_dates import resolve_dates, parse_date_value, DISTANT_PAST, DISTANT_FUTURE
 from openassessment.xblock.data_conversion import create_prompts_list, create_rubric_dict, update_assessments_format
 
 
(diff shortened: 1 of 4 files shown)


mushtaqak:
@adampalay @Qubad786 @cahrens May you review this.

cahrens:
@efischer19 Can you review this PR please (in my place)?

efischer19:
Will do

openassessment/xblock/openassessmentblock.py
```diff
@@ -834,7 +834,11 @@ def is_released(self, step=None):
         else:
             is_published = True
         is_closed, reason, __, __ = self.is_closed(step=step)
-        return is_published and (not is_closed or reason == 'due')
+        is_released = is_published and (not is_closed or reason == 'due')
+        is_live = dt.datetime.now(pytz.UTC) > parse_date_value(self.start, self._) if self.start else None
```
efischer19: I like this, but I think it would read a bit cleaner if we're more explicit about checking `self.start`, and avoid having `is_live` be None.

Perhaps something like this:
```
if self.start:
    is_live = ...
    is_released = is_released and is_live
```

efischer19:
I've tested this locally and it seems to perform as advertised, good fix. :+1: from me, one suggestion that can be addressed or ignored as you see fit.

Qubad786:
@mushtaqak LGTM :+1: Nice fix!

mushtaqak:
@efischer19 Thanks, Nice suggestion. I have addressed your comment.

adampalay:
@mushtaqak , no merging things when you're on vacation! ;)

@efischer19 , is ora released regularly or should I make a PR to update edx-platform?

efischer19:
@mushtaqak beat you to it! 
https://github.com/edx/edx-platform/pull/11884

Thanks again!

adampalay:
dang :)

PR #898, opened 2016-04-05
Handle IntegrityError according to 1.8 conventions.

dianakhuang:
@cahrens @efischer19 

This is a small fix that does not have any tests, but the acceptance tests do catch this error. I've verified them locally by running the acceptance tests against a sandbox that uses this branch.

Here's the relevant Django documentation page that explains why this fixes this: https://docs.djangoproject.com/en/1.8/topics/db/transactions/#controlling-transactions-explicitly

TNL-4342

### DIFF
--- a/openassessment/assessment/serializers/training.py
+++ b/openassessment/assessment/serializers/training.py
@@ -181,9 +181,10 @@ def deserialize_training_examples(examples, rubric_dict):
                 example = TrainingExample.objects.get(content_hash=content_hash)
             except TrainingExample.DoesNotExist:
                 try:
-                    example = TrainingExample.create_example(
-                        example_dict['answer'], example_dict['options_selected'], rubric
-                    )
+                    with transaction.atomic():
+                        example = TrainingExample.create_example(
+                            example_dict['answer'], example_dict['options_selected'], rubric
+                        )
                 except IntegrityError:
                     example = TrainingExample.objects.get(content_hash=content_hash)
 
--- a/openassessment/assessment/api/peer.py
+++ b/openassessment/assessment/api/peer.py
@@ -115,14 +115,15 @@ def on_start(submission_uuid):
 
     """
     try:
-        submission = sub_api.get_submission_and_student(submission_uuid)
-        workflow, __ = PeerWorkflow.objects.get_or_create(
-            student_id=submission['student_item']['student_id'],
-            course_id=submission['student_item']['course_id'],
-            item_id=submission['student_item']['item_id'],
-            submission_uuid=submission_uuid
-        )
-        workflow.save()
+        with transaction.atomic():
+            submission = sub_api.get_submission_and_student(submission_uuid)
+            workflow, __ = PeerWorkflow.objects.get_or_create(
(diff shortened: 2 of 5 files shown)


efischer19:
:+1: Good catch! I'm OK with this not having test coverage - I'd consider it a problem related to the one-time work in the 1.8 upgrade

Should we grep for other potential cases of this popping up, and wrap all of them in `with transaction.atomic():` for safety?

dianakhuang:
There are a couple of other places where we aren't handling `IntegrityError`s correctly. I'll wrap them all with these as well. Just because the behavior isn't popping up right now doesn't mean it won't pop up later.

dianakhuang:
Added some more protective `transaction.atomic`s, @efischer19 . It's probably worth it to give the PR another look.

efischer19:
:+1: right on. I figure this is a "no penalty, potentially protect against oddness in the future" scenario.

cahrens:
This looks reasonable to me, but can @macdiesel also take a quick look (Django 1.8/platform expert)? When I search for "except IntegrityError" in edx-platform, I see many places where we don't wrap the operation in transaction.atomic. I'd like to better understand when this is necessary.

mattdrayer:
I am so happy to see this is looking to be a straightforward fix after the convoluted process of determining the actual issue -- nice job!

dianakhuang:
Thanks so much for doing that digging @mattdrayer ! Your help was invaluable in figuring out what the correct fix is/was.

openassessment/assessment/serializers/training.py
```diff
@@ -181,9 +181,10 @@ def deserialize_training_examples(examples, rubric_dict):
                 example = TrainingExample.objects.get(content_hash=content_hash)
             except TrainingExample.DoesNotExist:
                 try:
-                    example = TrainingExample.create_example(
-                        example_dict['answer'], example_dict['options_selected'], rubric
-                    )
+                    with transaction.atomic():
+                        example = TrainingExample.create_example(
```
macdiesel: It would be preferable to more closely follow the django model api and override get_or_create and create method to push some of this logic down into the model. We are doing things here that the serializer shouldn't need to worry about.

macdiesel:
Overall I think that the usage of transaction.atomic is fine. There are some instances in here where we get an object, and if it doesn't exist create it but if it can't be created we run a get again. I commented on the one occurrence of this I saw. In these cases I would prefer to see this type of logic moved to the data models. 

Question: Do we understand why we have these race conditions? If we do not have a firm handle on what is causing these and when I would suggest adding some logging statements to the excepts where Integrity error is caught. as this could help us more accurately pinpoint where and when this is happening.

dianakhuang:
Thanks @macdiesel ! Unfortunately, we inherited this code base, and I'm not sure we understand most of the decisions made. I'm hesitant to do much more to the code other than protect the places where the IntegrityErrors are happening, just because we want to make sure that this bug doesn't block the next release.

mushtaqak:
Sweet :)

Thank you @mattdrayer and @dianakhuang for fixing this :)

PR #903, opened 2016-04-11
Feedback only cleanup

efischer19:
Adding acceptance tests and other minor cleanup in the aftermath of TNL-4351.


efischer19:
remaining issues:
- guaranteeing a peer grade to exist is causing all sorts of problems, to the point I'm thinking I may just change the problem definition to only include self and staff grades, or even just staff.
- staff tools feedback needs to be checked still

efischer19:
Latest commit (https://github.com/edx/edx-ora2/pull/903/commits/b65d67ada289a0f27a2f1a018c8559ea79a6321e) fixes a few more visual issues with no-options criterion.

Before screengrabs:
screen shot 2016-04-13 at 11 36 53 am
screen shot 2016-04-13 at 11 47 27 am

After screengrabs:
screen shot 2016-04-13 at 11 37 01 am
screen shot 2016-04-13 at 11 47 35 am

I'll be adding "ensure these fields aren't visible" checks to the new acceptance test as well.

efischer19:
Screenshot of the new problem I added, in case that is helpful.

screen shot 2016-04-13 at 3 05 09 pm

efischer19:
@dianakhuang @cahrens This PR is finally ready for review. I've modified our templates in a few places to clean up handling of optionless criterion (see screenshots above), and added a new acceptance test.

If you run this test yourself, point `ORA_SANDBOX_URL` at my sandbox, as I've already added the new problem to test there. (link: https://efischer19.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/8d9584d242b44343bc270ea5ef04ab03/a2875e0db1454d0b94728b9a7b28000b/)

Also @mushtaqak - I know we're both working on PRs that will add problems to the test course. We'll need to be careful to ensure that whichever PR lands last is rebased and the new problem re-added again, so that both additional problems make it into master.

test/acceptance/tests.py
```diff
@@ -232,13 +235,17 @@ def do_staff_override(self, username, final_score=STAFF_AREA_SCORE.format(STAFF_
...
 
-    def do_staff_assessment(self, number_to_assess=0, options_selected=OPTIONS_SELECTED):
+    def do_staff_assessment(self, number_to_assess=0, options_selected=OPTIONS_SELECTED, feedback=None):
         """
         Use staff tools to assess available responses.
 
         Args:
             number_to_assess: the number of submissions to assess. If not provided (or 0),
                 will grade all available submissions.
+            options_to_select (dict): the options to choose when grading. Using this parameter precludes leaving
```
cahrens: Wrong variable name.

test/acceptance/tests.py
```diff
@@ -1101,6 +1112,94 @@ def test_train_self_peer_staff(self, peer_grades_me):
                 [(u"YOUR SELF ASSESSMENT", u"Good"), (u"YOUR SELF ASSESSMENT", u"Excellent")],
             ])
 
+@ddt.ddt
+class FeedbackOnlyTest(OpenAssessmentTest, FullWorkflowMixin):
+    """
+    Test for a problem that only accepts feedback. Will make and verify peer, self, and staff assessments.
```
cahrens: The problem does have some criterion with options though, right?

test/acceptance/tests.py
```diff
@@ -75,6 +75,9 @@ class OpenAssessmentTest(WebAppTest):
         'full_workflow_staff_required':
             u'courses/{test_course_id}/courseware/'
             u'8d9584d242b44343bc270ea5ef04ab03/0b0dcc728abe45138c650732af178afb/'.format(test_course_id=TEST_COURSE_ID),
+        'feedback_only':
```
cahrens: I get concerned about calling this "feedback_only" because it is a hybrid problem now.

test/acceptance/tests.py
```diff
@@ -1101,6 +1112,94 @@ def test_train_self_peer_staff(self, peer_grades_me):
...
+        # Make self assessment
+        self.self_asmnt_page.visit()
+        self.self_asmnt_page.wait_for_page()
+        self.self_asmnt_page.provide_criterion_feedback(self.generate_feedback("self", "criterion"))
+        self.self_asmnt_page.provide_overall_feedback(self.generate_feedback("self", "overall"))
+        self.self_asmnt_page.assess("self", [0])
+        self.self_asmnt_page.wait_for_complete()
+        self.assertTrue(self.self_asmnt_page.is_complete)
+
+        # Staff assess all available submissions
```
cahrens: Update comment.

test/acceptance/tests.py
```diff
@@ -1101,6 +1112,94 @@ def test_train_self_peer_staff(self, peer_grades_me):
...
+        self.do_staff_assessment(
+            options_selected = [0],
```
cahrens: Why [0]?

test/acceptance/tests.py
```diff
@@ -1101,6 +1112,94 @@ def test_train_self_peer_staff(self, peer_grades_me):
...
+            # Criterion feedback first
+            expected = self.generate_feedback(assessment_type, "criterion")
+            actual = self.grade_page.feedback_entry(1, i)
+            self.assertEqual(actual, expected)
+            # Then overall
+            expected = self.generate_feedback(assessment_type, "overall")
+            actual = self.grade_page.feedback_entry("feedback", i)
+            self.assertEqual(actual, expected)
+
+        # Verify that all score-bearing items are hidden
```
cahrens: This is a bit weak because we don't know that the correct ones are showing scores. We just know that some things are hidden. Is there a way to verify the individual expected parts instead of just counts?

test/acceptance/tests.py
```diff
@@ -1101,6 +1112,94 @@ def test_train_self_peer_staff(self, peer_grades_me):
...
+        )
+        self.assertEqual(
+            self.staff_area_page.status_text('self__assessments')[5],
+            self.generate_feedback("self", "criterion")
+        )
+        self.assertEqual(
+            self.staff_area_page.overall_feedback('self__assessments'),
+            self.generate_feedback("self", "overall")
+        )
+        # No score should be shown
```
cahrens: This comment doesn't seem to make sense.

test/acceptance/tests.py
efischer19: As discussed - I'll add 2 additional assertions above for the 1-point option from each of staff and self, then make the comment say something along the lines of "Since we've verified the 6 reported answers that should exist, ensure there are no more"

test/acceptance/tests.py
efischer19: `options_selected` takes a list of indices, such that the `i-th` element represents the option to choose for criterion `i`. Since we only have one scored criterion, and we want to choose the first (zero-indexed) option for that criterion, the list becomes `[0]`. I'll add a comment to this effect.

test/acceptance/tests.py
efischer19: Correct, I need to update this.

test/acceptance/tests.py
efischer19: In what way? By not specifying `number_to_select`, we will be staff assessing all available submissions.

https://github.com/edx/edx-ora2/blob/efischer/no_options_cleanup/test/acceptance/tests.py#L243-L244

test/acceptance/tests.py
efischer19: Yeah, but by the same token, the primary purpose of this test is to check a criterion with only feedback (no scoring options). @dianakhuang, can you think of a better name?

I also call it the "Feedback Only Subsection" in the LMS, I'll need to change that as well.

openassessment/xblock/grade_mixin.py
```diff
@@ -361,7 +361,7 @@ def _get_assessment_part(title, feedback_title, part_criterion_name, assessment)
         )
         self_assessment_part = _get_assessment_part(
             _('Self Assessment Grade') if is_staff else _('Your Self Assessment'),
-            _('Self Comments'),
+            _('Your Comments'),
```
efischer19: This is in order to match the heading on overall feedback, I had it wrong initially

screen shot 2016-04-14 at 10 26 37 am

efischer19:
@cahrens I've addressed your comments, this is ready for another round of review.

mushtaqak:
@efischer19 I know that we would need to update the `test-course.tar.gz`. But thanks for the heads up.

First we would need to fetch latest copy, then add our new problem and push again. Seems like it's going to be me :P

openassessment/xblock/grade_mixin.py
```diff
@@ -350,10 +357,11 @@ def _get_assessment_part(title, part_criterion_name, assessment):
         else:
             peer_assessment_part = None
         example_based_assessment_part = _get_assessment_part(
-            _('Example-Based Grade'), criterion_name, example_based_assessment
+            _('Example-Based Grade'), _('Example-Based Comments'), criterion_name, example_based_assessment
         )
         self_assessment_part = _get_assessment_part(
             _('Self Assessment Grade') if is_staff else _('Your Self Assessment'),
+            _('Your Comments'),
```
cahrens: Nit: can you please add a comment stating that this new value is only used on the LMS view? I was worried because the line above made me think that we needed a different value when is_staff is True, but we don't because this new field is not used in the learner info part of the staff tools.

test/acceptance/tests.py
```diff
@@ -232,13 +235,16 @@ def do_staff_override(self, username, final_score=STAFF_AREA_SCORE.format(STAFF_
...
 
-    def do_staff_assessment(self, number_to_assess=0, options_selected=OPTIONS_SELECTED):
+    def do_staff_assessment(self, number_to_assess=0, options_selected=OPTIONS_SELECTED, feedback=None):
         """
         Use staff tools to assess available responses.
 
         Args:
             number_to_assess: the number of submissions to assess. If not provided (or 0),
                 will grade all available submissions.
+            selected_options (dict): the options to choose when grading. Defaults to OPTIONS_SELECTED.
```
cahrens: The name of the variable is options_selected.

test/acceptance/tests.py
cahrens: OK, I misunderstood.

test/acceptance/tests.py
cahrens: Please do add a comment.

cahrens:
A couple of minor doc/comments need to be updated, otherwise 👍

test/acceptance/pages.py
```diff
@@ -451,6 +451,11 @@ def grade_entry(self, question, column):
         Returns: the tuple of source and value information for the requested grade
 
         """
+        self.wait_for_element_visibility(
+            self._bounded_selector('.question--{} .answer .answer__source__value'.format(question + 1)),
+            "Grade entry was present",
+            2
+        )
```
efischer19: @cahrens I wanted to call out this particular change - I was getting intermittent failures on my acceptance test due to accessing this value before it was present, this `wait_for_element_visibility` resolves that.

Everything else in this commit is just addressing your comments from before.

cahrens:
@dianakhuang Please go ahead and review this PR. I'm 👍

dianakhuang:
Looking good. :rocket:

PR #908, opened 2016-06-03
Update for xblock static file handling.

cahrens:
TNL-4710

@tobz and @efischer19 please review. After this is merged, I will bump the ORA release number and do a new version.

I verified on my devstack that this works. There's no point in creating a sandbox for this because sandboxes worked before.

### DIFF
--- a/scripts/run-pep8.sh
+++ b/scripts/run-pep8.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-MAX_PEP8_VIOLATIONS=113
+MAX_PEP8_VIOLATIONS=110
 
 mkdir -p test/logs
 PEP8_VIOLATIONS=test/logs/pep8.txt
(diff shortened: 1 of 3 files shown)


scripts/run-pep8.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-MAX_PEP8_VIOLATIONS=113
+MAX_PEP8_VIOLATIONS=110
```
cahrens: While running tests, I noticed that we only had 110 pep8 violations, so I lowered this.

tobz:
👍

efischer19:
LGTM, confirmed working on my devstack as well.

👍

efischer19:
Let me know when you need me to run translations for the ORA release

cahrens:
Thanks @efischer19, can you go ahead and run translations? I will add to it the version bump.

PR #910, opened 2016-06-09
Update transifex with new location

efischer19:
@nedbat @cahrens I believe this is the only change we need to make to ora2 code to get the new translation location working properly.

I'll test and confirm this works with Natalia.


cahrens:
Thanks for jumping on this, @efischer19!

PR #911, opened 2016-06-13
Set ORA display name by default

ssemenova:
### Description
 
[TNL-4687](https://openedx.atlassian.net/browse/TNL-4687)

ORA display name should be set by default, instead of empty. I also added a unit test, and translated the display name in the template. I didn't translate the display name in openassessmentblock.py because it was strangely not being translated, and the quality test failed because the gettext import was unused.

### Sandbox
- [ ] https://studio-ssemenova.sandbox.edx.org/home/

### Testing
- [ ] i18n
- [ ] RTL
- [ ] safecommit shows 0 violations
- [ ] Analytics
- [ ] Performance

### Reviewers
- [ ] Code review: @cahrens 
- [ ] Code review: @clrux 

FYI: @nedbat - could you look at the way the translation is done?

### Post-review
- [ ] Squash commits

### DIFF
--- a/openassessment/templates/openassessmentblock/oa_base.html
+++ b/openassessment/templates/openassessmentblock/oa_base.html
@@ -3,7 +3,7 @@
 <div class="wrapper wrapper--xblock wrapper--openassessment theme--basic">
     <div class="openassessment problem" id="openassessment">
         <div class="wrapper--grid">
-            <h2 class="openassessment__title problem__header">{{ title }}</h2>
+            <h2 class="openassessment__title problem__header">{% trans title %}</h2>
 
             <div class="wrapper-openassessment__message">
                 {% block message %}
--- a/openassessment/xblock/openassessmentblock.py
+++ b/openassessment/xblock/openassessmentblock.py
@@ -151,7 +151,7 @@ class OpenAssessmentBlock(
     )
 
     title = String(
-        default="",
+        default="Open Response Assessment",
         scope=Scope.content,
         help="A title to display to a student (plain text)."
     )
--- a/openassessment/xblock/test/test_studio.py
+++ b/openassessment/xblock/test/test_studio.py
@@ -111,6 +111,11 @@ class StudioViewTest(XBlockHandlerTestCase):
     }
 
     @scenario('data/basic_scenario.xml')
+    def test_default_fields(self, xblock):
+        # Default value should not be empty
+        self.assertEqual(xblock.fields['title'].default, "Open Response Assessment")
+
+    @scenario('data/basic_scenario.xml')
     def test_render_studio_view(self, xblock):
         frag = self.runtime.render(xblock, 'studio_view')
         self.assertTrue(frag.body_html().find('openassessment-edit'))


openassessment/xblock/test/test_studio.py
```diff
@@ -111,6 +111,13 @@ class StudioViewTest(XBlockHandlerTestCase):
     }
 
     @scenario('data/basic_scenario.xml')
+    def test_default_fields(self, xblock):
+        # Reset all fields in the XBlock to their default values
+        for field_name, field in xblock.fields.iteritems():
```
cahrens: You might as well only set title to its default value.

openassessment/templates/openassessmentblock/oa_base.html
```diff
@@ -3,7 +3,7 @@
 <div class="wrapper wrapper--xblock wrapper--openassessment theme--basic">
     <div class="openassessment problem" id="openassessment">
         <div class="wrapper--grid">
-            <h2 class="openassessment__title problem__header">{{ title }}</h2>
+            <h2 class="openassessment__title problem__header">{% trans title %}</h2>
```
cahrens: So there is no separate Studio file?

ssemenova: Nope, it's kind of confusingly laid out, but I checked, and changing the title here changes it in both the lms view and the studio view, but not the studio edit.

openassessment/xblock/openassessmentblock.py
```diff
@@ -151,7 +151,7 @@ class OpenAssessmentBlock(
     )
 
     title = String(
```
clrux: This is translated?

clrux:
Aside from my one question this seems fairly straightforward. 👍

openassessment/xblock/openassessmentblock.py
ssemenova: In the template, not in the variable. For some reason it wasn't working if I translated it there, and the tests were saying the import was unused. ¯\_(ツ)_/¯

ssemenova:
fyi: @cahrens sandbox is up

cahrens:
:+

openassessment/templates/openassessmentblock/oa_base.html
nedbat: I don't see a reason to use `{% trans %}` here, since the title will be user-supplied text.  That means in general, we won't have translations for it.

cahrens: @nedbat We were trying translate the default title name. @ssemenova tried to translate the default title where it is declared in openassessmentblock.py, but she ran into issues. Perhaps you have some advice-- I know that xblock translation support is pretty spotty.

ssemenova: @nedbat Yes, for some reason when I added the translation into openassessmentblock.py, it never got added to the list of things to translate, and I was getting some quality errors about _ not being used.

nedbat: I see now.  I think you should then have `_("Open Response Assessment")` in the Python file, or the string may not be collected for translation at all.  You need to somehow ensure that the string is lexically marked as translatable (so that it will be collected), and also ensure that at run time the string is translated.  Often a simple `_()` function call is enough, but in your case it is trickier.

nedbat: I can help with the details (though I am at home today).

cahrens: Perhaps we are better off just not attempting to translate the default title. Since we expect course authors to replace this default title with something that will then not be translated, there is very little benefit in translating the default title.

PR #912, opened 2016-06-16
Update libsass version.

benpatterson:


benpatterson:
@efischer19 @andy-armstrong @cahrens (kind of guessing on the right folks).

We have python version coupling between platform and ora2, it turns out. I'm not aiming to loosen that coupling, but instead am looking to coordinate an upgrade between the two repositories. It'll result in smarter sass compilation and catch more compile errors on future code changes.

andy-armstrong:
👍

ormsbee:
👍 , but can't we also just unpin this version of libsass to something like >=0.11.0 or some such?

benpatterson:
The testing element of this occurred to me over the weekend, and I already put on the caveat that I was just upgrading, not thinking of decoupling... 

But I admit I'd rather fix the underlying requirements problem. So I'm going to leave this (and the accompanying platform PR) open for now, until I get a chance to make the requirements smarter, along the lines of what Dave is suggesting.

Will ping the thread when ready.

benpatterson:
Updating previous comments, obviously, I did not get back to this. Happy to help out folks that want to pick it up, but I likely won't have the cycles to see it all the way through. At least, not in the medium-term.

benpatterson:
FYI @andy-armstrong @ormsbee @arizzitano 

I'm closing this as well; it's a partner to the other change. You can see previous comments.

PR #915, opened 2016-07-06
Convert ORA Due Dates

kevinjkim:
### Description
 
[TNL-4900](https://openedx.atlassian.net/browse/TNL-4900)

Iniital commit. Converts ORA response, self-assessment, peer-assessment, and student training dates.

### Sandbox
[https://kevinjkim.sandbox.edx.org/](https://kevinjkim.sandbox.edx.org/)

### Testing
- [ ] i18n
- [ ] RTL
- [ ] safecommit shows 0 violations
- [ ] Analytics
- [ ] Performance

### Reviewers
- [x] Code review: @nasthagiri  
- [ ] Doc Review: @catong 
- [ ] UX review: @...
- [ ] Accessibility review: @...
- [ ] Product review: @...

### Post-review
- [x] Squash commits


jcdyer:
handing off review duties to @nasthagiri .

openassessment/xblock/resolve_dates.py
```diff
@@ -24,6 +24,18 @@ class DateValidationError(Exception):
 DISTANT_FUTURE = dt.datetime(dt.MAXYEAR, 1, 1, tzinfo=pytz.utc)
 
 
+def get_current_time_zone(open_assessment_block):
```
nasthagiri: It's usually better to pass in the granularity of the object that is actually needed by the function instead of a higher layer object.  In this case, it's not the ORA block that is really needed by this time_zone function.  Rather, it needs the runtime service in order to retrieve the user and their preferences.

openassessment/templates/openassessmentblock/response/oa_response.html
```diff
@@ -10,15 +10,16 @@ <h2 class="step__title">
             <span class="step__counter"></span>
             <span class="wrapper--copy">
                 <span class="step__label">{% trans "Your Response" %}</span>
+                {% get_current_timezone as TIME_ZONE %}
```
efischer19: I'm confused by this line.
- Should there be a `with` here? Usually I see python code that looks like `with <thing> as <other_thing>`
- Where is `TIME_ZONE` (all caps) being used?

kevinjkim: oh whoops. that line was supposed to be removed.

openassessment/xblock/resolve_dates.py
efischer19: 👍 Agreed. I'd even like to a see a property that gets the block's `user_service`, so the calls could then simply be `get_current_time_zone(self.user_service)`

openassessment/xblock/test/test_openassessment.py
```diff
@@ -91,95 +95,211 @@ def test_student_view_workflow_error(self, xblock):
             xblock_fragment = self.runtime.render(xblock, "student_view")
 
         # Expect that the page renders even if the update fails
-        self.assertTrue(xblock_fragment.body_html().find("Openassessmentblock"))
+        self.assertIn("OpenAssessmentBlock", xblock_fragment.body_html())
 
-    @scenario('data/dates_scenario.xml')
```
efischer19: Can you try deleting `data/dates_scenario.xml` and see if tests continue to pass? If we're no longer using this file, it should be removed.

efischer19:
Overall it looks good. Once @nasthagiri's comment from above is addressed, I think we'll be set.

Can we get a sandbox set up to inspect as well? I can help you get that once you've addressed review comments both here and in the edx-platform PR.

openassessment/xblock/resolve_dates.py
nasthagiri: Yeah.  Can we update this method to take in user_service as input instead?

openassessment/xblock/test/test_openassessment.py
kevinjkim: Oh, I actually still use that file on line 113. I just removed the decorator b/c it was acting janky with ddt.

openassessment/xblock/student_training_mixin.py
```diff
@@ -106,6 +109,7 @@ def training_path_and_context(self):
 
             if due_date < DISTANT_FUTURE:
                 context['training_due'] = due_date
+                context['time_zone'] = get_current_time_zone(self)
```
nasthagiri: Shouldn't you pass in `user_service` here instead of `self`?
Also, can you reduce the number of times this is done within this same method?  That is, set the time_zone in the context only once before the outer if statement.

openassessment/xblock/self_assessment_mixin.py
```diff
@@ -54,12 +54,14 @@ def self_path_and_context(self):
         context = {'allow_latex': self.allow_latex}
```
nasthagiri: Set the `time_zone` in the `context` at the time the `context` is initially created here - so it doesn't need to be added in each individual case within this method.  Update other cases in the PR where this happens.

openassessment/xblock/resolve_dates.py
```diff
@@ -24,6 +24,18 @@ class DateValidationError(Exception):
 DISTANT_FUTURE = dt.datetime(dt.MAXYEAR, 1, 1, tzinfo=pytz.utc)
 
 
+def get_current_time_zone(user_service):
+    """
+    Returns the current time zone for the user, if specified, or UTC if not
```
efischer19: "Returns the *preferred* time zone for the *current* user"

openassessment/xblock/self_assessment_mixin.py
efischer19: Agreed - there's not a huge penalty when including it in the few cases it isn't needed. Let's just do it once for simplicity.

openassessment/xblock/submission_mixin.py
```diff
@@ -376,6 +376,7 @@ def submission_path_and_context(self):
         """
         workflow = self.get_workflow_info()
         problem_closed, reason, start_date, due_date = self.is_closed('submission')
```
efischer19: (this applies to each of the files we've called out for time zone fetching)

I think a good way to treat this is "If and when dates are defined, I should fetch the time zone"

openassessment/xblock/test/test_openassessment.py
efischer19: Oh cool. Carry on then.

openassessment/xblock/test/test_openassessment.py
```diff
@@ -91,95 +96,246 @@ def test_student_view_workflow_error(self, xblock):
...
+            xblock_fragment = self.runtime.render(xblock, "student_view")
+            self.assertIn("OpenAssessmentBlock", xblock_fragment.body_html())
+
+            # Validate Submission Rendering.
+            submission_response = xblock.render_submission({})
+            self.assertIsNotNone(submission_response)
+            self.assertIn("openassessment__response", submission_response.body)
+            self.assertIn(expected_date, submission_response.body)
+
+    time_zone_fn_path = 'openassessment.xblock.peer_assessment_mixin.get_current_time_zone'
```
efischer19: It's weird to see this definition here. can you move it to the top of the class, and ALL_CAPS to indicate a constant?

openassessment/xblock/test/test_openassessment.py
```diff
@@ -91,95 +96,246 @@ def test_student_view_workflow_error(self, xblock):
...
+            resp = xblock.render_peer_assessment(request)
+
+            self.assertIn(expected_start_date, resp.body)
+
+    @ddt.data(('utc', 'May 1, 2014 00:00 UTC'),
+              ('America/Los_Angeles', 'April 30, 2014 17:00 PDT'))
+    @ddt.unpack
+    @patch.object(openassessmentblock.OpenAssessmentBlock, 'is_beta_tester', new_callable=PropertyMock)
+    def test_formatted_end_dates_for_beta_tester_without_days_early(self, time_zone, expected_end_date,
+                                                                    mock_is_beta_tester):
```
efischer19: Can you format this as follows?
```
function_name(
    param1,
    param2,
    ...
)
```

efischer19:
Done with my latest pass - looking great! The only big-ish issue remaining is setting `context[time_zone]' initially, my other comments are smaller doc/formatting suggestions.

Also, it looks like the minified js files are outdated on master, which is why they unexpectedly appear in this diff. I've filed https://github.com/edx/edx-ora2/pull/918 to address.

efischer19:
👍 LGTM once squashed/rebased, tested once more, etc.

openassessment/xblock/resolve_dates.py
```diff
@@ -46,7 +58,11 @@ def _parse_date(value, _):
         try:
             return parse_date(value).replace(tzinfo=pytz.utc)
```
efischer19: Just out of curiosity - should we be respecting the user's timezone here too, when parsing dates?

efischer19:
Sandbox looks great! I don't know why I never noticed this before - but it's really cool that future dates will use DST or not depending on if it's applicable at the date in question

screen shot 2016-07-15 at 3 28 30 pm

openassessment/xblock/resolve_dates.py
kevinjkim: I don't think it's necessary. That's only used to compare dates to each other, so i just keep all of those in UTC and convert to timezone only when displaying to user.

efischer19: Cool. I figured it was alright, just wanted to make sure nothing slipped through the cracks.

kevinjkim:
Ready for a second reviewer.

openassessment/xblock/test/test_openassessment.py
```diff
@@ -91,95 +98,252 @@ def test_student_view_workflow_error(self, xblock):
...
+            time_zone_fn.return_value = pytz.timezone(time_zone)
+
+            # Set start dates'
+            xblock = self.load_scenario('data/basic_scenario.xml')
+            xblock.start = dt.datetime(2014, 4, 1, 1, 1, 1)
+            request = namedtuple('Request', 'params')
+            request.params = {}
+            resp = xblock.render_peer_assessment(request)
+
+            self.assertIn(expected_start_date, resp.body)
```
nasthagiri: These test methods can be DRYed up by creating a few helper methods.  This way, it makes it better for code maintenance - rather than having duplicate code - even if (especially if) it's test code.

So, for example, creating helper methods for (1) setting start/due dates and (2) setting up `days_early_for_beta`.

nasthagiri:
Completed my pass.  Other than DRYing up the test methods, it looks good.

kevinjkim:
All comments addressed. Ready for another pass.

nasthagiri:
👍 looks good!

catong:
No doc concerns in this PR so consider my review complete. Thanks, @kevinjkim!

PR #926, opened 2016-08-22
Use proper platform details in user facing string

efischer19:
While reviewing translations, @ "Natalia Berdnikov" noticed that this string uses `edX`, which may not be correct for all openedx installations.

@nedbat @shaunagm Is there a standard way to refer to "the people running this instance" in code? @sstack22 and @catong may have suggestions as well.

### DIFF
--- a/openassessment/templates/openassessmentblock/oa_uploaded_file.html
+++ b/openassessment/templates/openassessmentblock/oa_uploaded_file.html
@@ -20,7 +20,7 @@
                 {% trans "View the file associated with this submission." %}
             </a>
             {% if show_warning %}
-                <p class="submission_file_warning">{% trans "(Caution: This file was uploaded by another course learner and has not been verified, screened, approved, reviewed, or endorsed by edX. If you decide to access it, you do so at your own risk.)" %}</p>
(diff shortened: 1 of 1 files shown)


openassessment/templates/openassessmentblock/oa_uploaded_file.html
```diff
@@ -20,7 +20,7 @@ <h3 class="{{ class_prefix }}__display__title">
                 {% trans "View the file associated with this submission." %}
             </a>
             {% if show_warning %}
-                <p class="submission_file_warning">{% trans "(Caution: This file was uploaded by another course learner and has not been verified, screened, approved, reviewed, or endorsed by edX. If you decide to access it, you do so at your own risk.)" %}</p>
+                <p class="submission_file_warning">{% trans "(Caution: This file was uploaded by another course learner and has not been verified, screened, approved, reviewed, or endorsed by the platform maintainer. If you decide to access it, you do so at your own risk.)" %}</p>
```
catong: Perhaps "administrator" rather than "maintainer"?

nedbat: I don't think typical users will know what a "platform" is.  "site administrator" seems like something people will recognize.

efischer19:
@catong @nedbat I went with `site administrator`. Can I get thumbs to merge?

catong:
👍 Thanks, @efischer19!

PR #928, opened 2016-08-25
Transifex updates

efischer19:


efischer19:
Merging, translation updates only

PR #931, opened 2016-09-26
Override submitter requirements on staff override assessment

efischer19:
## [TNL-4696](https://openedx.atlassian.net/browse/TNL-4696)

When a staff override assessment is recorded, we now treat all the submitter's
requirements as being fulfilled. This allows ORA tooling to match the progress
page reported grade for the student by placing the AssessmentWorkflow into a
DONE state.

### DIFF
--- a/openassessment/assessment/api/peer.py
+++ b/openassessment/assessment/api/peer.py
@@ -436,6 +436,8 @@ def get_assessment_median_scores(submission_uuid):
         assessments = [item.assessment for item in items]
         scores = Assessment.scores_by_criterion(assessments)
         return Assessment.get_median_score_dict(scores)
+    except PeerWorkflow.DoesNotExist:
+        return {}
     except DatabaseError:
         error_message = (
             u"Error getting assessment median scores for submission {uuid}"
--- a/openassessment/workflow/api.py
+++ b/openassessment/workflow/api.py
@@ -177,7 +177,7 @@ def get_workflow_for_submission(submission_uuid, assessment_requirements):
     return update_from_assessments(submission_uuid, assessment_requirements)
 
 
-def update_from_assessments(submission_uuid, assessment_requirements):
+def update_from_assessments(submission_uuid, assessment_requirements, override_submitter_requirements=False):
     """
     Update our workflow status based on the status of the underlying assessments.
 
@@ -259,7 +259,7 @@ def update_from_assessments(submission_uuid, assessment_requirements):
     workflow = _get_workflow_model(submission_uuid)
 
     try:
-        workflow.update_from_assessments(assessment_requirements)
+        workflow.update_from_assessments(assessment_requirements, override_submitter_requirements)
         logger.info((
             u"Updated workflow for submission UUID {uuid} "
             u"with requirements {reqs}"
--- a/openassessment/workflow/api.py
+++ b/openassessment/workflow/api.py
@@ -204,6 +204,9 @@ def update_from_assessments(submission_uuid, assessment_requirements, override_s
             `must_be_graded_by` to ensure that everyone will get scored.
             The intention is to eventually pass in more assessment sequence
             specific requirements in this dict.
+        override_submitter_requirements (bool): If True, the presence of a new
+            staff score will cause all of the submitter's requirements to be
+            fulfilled, moving the workflow to DONE and exposing their grade.
 
     Returns:
         dict: Assessment workflow information with the following
--- a/openassessment/workflow/models.py
+++ b/openassessment/workflow/models.py
@@ -171,7 +171,7 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
 
                 # If we auto-added a staff step, it is optional and should be marked complete immediately
                 if step.name == "staff" and staff_auto_added:
-                    step.assessment_completed_at=now()
+                    step.assessment_completed_at = now()
                     step.save()
 
                 # For the first valid step, update the workflow status
@@ -209,6 +209,12 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
         return score
 
     def status_details(self):
+        """
+        Returns workflow status in the form of a dictionary. Each step in the
+        workflow is a key, and each key maps to a dictionary defining whether
+        the step is complete (submitter requirements fulfilled) and graded (the
+        submission has been assessed).
+        """
         status_dict = {}
         steps = self._get_steps()
         for step in steps:
@@ -259,7 +265,7 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
 
         return score
 
-    def update_from_assessments(self, assessment_requirements):
+    def update_from_assessments(self, assessment_requirements, override_submitter_requirements=False):
         """Query assessment APIs and change our status if appropriate.
 
         If the status is done, we do nothing. Once something is done, we never
@@ -320,7 +326,9 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
                 # Update the assessment_completed_at field for all steps
                 # All steps are considered "assessment complete", as the staff score will override all
                 for step in steps:
-                    step.assessment_completed_at=now()
+                    step.assessment_completed_at = now()
+                    if override_submitter_requirements:
+                        step.submitter_completed_at = now()
                     step.save()
 
         if self.status == self.STATUS.done:
@@ -346,8 +354,10 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
 
         # If the submitter has done all they need to do, let's check to see if
         # all steps have been fully assessed (i.e. we can score it).
-        if (new_status == self.STATUS.waiting and
-            all(step.assessment_completed_at for step in steps)):
+        if (
+                new_status == self.STATUS.waiting and
+                all(step.assessment_completed_at for step in steps)
+        ):
 
             score = self.get_score(assessment_requirements, step_for_name)
             # If we found a score, then we're done
@@ -398,7 +408,7 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
 
         return steps
 
-    def set_staff_score(self, score, is_override=False, reason=None):
+    def set_staff_score(self, score, reason=None):
         """
         Set a staff score for the workflow.
 
@@ -425,9 +435,9 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
             self.submission_uuid,
             score["points_earned"],
             score["points_possible"],
-            annotation_creator = score["staff_id"],
-            annotation_type = annotation_type,
-            annotation_reason = reason
+            annotation_creator=score["staff_id"],
+            annotation_type=annotation_type,
+            annotation_reason=reason
         )
 
     def set_score(self, score):
@@ -602,9 +612,16 @@ class AssessmentWorkflowStep(models.Model):
         ordering = ["workflow", "order_num"]
 
     def is_submitter_complete(self):
+        """
+        Used to determine whether the submitter of the response has completed
+        their required actions.
+        """
         return self.submitter_completed_at is not None
 
     def is_assessment_complete(self):
+        """
+        Used to determine whether the response has been assessed at this step.
+        """
         return self.assessment_completed_at is not None
 
     def api(self):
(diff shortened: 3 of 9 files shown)


efischer19:
FYI @cahrens - will you or someone else from your subteam be able to be one of my reviewers for this PR, as Beryl is lacking ORA expertise? It's not quite ready yet - I still need to spin up a sandbox and update tests - just wanted to get it on your radar.

cahrens:
@efischer19 I'll do my best... I have a lot of other balls in the air right now, and need to review a lot of Chris's PRs this week. AJ doesn't have any ORA experience. If I can't do it, perhaps @dianakhuang can pick it up?

openassessment/workflow/models.py
```diff
@@ -171,7 +171,7 @@ def start_workflow(cls, submission_uuid, step_names, on_init_params):
 
                 # If we auto-added a staff step, it is optional and should be marked complete immediately
                 if step.name == "staff" and staff_auto_added:
-                    step.assessment_completed_at=now()
+                    step.assessment_completed_at = now()
```
efischer19: Note to reviewers: A fair number of the changes in this file are due to my text editor flagging quality failures that were easy to address.

openassessment/xblock/grade_mixin.py
```diff
@@ -567,15 +567,16 @@ def _assessment_grade_context(self, assessment):
...
-            criterion_label_key = part['criterion']['name']
-            part['criterion']['label'] = criterion_labels.get(criterion_label_key, part['criterion']['name'])
-
-            # We need to be a little bit careful here: some assessment parts
-            # have only written feedback, so they're not associated with any options.
-            # If that's the case, we don't need to add the label field.
-            if part.get('option') is not None:
-                option_label_key = (part['criterion']['name'], part['option']['name'])
-                part['option']['label'] = option_labels.get(option_label_key, part['option']['name'])
+        if assessment is not None:
```
efischer19: [click here](https://github.com/edx/edx-ora2/pull/931/files?w=1#diff-ac25c15240962030fe45023e217279ba) to see this change more clearly without whitespace changes (indented a block)

efischer19:
@catong This is the PR we spoke about earlier.

openassessment/workflow/models.py
```diff
@@ -320,7 +326,9 @@ def update_from_assessments(self, assessment_requirements):
                 # Update the assessment_completed_at field for all steps
                 # All steps are considered "assessment complete", as the staff score will override all
                 for step in steps:
-                    step.assessment_completed_at=now()
+                    step.assessment_completed_at = now()
+                    if override_submitter_requirements:
+                        step.submitter_completed_at = now()
```
efischer19: These 2 additional lines are the only functional (not quality-checker-based) changes in this file.

scripts/run-pep8.sh
```diff
@@ -1,12 +1,12 @@
...
 
-MAX_PEP8_VIOLATIONS=110
+MAX_PEP8_VIOLATIONS=111
 
 mkdir -p test/logs
 PEP8_VIOLATIONS=test/logs/pep8.txt
 touch $PEP8_VIOLATIONS
 
-pep8 --config=.pep8 openassessment > $PEP8_VIOLATIONS
+pep8 --config=.pep8 openassessment test > $PEP8_VIOLATIONS
```
efischer19: Fun fact: everything in the `test` directory was not being included in quality tests. I've added it here and updated the violations numbers to match.

test/acceptance/tests.py
```diff
@@ -8,13 +8,13 @@
...
-from bok_choy.promise import BrokenPromise, EmptyPromise
-from auto_auth import AutoAuthPage
-from pages import (
+from acceptance.auto_auth import AutoAuthPage
+from acceptance.pages import (
     SubmissionPage, AssessmentPage, GradePage, StaffAreaPage
 )
+from bok_choy.web_app_test import WebAppTest
+from bok_choy.promise import BrokenPromise, EmptyPromise
+from nose.plugins.attrib import attr
```
efischer19: There are multiple quality fixes in this file as well, which is how I first realized that `test/` was not being quality checked (made some fixes, and the reported # of violations did not decrease)

test/acceptance/tests.py
```diff
@@ -179,16 +178,14 @@ def submit_self_assessment(self, options=OPTIONS_SELECTED):
...
+    def _verify_staff_grade_section(self, expected_status):
         """
         Verifies the expected status and message text in the Staff Grade section
         (as shown to the learner).
         """
         self.staff_asmnt_page.wait_for_page()
         self.assertEqual("Staff Grade", self.staff_asmnt_page.label)
         self.staff_asmnt_page.verify_status_value(expected_status)
-        message_title = self.staff_asmnt_page.open_step().message_title
-        self.assertEqual(expected_message_title, message_title)
```
efischer19: `expected_message_title` was the now-irrelevant `STAFF_OVERRIDE_LEARNER_STEPS_NOT_COMPLETE` or `None` for all callers.

efischer19:
I'm leaving my second commit separate for now - could reviewers double-check my logic? It looked to me like there were 3 tests all checking essentially the same thing, so I removed the 2 least detailed of them.

efischer19:
@dianakhuang @jcdyer, could you be my reviewers? This is a feature change, outlined in the first comment on this PR, driven by decisions made on the linked ticket.

I've set up a sandbox and verified acceptance tests at https://efischer19.sandbox.edx.org/

@catong If you need screenshots, load https://efischer19.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/4dbd99f691064633a9be0daadb805ce6/ as the `audit` user - that user has made a submission, did not assess any peers, and was given a staff override grade.

catong:
@efischer19 Doc story is https://openedx.atlassian.net/browse/DOC-3368. When I have a doc PR I would like to tag you as a reviewer. (Will be no earlier than next week) Thanks!

efischer19:
Sounds great @catong, I'll keep an eye out

test/acceptance/tests.py
```diff
@@ -389,64 +386,6 @@ def test_peer_assessment(self):
         self.do_peer_assessment()
 
 
-class PeerAssessmentTestStaffOverride(OpenAssessmentTest):
```
dianakhuang: Where did this test go? Also, the one below called `test_staff_grade_override`.

openassessment/workflow/models.py
```diff
@@ -259,7 +265,7 @@ def get_score(self, assessment_requirements, step_for_name):
 
         return score
 
-    def update_from_assessments(self, assessment_requirements):
+    def update_from_assessments(self, assessment_requirements, override_submitter_requirements=False):
```
dianakhuang: Don't forget to update the docstring with the new arguments and what they mean.

test/acceptance/tests.py
efischer19: https://github.com/edx/edx-ora2/pull/931#issuecomment-249954309

dianakhuang: Whoops, sorry about missing that comment. It seems like even though the tests are doing the same thing, they're also starting with different problem configurations. One of them is set up with `peer_only` and the other is set up with `self_only`. I would say it makes sense to keep them both in for coverage's sake.

efischer19: I had thought that by leaving the third test (`full_workflow_staff_override`, which includes self and peer grading IIRC) we'd cover all the bases. I can leave these tests in though if you still think the difference in setup is worth it.

dianakhuang: I think it is. ORA is so massively complicated that I can imagine a bug that only affects problems with only one step in them that won't be caught by tests that have problems with two or more steps.

openassessment/assessment/api/peer.py
```diff
@@ -436,6 +436,8 @@ def get_assessment_median_scores(submission_uuid):
         assessments = [item.assessment for item in items]
         scores = Assessment.scores_by_criterion(assessments)
         return Assessment.get_median_score_dict(scores)
+    except PeerWorkflow.DoesNotExist:
+        return {}
```
jcdyer: Was this just crashing before?

openassessment/workflow/models.py
jcdyer: I'd recommend calculating now once, and using that for both completed_at values.  That way, we can correlate the values by equality.

openassessment/workflow/models.py
```diff
@@ -660,12 +677,12 @@ def update(self, submission_uuid, assessment_requirements):
         assessment_finished = getattr(self.api(), 'assessment_is_finished', default_finished)
```
jcdyer: Why are these values only sometimes available?  Or put another way, why do we need to do a `getattr()` dance on them?

test/acceptance/tests.py
jcdyer: Nit: Can you put a `from __future__ import absolute_import` at the top here to verify that implicit relative imports (like `from auto_auth import AutoAuthPage`) don't come back?

test/acceptance/tests.py
```diff
@@ -337,22 +334,22 @@ def test_staff_assessment(self):
...
-        self._verify_staff_grade_section("NOT AVAILABLE", None)
+        self._verify_staff_grade_section("NOT AVAILABLE")
 
         # User submits a response
         self.submission_page.submit_response(self.SUBMISSION)
         self.assertTrue(self.submission_page.has_submitted)
 
         # Verify staff grade section appears as expected
-        self._verify_staff_grade_section("NOT AVAILABLE", "Waiting for a Staff Grade")
+        self._verify_staff_grade_section("NOT AVAILABLE")
```
jcdyer: This is not None or "You Must Complete the Steps Above to View Your Grade".

test/acceptance/tests.py
```diff
@@ -522,8 +461,14 @@ def test_staff_area_panel(self):
...
-            ("staff-tools", "MANAGE INDIVIDUAL LEARNERS"),
-            ("staff-info", "VIEW ASSIGNMENT STATISTICS"),
+                (
+                    "staff-tools",
+                    "MANAGE INDIVIDUAL LEARNERS"
+                ),
+                (
+                    "staff-info",
+                    "VIEW ASSIGNMENT STATISTICS"
+                ),
```
jcdyer: Nit: I actually like the previous version's line breaks better.

openassessment/assessment/api/peer.py
efischer19: This was never called in a situation where no peer workflow existed before, and yes it did crash once that situation became a possibility.

test/acceptance/tests.py
efischer19: I did too, but pylint complained `¯\_(ツ)_/¯`

openassessment/workflow/models.py
efischer19: Correct - a given instance of `AssessmentWorkflowStep` can correspond to any of a number of `api()`-s (self, peer, staff, etc), and those apis are not forced to define these attributes, so we use the `lambda => True` default in those cases.

test/acceptance/tests.py
efischer19: Good catch!

test/acceptance/tests.py
jcdyer: What if you indent it twice, but keep each one on a single line?

openassessment/workflow/api.py
```diff
@@ -177,7 +177,7 @@ def get_workflow_for_submission(submission_uuid, assessment_requirements):
     return update_from_assessments(submission_uuid, assessment_requirements)
 
 
-def update_from_assessments(submission_uuid, assessment_requirements):
+def update_from_assessments(submission_uuid, assessment_requirements, override_submitter_requirements=False):
```
efischer19: Need to add new param to docstring here

efischer19:
Comments have been addressed in the latest two commits.

openassessment/assessment/api/peer.py
jcdyer: I didn't notice a test on this case.  Did I miss it, or was it not written?

openassessment/workflow/test/test_api.py
```diff
@@ -254,6 +255,39 @@ def test_unexpected_peer_workflow_errors_wrapped(self, mock_create):
...
+            "staff_id": "staff 1",
+        }
+        _, submission = self._create_workflow_with_status(
+            "user 1",
+            "test/1/1",
+            "peer-problem",
+            "peer",
+            steps=["peer"]
+        )
+        workflow_api.update_from_assessments(
```
efischer19: This is the entry point for the crashed-previously codepath. The key difference from prior behavior is that it's now possible to have a "completed" peer step without any peer workflows existing.

efischer19:
@jcdyer Ping for Monday review - the last two commits include the test you were asking about.

PR #933, opened 2016-10-05
Fix broken/flaky sandbox tests

efischer19:
These tests were broken on master as part of my latest PR.

- a11y tests were not updated, breaking a usage of the common `_verify_staff_grade_section` method.
- `verify_learner_final_score` runs ~~too quickly~~ differently on jenkins, ~~and needs a wait to succeed there~~ revealing some test method calls that were never properly updated to use a new default parameter. These should not have been passing locally

I'm going to have jenkins use this branch to run the tests; once they pass I'll go about updating the 1.1.9 release tag.

### DIFF
--- a/test/acceptance/accessibility.py
+++ b/test/acceptance/accessibility.py
@@ -160,7 +160,7 @@ class StaffAreaA11yTest(OpenAssessmentA11yTest):
 
         # Refresh the page, then verify accessibility of the Staff Grade section (marked Complete).
         self.browser.refresh()
-        self._verify_staff_grade_section(self.STAFF_GRADE_EXISTS, None)
+        self._verify_staff_grade_section(self.STAFF_GRADE_EXISTS)
 
         self._check_a11y(self.staff_asmnt_page)
 
--- a/test/acceptance/tests.py
+++ b/test/acceptance/tests.py
@@ -427,7 +427,7 @@ class StaffOverrideTest(OpenAssessmentTest):
         self.assertFalse(self.staff_asmnt_page.is_browser_on_page())
 
         # Submit a staff override.
-        self.do_staff_override(username, self.STAFF_OVERRIDE_STAFF_AREA_NOT_COMPLETE)
+        self.do_staff_override(username)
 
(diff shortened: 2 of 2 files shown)


efischer19:
Tests running at http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/1041/

test/acceptance/accessibility.py
```diff
@@ -160,7 +160,7 @@ def test_staff_grade(self):
 
         # Refresh the page, then verify accessibility of the Staff Grade section (marked Complete).
         self.browser.refresh()
-        self._verify_staff_grade_section(self.STAFF_GRADE_EXISTS, None)
+        self._verify_staff_grade_section(self.STAFF_GRADE_EXISTS)
```
efischer19: This method was updated to only take 1 parameter; I forgot to run a11y tests locally before merging so this method call snuck through.

test/acceptance/tests.py
```diff
@@ -1045,7 +1045,7 @@ def test_staff_override_at_beginning(self):
         self.staff_area_page.verify_learner_final_score(self.STAFF_OVERRIDE_STAFF_AREA_NOT_COMPLETE)
 
         # Do staff override
-        self.do_staff_override(learner, self.STAFF_OVERRIDE_STAFF_AREA_NOT_COMPLETE)
+        self.do_staff_override(learner)
```
efischer19: Both of the changes in this file are the same. I ought to have updated this method call in https://github.com/edx/edx-ora2/pull/931, but missed it due to a quirk of running tests locally (I imagine my devstack was slow enough that the previous `STAFF_OVERRIDE_STAFF_AREA_NOT_COMPLETE` string was visible longer than it ought to have, Jenkins was fast enough to expose this)

efischer19:
I'm merging this to master, as it's only test fixes. The jenkins system has kicked off the daily "rebuild the sandbox and run acceptance tests job" (which fails any currently running test runs, it's hard to ping a server that's being rebuilt), the end result of that will be confirmation that my changes have worked.

PR #939, opened 2016-11-07
Update bok choy, selenium versions

efischer19:
I'm fairly positive this is due to the fact that grades handlers are now taking longer to respond than before. More to come.

Update: yup, that was it.


efischer19:
Turns out, none of this is needed - the tests are passing again now that some optimizations to grade handlers have landed in edx-platmorm master.

I'm going to back out everything except the version upgrades (which really ought to be merged), run the tests a few more times, and close this ticket.

PR #943, opened 2016-11-16
Make disabled buttons truly disabled.

cahrens:
### Description

For all buttons shown in the LMS, use disabled attribute instead of disabled CSS class. The disabled CSS class allowed the button to still be clickable via keyboard navigation, and it caused color contrast a11y failures.

Note that _is--disabled_ is still used some in Studio for the ORA editing UI (on divs no less...). I created [TNL-5958](https://openedx.atlassian.net/browse/TNL-5958) for that.

### Sandbox
- [x] https://tip.sandbox.edx.org/
- [x] RTL (buttons look fine, although other parts of ORA have RTL issues)

### Testing

### Reviewers
- [x] Code review: @staubina 
- [ ] Code review: @andy-armstrong 
- [x] UX review: @chris-mike 

FYI @catong @sstack22 @cptvitamin 

### Post-review


openassessment/templates/openassessmentblock/grade/oa_grade_complete.html
```diff
@@ -144,7 +144,7 @@ <h4 id="feedback__overall__prompt__{{ xblock_id }}" class="field-group__label">{
                                                            class="option__input feedback__overall__value"
                                                            value="These assessments were useful."
                                                            aria-labelledby="feedback__overall__prompt__{{ xblock_id }}"/>
-                                                    <label class="option__label" for="feedback__overall__value--useful">{% trans "These assessments were useful." %}</label>
+                                                    <label class="option__label" for="feedback__overall__value--useful__{{ xblock_id }}">{% trans "These assessments were useful." %}</label>
```
cahrens: These checkboxes are part of a survey for learners to say how useful their peer grading experience was.

I just happened to notice that these labels were not correctly linked to their inputs (Pycharm had highlighted them as red).

As far as I know, there is no way to enable this part of the HTML via Studio. I manually edited the HTML file to remove the conditional that hides these options on my devstack, but these won't be enabled on the sandbox.

openassessment/templates/openassessmentblock/peer/oa_peer_assessment.html
```diff
@@ -94,8 +94,8 @@ <h3 class="message__title">{% trans "We could not submit your assessment" %}</h3
 
                 <ul class="list list--actions">
                     <li class="list--actions__item">
-                        <button type="submit" class="peer-assessment--001__assessment__submit action action--submit is--disabled">
-                            <span class="copy">{{ submit_button_text }}</span>
+                        <button type="submit" class="peer-assessment--001__assessment__submit action action--submit" disabled>
```
cahrens: No more "is--disabled" class on the LMS side of things.

Also, I removed the extra span around button text. The a11y tests were failing due to color contrast when the spans were present and the button was disabled. Apparently the axe toolkit isn't very smart about examining children of disabled components (disabled components do not need to meet color contrast standards).

openassessment/templates/openassessmentblock/peer/oa_peer_cancelled.html
```diff
@@ -8,7 +8,7 @@
 {% endblock %}
 
 {% block button %}
-    <button class="ui-slidable" aria-expanded="false" disabled="disabled">
+    <button class="ui-slidable" aria-expanded="false" disabled>
```
cahrens: The most correct syntax is simply _disabled_, though other variations seem to work in most browsers. Went ahead and cleaned this up while searching for usages of disabled.

openassessment/templates/openassessmentblock/response/oa_response.html
```diff
@@ -103,22 +103,22 @@ <h3 class="message__title">{% trans "We could not save your progress" %}</h3>
 
                         <ul class="list list--actions">
                             <li class="list--actions__item">
-                                <button type="submit" class="action action--save submission__save is--disabled" aria-describedby="response__save_status__{{ xblock_id }}">
+                                <button type="submit" class="action action--save submission__save" aria-describedby="response__save_status__{{ xblock_id }}" disabled>
                                     {% trans "Save your progress" %}
                                 </button>
 
-                                <div id="response__save_status__{{ xblock_id }}" class="response__submission__label">
+                                <div id="response__save_status__{{ xblock_id }}" class="save__submission__label response__submission__label">
```
cahrens: Note that "Preview in LaTeX" (which I can't figure out how to enable on sandbox... I hacked the HTML file to make it visible on my devstack) has a text area with the class response__submission__label as well. They both use that class for styling of text beside the button. However, there is JS that is injecting a message about saving the response, and because the response__submission__label class was used, the message erroneously appeared beside both buttons.

Added save__submission__label so that the JS will only inject the message into the text area beside the Save button.

openassessment/templates/openassessmentblock/response/oa_response.html
```diff
@@ -103,22 +103,22 @@ <h3 class="message__title">{% trans "We could not save your progress" %}</h3>
...
                             </li>
                             {% if allow_latex %}
                                 <li class="list--actions__item">
-                                    <button type="submit" class="submission__preview  action action--save submission__save is--disabled" aria-describedby="response__preview_explanation__{{ xblock_id }}">
+                                    <button type="submit" class="submission__preview  action action--save" aria-describedby="response__preview_explanation__{{ xblock_id }}">
                                         {% trans "Preview in LaTeX"%}
                                     </button>
                                     <div id="response__preview_explanation__{{ xblock_id }}" class="response__submission__label">
-                                        {% trans "Click to preview your submission in LaTeX"%}
+                                        {% trans "Click to preview your submission in LaTeX."%}
```
cahrens: Added a period to be consistent with other messages.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -144,13 +144,13 @@ <h2 class="staff-info__title ui-staff__subtitle">
                     </tbody>
                 </table>
             {% elif workflow_status == "waiting" %}
-                <div class="staff-info__final__grade__score">{% trans "The submission is waiting for assessments." %}</div>
+                <div class="staff-info__final__grade__score"><p>{% trans "The submission is waiting for assessments." %}</p></div>
```
cahrens: Added p-tags around these message to make the styling more consistent with other parts of the staff grading UI (a different font was being used without the p-tags).

openassessment/xblock/static/js/spec/lms/oa_staff_area.js
```diff
@@ -643,9 +643,9 @@ describe('OpenAssessment.StaffAreaView', function() {
             $assessment = getAssessment(staffArea, staffAreaTab);
             $submitButtons = $('.action--submit', $assessment);
             expect($submitButtons.length).toBe(2);
-            expect($submitButtons).toHaveClass('is--disabled');
+            expect($submitButtons).toHaveAttr('disabled');
```
cahrens: Since this test is specifically checking 2 different buttons, I changed it to look directly at the disabled attribute instead of using helper methods.

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -228,7 +206,7 @@ OpenAssessment.ResponseView.prototype = {
      string: The current status message.
      **/
     saveStatus: function(msg) {
-        var sel = $('.response__submission__label', this.element);
+        var sel = $('.save__submission__label', this.element);
```
cahrens: Related to the message beside the Save button (discussed above).

openassessment/xblock/static/sass/xb/elements/_controls.scss
```diff
@@ -41,20 +41,11 @@
     text-decoration: none;
   }
 
-  &.disabled, &[disabled], &.is--disabled {
-    cursor: default;
```
cahrens: Seems unnecessary, as the buttons are now truly disabled.

openassessment/xblock/static/sass/xb/elements/_controls.scss
```diff
@@ -41,20 +41,11 @@
...
-  }
-
   // CASE: icons in buttons
-  .copy, .icon {
+  .icon {
     display: inline-block;
     vertical-align: middle;
   }
-
-  .copy {
```
cahrens: Button text appears to be fine even without the inner span (and its associated css).

openassessment/xblock/static/sass/xb/elements/_controls.scss
```diff
@@ -92,10 +83,6 @@
   border: 1px solid $action-primary-color;
   color: $copy-inverse-color;
 
-  .copy {
```
cahrens: Button text is still showing with the same weight.

openassessment/templates/openassessmentblock/grade/oa_grade_complete.html
staubina: Is trans how they are handling i18n? I am looking at the files also.

openassessment/xblock/static/js/src/lms/oa_peer.js
```diff
@@ -159,13 +154,7 @@ OpenAssessment.PeerView.prototype = {
      >> true
      **/
     peerSubmitEnabled: function(enabled) {
-        var button = $('.peer-assessment--001__assessment__submit', this.element);
```
staubina: Nice removal of duplicated code.

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -523,11 +508,11 @@ OpenAssessment.ResponseView.prototype = {
     fileUpload: function() {
         var view = this;
         var fileUpload = $(".file__upload");
-        fileUpload.addClass("is--disabled");
+        fileUpload.prop('disabled', true);
 
         var handleError = function(errMsg) {
             view.baseView.toggleActionError('upload', errMsg);
-            fileUpload.removeClass("is--disabled");
+            fileUpload.prop('disabled', false);
```
staubina: This will add disabled="true" right?  I thought we were attempting to just use disabled as a property now?

openassessment/templates/openassessmentblock/grade/oa_grade_complete.html
cahrens: Yes, this is how i18n is done in ORA.

openassessment/xblock/static/js/src/lms/oa_peer.js
cahrens: Yes, it was satisfying (though also frightening that previous authors felt OK about duplicating the same thing 9 times).

openassessment/xblock/static/js/src/lms/oa_response.js
cahrens: @staubina I can't remember exactly how you interacted with "disabled" in your recent PR. Do you think something should be updated there to reflect the documentation above?

staubina: I have to look at it again.  

So if I am understanding this correctly, this will only add disabled when you use `$( "input" ).prop( "disabled", true );` ?

I believe I used attr and removeAttr().

cahrens: Right, and 

> $( "input" ).prop( "disabled", false );

 will remove the disabled attribute. At least that's what I saw when I tried debugging it, and it seems consistent with their documentation.

cahrens: It may be that attr/removeAttr was the correct way to do it in older versions of JQuery. I did read something about that here--

http://stackoverflow.com/questions/6048022/jquery-attrdisabled-disabled-not-working-in-chrome

staubina: OK, I will create a small pr to change it on the one file if you think it is necessary.

staubina:
:shipit: 👍

openassessment/xblock/static/sass/xb/elements/_controls.scss
```diff
@@ -101,6 +101,7 @@
     color: $action-primary-color-disabled;
     border-color: $action-primary-color-disabled;
     background: $action-primary-color-disabled-back;
+    opacity: 1;
```
cahrens: Since these buttons set their colors explicitly based on disabled state, they shouldn't respect the default opacity setting.

cahrens:
@staubina FYI that I just pushed up another commit to address a styling issue I was seeing in Studio with disabled step labels. This is how it looked before:

image

And after:
image

The sandbox has been updated to include this change.

This issue existed on Studio only. Opacity is set to .5 in LMS for disabled buttons, but currently that is not true in Studio. This may be because Studio still has a lot of elements that are "disabled" via classes only, and thus the CSS over there hasn't been updated yet to handle truly disabled elements. But in any case, it seems wrong for ORA to depend on styling in edx-platform, and therefore I explicitly added the opacity setting to ORA's CSS.

staubina:
👍

openassessment/templates/openassessmentblock/grade/oa_grade_complete.html
andy-armstrong: @staubina FYI, this is how i18n works in Django Templates in general:

https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#internationalization-in-template-code

openassessment/templates/openassessmentblock/peer/oa_peer_cancelled.html
andy-armstrong: FYI, this is a legacy from XHTML that never stuck.

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
andy-armstrong: Nit: I'd put the paragraph on its own line for all of these changes. These lines are already long, and paragraphs are semantically their own thing. It also makes the ``<p>`` tag more visible, as it is somewhat hidden when tucked away here.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -257,6 +257,24 @@ OpenAssessment.BaseView.prototype = {
...
+     *
+     * @param {string} className The css class to find the button
+     * @param {boolean} enabled If specified enables or disables the button. If not specified,
+     *     the state of the button is not changed, but the current enabled status is returned.
+     * @returns {boolean} whether or not the button is enabled
+     */
+    buttonEnabled: function(className, enabled) {
+        var $element = $(className, this.element);
+        if (typeof enabled === 'undefined') {
+            return !$element.attr('disabled');
```
andy-armstrong: Nit: it seems odd to use ``attr`` here but ``prop`` when setting it. Is it possible to use ``prop`` for both?

openassessment/xblock/static/js/src/lms/oa_peer.js
andy-armstrong: 💯

openassessment/xblock/static/sass/xb/elements/_controls.scss
andy-armstrong: 👍

openassessment/xblock/static/sass/xb/utilities/_reset.scss
```diff
@@ -288,6 +288,14 @@
   }
 
   /**
+   * Set opacity on buttons to 50% (which matches LMS styling).
```
andy-armstrong: I'd rather this use the Pattern Library definition. Does that also use opacity of 50%? The LMS will be switching over to Pattern Library buttons.

chris-mike:
👍 Looks good.

openassessment/xblock/static/js/src/lms/oa_base.js
cahrens: Sure, will do. FYI, this return value is only used by unit tests.

openassessment/xblock/static/sass/xb/utilities/_reset.scss
cahrens: @andy-armstrong the pattern library doesn't address buttons that are styled in this way. It only covers buttons where the disabled colors are explicitly set (which is what I patterned _btn--primary_ and _btn--secondary_ off of).

http://ux.edx.org/components/buttons/

These buttons are styled to look like links, with no outlines and backgrounds. Let's chat about how the variety of different buttons will be handled when LMS is using the pattern library.

PR #944, opened 2016-11-17
Multiple file uploads

dmitry-viskov:
Ability to attach more than one file.

New selector in Studio to choose number of available files:
ora_cms

LMS part:
ora_lms1

ora_lms2

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -25,3 +25,4 @@ Eric Fischer <efischer@edx.org>
 Andy Armstrong <andya@edx.org>
 Christina Roberts <christina@edx.org>
 Mushtaq Ali <mushtaak@gmail.com>
+Dmitry Viskov <dmitry.viskov@webenterprise.ru>
--- a/openassessment/templates/openassessmentblock/edit/oa_edit.html
+++ b/openassessment/templates/openassessmentblock/edit/oa_edit.html
@@ -128,6 +128,16 @@
                         </p>&nbsp;
                         <p class="setting-help message-status error"></p>
                     </div>
+                    <div id="openassessment_submission_file_count_wrapper" class="{% if not file_upload_type %}is--hidden{% endif %}">
+                         <label for="openassessment_submission_file_count" class="setting-label">{% trans "Count of Files to Upload"%}</label>
+                         <input
+                             id="openassessment_submission_file_count"
+                             class="input setting-input"
+                             type="number"
+                             value="{{ upload_file_count }}"
+                             min="0" max="99"
+                         />
+                     </div>
                 </li>
                 <li id="openassessment_submission_latex_wrapper" class="field comp-setting-entry">
                     <div class="wrapper-comp-setting">
@@ -152,7 +162,6 @@
                     </div>
                     <p class="setting-help">{% trans "Specify the number of top scoring responses to display. Valid numbers are 0 to 99. If the number is 0, the Top Responses section does not appear to learners." %}</p>
                 </li>
-                <li>
             </ul>
 
             <p class="openassessment_description" id="openassessment_step_select_description">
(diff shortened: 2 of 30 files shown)


edx-webhook:
Thanks for the pull request, @dmitry-viskov! I've created [OSPR-1594](https://openedx.atlassian.net/browse/OSPR-1594) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

cahrens:
@dmitry-viskov it would be a lot more attractive if multi-select was enabled on a single file chooser. Did you try that?

FYI @sstack22

dmitry-viskov:
@cahrens not quite understand what do you mean. If you mean to add the "multiple" attribute to the current existed select box "Allow File Upload" I think it does not help in the case if for example I want to provide ability to load more than one images

cahrens:
What I mean is that the normal user experience is that one can select more than one file at a time (with a single chooser), vs. having to upload a single file at a time through multiple choosers.

dmitry-viskov:
@cahrens Do you mean modify file selector to something like this `<input type="file" multiple />` ?
I think this is a controversial issue. Probably from the abstract UI point of view it is better to use one HTML selector but from another point of view (and based on the current existed UI) if student see that there are 3 separate fields he explicitly knows that uploaded files could not be more than three. Otherwise we need to add some additional text fields with explanation or something like this (+ functionality for validation). Also in this case you need to create "remove" feature because without it user can't replace one of the uploaded files. I think my feature could be the first step of implementation Multiple file uploads. And it could be improved in the future with implementation one field instead of many.

dmitry-viskov:
@marcotuts So as I understand you correctly we should stay studio settings as they exists now and allow all students to upload unlimited count of files? (whether the maximum number of uploaded files limit?)

cahrens:
I think there has to be some upper limit to the number of files (and/or size) for platform stability.

marcotuts:
Is there no upper limit on file size today? If we enabled multiple file upload we might want to specify both a per-file size and a total-uploads file size limit (or just use the per-file size limit as the overall upper bound no matter how many files). 

@dmitry-viskov - Yes, the suggestion would be to not necessarily change any Studio settings, move to using a multiple file input in the html, and then have some sort of file size limit as @cahrens  noted. Does that make sense? Do you believe much if anything is lost with that suggested approach?

dmitry-viskov:
@marcotuts 

>Is there no upper limit on file size today? 
Upper limit already exists:
```
// Maximum file size (5 MB) for an attached file.
MAX_FILE_SIZE: 5242880
```
https://github.com/edx/edx-ora2/blob/9b682b64b4fab0ae950577a080432615cf816598/openassessment/xblock/static/js/src/lms/oa_response.js#L39
https://github.com/edx/edx-ora2/blob/9b682b64b4fab0ae950577a080432615cf816598/openassessment/xblock/static/js/src/lms/oa_response.js#L470

>Does that make sense? Do you believe much if anything is lost with that suggested approach?

I not quite understand about file size and count limitations.
1. Let allow to upload unlimited count of files? If no we need to save the suggested Studio setting (Count of Files to upload).
2. Let all upload files will be limited with `MAX_FILE_SIZE` constant ? If no we need some interface to set upload size for each file. I suggest to start with `MAX_FILE_SIZE` for each upload file.

marcotuts:
@dmitry-viskov  Thanks for the update here, my replies below:

1.) I'm ok removing the option to limit file count, I don't know that this is a critical setting for authors. This would remove the introduced studio setting. 

2.) One concern about having the 5mb be for each file uploaded instead of the total upload size being 5mb is that you could upload hundreds of files (accidentally or otherwise) that are each under 5mb. That could cause performance exposure / issues that we likely don't want to introduce.  My suggestion would be to make the 5mb limit be a limit for the cumulative file size uploaded, whether that be in 1 file or 100 files.

In the future we could have a separate story introduce a second file size max cap that is higher than 5mb. That way for example you could set MAX-FILE-SIZE to 5mb but a new MAX-UPLOAD-TOTAL-SIZE (or something with a better name than what I suggested :) to be 25mb. That way learners could upload 5 different ~5mb sized files, which limits the accidental upload of many more files. 

Thoughts on this? Thanks!

dmitry-viskov:
@marcotuts ok. I'll update PR when have a free time

cahrens:
Hi @dmitry-viskov, given the age of this PR, I'm going to go ahead and close it. If you have time to come back to this project, please just open a new PR.

PR #945, opened 2016-11-18
Enable the preview button if the submit button is enabled.

cahrens:
Really, the preview button should be enabled whenever there
is something to render. Since the submit button will only be
disabled upon initial rendering if there is nothing to submit,
it seems like a reasonable proxy.

@andy-armstrong and @staubina can you please review? The code is on https://tip.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/a4dfec19cf9b4a6fb5b18be6ccd9cecc/338a4affb58a45459629e0566291381e/ (link to a problem with LaTeX preview enabled).

I have run both acceptance and a11y tests with these changes, and they all passed.

### DIFF
--- a/openassessment/templates/openassessmentblock/response/oa_response.html
+++ b/openassessment/templates/openassessmentblock/response/oa_response.html
@@ -115,7 +115,8 @@
                             </li>
                             {% if allow_latex %}
                                 <li class="list--actions__item">
-                                    <button type="submit" class="submission__preview  action action--save" aria-describedby="response__preview_explanation__{{ xblock_id }}">
+                                    <button type="submit" class="submission__preview  action action--save" aria-describedby="response__preview_explanation__{{ xblock_id }}"
+                                        {{submit_enabled|yesno:",disabled" }}>
                                         {% trans "Preview in LaTeX"%}
                                     </button>
                                     <div id="response__preview_explanation__{{ xblock_id }}" class="response__submission__label">
--- a/test/acceptance/pages.py
+++ b/test/acceptance/pages.py
@@ -85,7 +85,7 @@ class OpenAssessmentPage(BaseAssessmentPage):
         """
         submit_button_selector = self._bounded_selector(button_css)
         EmptyPromise(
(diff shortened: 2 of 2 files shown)


openassessment/templates/openassessmentblock/response/oa_response.html
```diff
@@ -115,7 +115,8 @@ <h3 class="message__title">{% trans "We could not save your progress" %}</h3>
                             </li>
                             {% if allow_latex %}
                                 <li class="list--actions__item">
-                                    <button type="submit" class="submission__preview  action action--save" aria-describedby="response__preview_explanation__{{ xblock_id }}">
+                                    <button type="submit" class="submission__preview  action action--save" aria-describedby="response__preview_explanation__{{ xblock_id }}"
```
cahrens: Once I fixed the test, it failed because the Preview button was enabled upon initial rendering. The logic about when it should be enabled is a bit funky, but I decided that for initial rendering, going with whether or not there is something to Submit is good enough.

cahrens: FYI, before my previous PR, the preview button was "disabled" always upon initial rendering. I removed that because I thought you should be able to preview your response as long as you had non-empty text, but I didn't add anything to make sure that it was disabled if there was no text present.

test/acceptance/pages.py
```diff
@@ -85,7 +85,7 @@ def submit(self, button_css=".action--submit"):
         """
         submit_button_selector = self._bounded_selector(button_css)
         EmptyPromise(
-            lambda: 'is--disabled' not in " ".join(self.q(css=submit_button_selector).attrs('class')),
+            lambda: False == any(self.q(css=submit_button_selector).attrs('disabled')),
```
andy-armstrong: Nit: can you use ``not any(...``, as ``False == any(`` is an odd looking idiom.

staubina:
👍

test/acceptance/pages.py
cahrens: I think I'll leave this one. I was going with the doc string for the method:

> Return True if any element of the iterable is true. If the iterable is empty, return False.

PR #951, opened 2016-12-02
Actually check if staff grade is staff grade

efischer19:
Companion PR to take advantage of the functionality made in https://github.com/edx/edx-submissions/pull/50

### DIFF
--- a/openassessment/assessment/test/test_staff.py
+++ b/openassessment/assessment/test/test_staff.py
@@ -302,6 +302,24 @@ class TestStaffAssessment(CacheResetTest):
         self.assertEqual(tim_workflow["score"], None)
         self.assertNotEqual(tim_workflow["status"], "done")
 
+    def test_update_with_override(self):
+        """
+        Test that, when viewing a submission with a staff override present, the workflow is not updated repeatedly.
+
+        See TNL-6092 for some historical context.
+        """
+        tim_sub, tim = TestStaffAssessment._create_student_and_submission("Tim", "Tim's answer", problem_steps=['self'])
+        tim_assessment = staff_api.create_assessment(
+            tim_sub["uuid"],
+            "Dumbledore",
+            OPTIONS_SELECTED_DICT["none"]["options"], dict(), "",
+            RUBRIC,
+        )
+        tim_workflow = workflow_api.get_workflow_for_submission(tim_sub["uuid"], {})
+        with mock.patch('openassessment.workflow.models.sub_api.reset_score') as mock_reset:
+            tim_workflow = workflow_api.get_workflow_for_submission(tim_sub["uuid"], {})
+            self.assertFalse(mock_reset.called)
+
     def test_invalid_rubric_exception(self):
         # Create a submission
         tim_sub, tim = self._create_student_and_submission("Tim", "Tim's answer")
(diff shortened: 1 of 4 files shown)


efischer19:
The fallout of this bug is that, in the event of a staff score being present, the score will never be recognized as a previously set staff score, so the staff score is "updated" every single time the workflow is read, which causes a lot of noise in the grades signal handlers.

efischer19:
@staubina @cahrens @nasthagiri @jcdyer @sanfordstudent @yro Whoever is reviewing my edx-submissions PR, this ORA companion is ready to be reviewed as well.

The submissions work should be merged first; I'll update the requirements listed in this repo when I squash/rebase before merging this PR.

openassessment/workflow/models.py
```diff
@@ -316,8 +318,8 @@ def update_from_assessments(self, assessment_requirements, override_submitter_re
             old_score = sub_api.get_latest_score_for_submission(self.submission_uuid)
             if (
                     not old_score or # There is no recorded score
-                    not old_score.get('staff_id') or # The recorded score is not a staff score
-                    old_score['points_earned'] != new_staff_score['points_earned'] # Previous staff score doesn't match
+                    not self.STAFF_ANNOTATION_TYPE in old_score.get('annotation_types', []) or  # Not a staff score
```
nasthagiri: What's the rationale for the defensive coding?  When will `old_score` ever _not_ have `annotation_types` now?

openassessment/assessment/test/test_staff.py
```diff
@@ -302,6 +302,22 @@ def test_provisionally_done(self):
         self.assertEqual(tim_workflow["score"], None)
         self.assertNotEqual(tim_workflow["status"], "done")
 
+    def test_update_with_override(self):
+        """
+        Test that, when viewing a submission with a staff override present, the workflow is not updated repeatedly.
```
nasthagiri: Perhaps putting the TNL ticket in the docstring will help clarify the reason for this test.

openassessment/workflow/models.py
efischer19: This is actually largely a result of a [previous test run](https://travis-ci.org/edx/edx-ora2/builds/180817774), and allowed me to not error out during "do/don't use the updated edx-submissions code" A/B testing I was running. Should I remove it, since it isn't strictly needed?

nasthagiri: I would vote for removing it since it could potentially hide future issues - making them harder to debug.

cahrens:
👍, but I would suggest running acceptance tests against this change, if you haven't already.

efischer19:
Will do @cahrens, I'll run those overnight before proceeding with the release. Thanks!

openassessment/assessment/test/test_staff.py
```diff
@@ -1,13 +1,15 @@
 # coding=utf-8
+"""
+Tests for staff assessments.
+"""
```
efischer19: The number of lines in this diff view has changed significantly, as I was getting a frustrating quality failure on travis. To fix it, I went through and manually fixed all the quality errors in this file, and lowered the error threshold to match.

nasthagiri: Thanks for keeping those changes in a separate commit.

PR #958, opened 2017-01-03
Status updates sent to the Aria live region as the user completes eac…

staubina:
…h section of an ORA2 problem

### Description

This PR focuses on Sub Task https://openedx.atlassian.net/browse/TNL-6241

### Sandbox
https://staubina-ora.sandbox.edx.org/

### Testing

### Reviewers
- [x] Code review: @cahrens 
- [x] Accessibility review: @cptvitamin 

### Post-review

FYI: @jlajoie


openassessment/templates/openassessmentblock/peer/oa_peer_waiting.html
```diff
@@ -13,7 +13,7 @@
 {% block title %}
 <span class="step__status">
   <span class="step__status__value">
-    <span class="copy">{{review_num}}
+    <span class="copy">
```
cahrens: Was this just a bug?

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,67 @@ OpenAssessment.BaseView.prototype = {
     },
 
     /**
+     * Checks the rendering status of the views.
+     * @returns {boolean} true if any step's view is still loading.
```
cahrens: Learner-facing views? I note that messageView, leaderboardView, and staffAreaView are not listed here.

If new views are added, it would be helpful to understand the contract of this method, so as to know whether or not the new views should be added here as well.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,67 @@ OpenAssessment.BaseView.prototype = {
...
+            this.srStatusUpdates.push(text);
+        }
+        if (!this.isLoading() && this.srStatusUpdates.length > 0){
+            window.SR.readTexts(this.srStatusUpdates);
+            this.srStatusUpdates = []
+        }
+    },
+
+    /**
+     * Retrieves and returns the current status of a given step.
```
cahrens: It is not clear to me how this method differs from getAssessmentGradeStatus.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,67 @@ OpenAssessment.BaseView.prototype = {
...
+     *
+     * @param stepID - The id of the Step to retrieve grade status for.
+     * @param usageID - The Usage ID of the current grade status.
+     * @returns {string} - the current grade status.
+     */
+    getAssessmentGradeStatus: function(stepID, usageID) {
+        var block_id = "[data-usage-id='" + usageID + "'] ";
+        var css_base = block_id + stepID + " .step__header .step__title ";
+        var css_string_title = css_base + ".step__label";
+        var css_string_status = css_base + ".grade__value";
```
cahrens: It looks like this line is the only difference between this method and getAssessmentStatus. Can refactor and share code.

openassessment/xblock/static/js/src/lms/oa_grade.js
```diff
@@ -20,17 +22,31 @@ OpenAssessment.GradeView.prototype = {
...
+                if (typeof usageID !== 'undefined' &&
+                    $(stepID, view.element).hasClass("is--showing")){
+                    view.announceStatusChange = true;
+                }
+                if (view.announceStatusChange) {
+                    // Update the SR and announce the status change
+                    view.baseView.announceStatusChangeToSR(view.baseView.getAssessmentGradeStatus(stepID, usageID));
+                    view.announceStatusChange = false;
+                } else {
+                    view.baseView.announceStatusChangeToSR();
```
cahrens: I'm not clear about the logic of why this method is called with no arguments.

openassessment/xblock/static/js/src/lms/oa_grade.js
```diff
@@ -20,17 +22,31 @@ OpenAssessment.GradeView.prototype = {
...
                 if (typeof usageID !== 'undefined' &&
                     !$(stepID, view.element).hasClass("is--unfinished") &&
                     !$(stepID, view.element).hasClass("is--unstarted") &&
                     !$(stepID, view.element).hasClass("is--waiting--staff")) {
                     $("[id='oa_grade_" + usageID + "']", view.element).focus();
+                    view.announceStatusChange = true;
+                }
+                if (typeof usageID !== 'undefined' &&
+                    $(stepID, view.element).hasClass("is--showing")){
+                    view.announceStatusChange = true;
```
cahrens: Can you explain why in this case we do want to announce a status change, but we don't want to focus (a comment in the code would be helpful)?

openassessment/xblock/static/js/src/lms/oa_peer.js
```diff
@@ -26,14 +28,28 @@ OpenAssessment.PeerView.prototype = {
     load: function(usageID) {
         var view = this;
         var stepID = ".step--peer-assessment";
+
+        view.isRendering = true;
         this.server.render('peer_assessment').done(
             function(html) {
```
cahrens: The duplication between the various JS classes is painful. Is there any way to create a shared "done" handler?

openassessment/xblock/static/js/src/lms/oa_peer.js
```diff
@@ -123,9 +152,14 @@ OpenAssessment.PeerView.prototype = {
                 // Override default form submission
                 eventObject.preventDefault();
 
+
+                //Status will change in update announce it to the Screen Reader after Render
+                view.announceStatusChange = true;
```
cahrens: I'm not clear why this explicit call is necessary (as opposed to the handling above in the render callbacks.

openassessment/xblock/static/js/src/lms/oa_peer.js
```diff
@@ -14,6 +14,8 @@ OpenAssessment.PeerView = function(element, server, baseView) {
     this.server = server;
     this.baseView = baseView;
     this.rubric = null;
+    this.isRendering = false;
```
cahrens: Should this be pushed down into the baseView?

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -24,6 +24,8 @@ OpenAssessment.ResponseView = function(element, server, fileUploader, baseView,
     this.autoSaveTimerId = null;
     this.data = data;
     this.fileUploaded = false;
+    this.announceStatus = false;
```
cahrens: Why announceStatus vs. announceStatusChange?

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -423,9 +435,13 @@ OpenAssessment.ResponseView.prototype = {
     moveToNextStep: function() {
         var baseView = this.baseView;
         var usageID = baseView.getUsageID();
+        var view = this;
+
         this.load(usageID);
         baseView.loadAssessmentModules(usageID);
 
+        view.announceStatus = true;
```
cahrens: Why is this set here vs. the render callback (like in the other views)?

cahrens:
@staubina it's not clear to me yet how you are determining when status text needs to be read vs. not. And I feel like there should be some way to share code, but it seems that there are slight variations in some of the views that I don't understand.

Let's talk through this in person tomorrow morning.

openassessment/templates/openassessmentblock/peer/oa_peer_waiting.html
staubina: I believe so, I need to verify with Ux.

openassessment/xblock/static/js/src/lms/oa_base.js
staubina: Yes. In the next PR.

openassessment/xblock/static/js/src/lms/oa_grade.js
staubina: I was avoiding changing existing functionality, there is a specific case on line 37 that has the focus.  I will test this case though to see if it would be valuable to focus here as well.

openassessment/xblock/static/js/src/lms/oa_grade.js
staubina: The views can complete their render in a different order, this method without parameters checks to see if the views are still rendering and if they are done adds the text to the aria live region.

openassessment/xblock/static/js/src/lms/oa_peer.js
staubina: This call sets the flag on a submit to announce the status change after the render is complete.

openassessment/xblock/static/js/src/lms/oa_response.js
staubina: Typo.

openassessment/xblock/static/js/src/lms/oa_response.js
staubina: This is on submit.  Then when the page refreshes and renders from the submit it will announce the status.

openassessment/xblock/static/js/src/lms/oa_base.js
staubina: messageView, I am still looking at.

leaderboardView we discussed.  I do not believe it needs an aria-live update.

I will take a look at staffAreaView, My initial thought was that it would not need an aria live region update.  That might be wrong though.

openassessment/xblock/static/js/src/lms/oa_grade.js
```diff
@@ -20,17 +22,31 @@ OpenAssessment.GradeView.prototype = {
...
         var stepID = ".step--grade";
+        view.isRendering = true;
         this.server.render('grade').done(
             function(html) {
                 // Load the HTML and install event handlers
                 $(stepID, view.element).replaceWith(html);
                 view.server.renderLatex($(stepID, view.element));
+                view.isRendering = false;
                 view.installHandlers();
                 if (typeof usageID !== 'undefined' &&
```
cahrens: Simplify to just is--showing if possible (for both focus change and announce status).

openassessment/xblock/static/js/src/lms/oa_self.js
```diff
@@ -134,6 +148,7 @@ OpenAssessment.SelfView.prototype = {
         ).done(
             function() {
                 baseView.unsavedWarningEnabled(false, view.UNSAVED_WARNING_KEY);
+                view.announceStatusChange = true;
                 baseView.loadAssessmentModules(usageID);
                 view.load(usageID);
                 baseView.scrollToTop(".step--self-assessment");
```
cahrens: Remove this scroll?

openassessment/xblock/static/js/src/lms/oa_self.js
```diff
@@ -134,6 +148,7 @@ OpenAssessment.SelfView.prototype = {
         ).done(
             function() {
                 baseView.unsavedWarningEnabled(false, view.UNSAVED_WARNING_KEY);
+                view.announceStatusChange = true;
                 baseView.loadAssessmentModules(usageID);
                 view.load(usageID);
```
cahrens: Can this load be removed?

openassessment/xblock/static/js/src/lms/oa_self.js
cahrens: And announce errors in toggleActionError.

staubina:
Notes from conversation with @cahrens 

Convert 

```
                if (view.announceStatus){
                    // Update the Aria-Live region with the current status.
                            
                   view.baseView.announceStatusChangeToSR(view.baseView.getAssessmentStatus(stepID, usageID));
                    view.announceStatus = false;
                } else {
                    view.baseView.announceStatusChangeToSR();
                }
```

To a single method in the base view and remove the `view.announceStatus = true;` line and use the conditional to know when to announce status on a load.

staubina:
@cahrens I have made the changes that we discussed. I am going to convert this to a full PR once I get your feedback on the recent changes.

staubina:
@cahrens Please wait to review, I just remembered the error messaging changes.  Also want to resolve some test failures I missed.

karma.conf.js
```diff
@@ -26,6 +26,7 @@ module.exports = function(config) {
     // list of files / patterns to load in the browser
     files: [
       'lib/jquery.min.js',
+      'src/common/lib/xmodule/xmodule/js/common_static/js/src/accessibility_tools.js',
```
cahrens: This won't work-- this file lives in edx-platform. The lms and studio directories are subdirectories within ora itself.

The code either needs to be robust to that global function not existing, or we need to create our own sr div.

staubina:
@cahrens This is now ready for you to take another look.  The Sandbox is also updated.

openassessment/xblock/static/js/src/lms/oa_grade.js
```diff
@@ -20,18 +22,24 @@ OpenAssessment.GradeView.prototype = {
...
                 view.installHandlers();
-                if (typeof usageID !== 'undefined' &&
-                    !$(stepID, view.element).hasClass("is--unfinished") &&
-                    !$(stepID, view.element).hasClass("is--unstarted") &&
-                    !$(stepID, view.element).hasClass("is--waiting--staff")) {
-                    $("[id='oa_grade_" + usageID + "']", view.element).focus();
-                }
+
+                view.baseView.announceStatusChangeToSRandFocus(view.announceStatus,
+                    stepID,
```
cahrens: Nit: I hate taking up all this room, and would put more than one arg on a line to minimize the number of lines.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,108 @@ OpenAssessment.BaseView.prototype = {
     },
 
     /**
+     * Check if the Aria live region already exists, if it doesn't then add it.
```
cahrens: Nit: I'd reword this (the important part of the method is the reading of the text messages)--

Add the text messages to the Aria live region (first creating it if necessary).

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,108 @@ OpenAssessment.BaseView.prototype = {
...
+        if ($readerFeedbackSelector.length === 0) {
+            $(this.element).append(
+                '<div class="sr reader-feedback" aria-live="polite"></div>'
+            );
+            $readerFeedbackSelector = $(readerFeedbackID);
+        }
+        $.each(texts, function(ids, value) {
+            htmlFeedback = htmlFeedback + '<p>' + value + '</p>\n';
+        });
+        $readerFeedbackSelector.html(htmlFeedback);
```
cahrens: Do you have any unit test coverage that these messages are added to the sr div?

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,108 @@ OpenAssessment.BaseView.prototype = {
...
+
+    /**
+     * Checks the rendering status of the views that may require Screen Reader Status updates.
+     *
+     * The only views that should be added here are those that require Screen Reader updates when moving from one
+     * step to another.
+     *
+     * @returns {boolean} true if any step's view is still loading.
+     */
+    areSRStepsLoading: function() {
```
cahrens: This is a better name, and documentation is much better.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,108 @@ OpenAssessment.BaseView.prototype = {
...
+        return this.responseView.isRendering ||
+            this.peerView.isRendering ||
+            this.selfView.isRendering ||
+            this.gradeView.isRendering ||
+            this.trainingView.isRendering ||
+            this.staffView.isRendering;
+    },
+
+    /**
+     * Updates text int the Aria live region if all sections are rendered and focuses on the specified ID.
```
cahrens: Typo: "int"

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,108 @@ OpenAssessment.BaseView.prototype = {
...
+     *
+     * @param {boolean} announceCurrentStatus - true if the current status should be added to the aria-live region
+     *     even if it is now showing
```
cahrens: Do you mean "even if it is NOT showing"?

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,108 @@ OpenAssessment.BaseView.prototype = {
...
+     * @param {String} stepID - The id of the Step being worked on.
+     * @param {String} usageID  - The Usage id of the Step being worked on.
```
cahrens: This is the usageID of the xblock, no?

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,108 @@ OpenAssessment.BaseView.prototype = {
...
+     * @param {boolean} gradeStatus - true if this is a Grade status, false if it is an assessment status.
+     * @param {Object} currentView - Current active view.
+     * @param {String} focusID - The ID of the region to focus on.
+     */
+    announceStatusChangeToSRandFocus: function(announceCurrentStatus,
```
cahrens: Nit: don't like each of these on a separate line.

openassessment/xblock/static/js/src/lms/oa_base.js
cahrens: Maybe some documentation about the fact that status will be added to the aria-live region regardless of this value, if usageID and focusID are defined and the step is showing.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,108 @@ OpenAssessment.BaseView.prototype = {
...
+        } else if (announceCurrentStatus) {
+            this.srStatusUpdates.push(text);
+        }
+        if (!this.areSRStepsLoading() && this.srStatusUpdates.length > 0) {
+            this.srReadTexts(this.srStatusUpdates);
+            this.srStatusUpdates = [];
+        }
+    },
+
+    announceErrorMsgToSR: function(message) {
```
cahrens: This method is oddly named-- it has nothing to do with it being an "error message" per se. Since it is only used one place, I would just inline the code where it is called.

openassessment/xblock/static/js/src/lms/oa_grade.js
cahrens: Are you comfortable with announceStatusChangeToSRandFocus checking view.announceStatus directly and setting it to false at the end of the method (since view is already passed into the method)?

staubina:
@cptvitamin Can you take a look at this Sandbox and let me know if the changes are sufficient for TNL-5883 Sub Task https://openedx.atlassian.net/browse/TNL-6241?

efischer19:
All this code looks fine to me. I liked @cahrens's suggestion about a bok choy test addition, but I'll leave the final decision whether to add that up to you.

Consider this 👍'd from me, unless there are changes and you'd like me to re-review.

staubina:
@cahrens I made some minor updates to the bok choy tests.  @cptvitamin is going to take a look today if he has time, most likely tomorrow.

test/acceptance/tests.py
```diff
@@ -1100,7 +1109,6 @@ def test_train_self_peer_staff(self, peer_grades_me):
         # Do staff assessment step
         self.staff_assessment(peer_grades_me)
 
-
```
efischer19: This line shouldn't have been deleted - there ought to be 2 blank lines between classes.

cahrens:
👍

staubina:
@efischer19 I made the correction in cf9990f

staubina:
@cptvitamin Please use this Sandbox for testing.  The other one has died.

https://staubina-ora.sandbox.edx.org/

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -68,6 +69,95 @@ OpenAssessment.BaseView.prototype = {
...
+     * Add the text messages to the Aria live region (first creating it if necessary).
+     *
+     * @param {string[]} texts
+     */
+    srReadTexts: function(texts) {
+        var readerFeedbackID = '.sr.reader-feedback',
+            $readerFeedbackSelector = $(readerFeedbackID),
+            htmlFeedback = '';
+
+        if ($readerFeedbackSelector.length === 0) {
```
cahrens: Actually, instead of appending dynamically why can't this just always exist (via a template)? I'm concerned that text may not be read initially when it is dynamically injected-- @cptvitamin should know whether or not that is a problem.

But also for simplicity, it seems that it would be better if this region is just always there. I think you could put it in oa_base.html.

staubina:
@cahrens I have made the feedback region a static div in the template.  I am testing acceptance and a11y now.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -78,12 +78,6 @@ OpenAssessment.BaseView.prototype = {
             $readerFeedbackSelector = $(readerFeedbackID),
             htmlFeedback = '';
 
-        if ($readerFeedbackSelector.length === 0) {
```
cahrens: Update the doc as well.

staubina:
@cahrens Updated.

cahrens:
👍

cptvitamin:
👍  this rocks.  The feedback was timely and informative.  Loved using it with a screen reader.

PR #959, opened 2017-01-03
implement dateutils from edx-ui-toolkit

yro:
### Description
Render time zones via date-utils from edx-ui-toolkit. Utilizes a 'shimmed' version of requirejs for testing, but otherwise piggybacks the process from edx-platform.

### Testing
- [x] Unit tests

### Reviewers
- [x] Code review: @nasthagiri 

FYI: @efischer19 

### Post-review


.gitignore
```diff
@@ -71,3 +71,5 @@ test/selenium/screenshots/*
 
 # logging
 logs/*.log*
+
+edx-ora2/
```
yro: This was to eliminate the build from my devstack -- will remove before final merge.

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
```diff
@@ -30,18 +30,20 @@ <h5 class="message__title">{% trans "Status" %}</h5>
 
                 <div class="message__content">
                     {% trans "Your submission was cancelled. " %}
-                    <p>
+                    <p class="xnxx_placeholder">
                         {% if workflow_cancellation.cancelled_by %}
```
yro: The UX for this has always shown a datestamp and UTC. I think, as cancellation is rather serious (and rare), it should remain as so. This, I believe, emphasizes the sense of an 'official action' to the user, which is the only way to generate this page.

karma.conf.js
```diff
@@ -4,7 +4,7 @@ module.exports = function(config) {
   config.set({
 
     // base path that will be used to resolve all patterns (eg. files, exclude)
-    basePath: 'openassessment/xblock/static/js',
+    basePath: 'openassessment/xblock/static/',
```
yro: This can probably be changed back.

openassessment/assessment/tasks.py
```diff
@@ -4,4 +4,4 @@
 """
 # pylint:disable=W0611
 from .worker.training import train_classifiers, reschedule_training_tasks
-from .worker.grading import grade_essay, reschedule_grading_tasks
\ No newline at end of file
+from .worker.grading import grade_essay, reschedule_grading_tasks
```
yro: Simple linter change

yro:
Setting up sandbox:
https://yro.sandbox.edx.org

yro:
https://yro.sandbox.edx.org/courses/course-v1:edx+ORA203+course/courseware/575dacda404e4f94ae4b666a1c00ca7c/8ad4e160407b4707a238106afadb2749/ as 'audit' user for demo.

openassessment/templates/openassessmentblock/peer/oa_peer_assessment.html
```diff
@@ -1,6 +1,7 @@
 {% load tz %}
 {% load i18n %}
 {% block list_item %}
+{% load staticfiles %}
 <li
```
yro: This staticfile load is a remnant from a previous iteration (dateutils in edx-platform uses a static file loader). Will delete.

openassessment/xblock/peer_assessment_mixin.py
```diff
@@ -1,3 +1,8 @@
+"""
+The Peer Assessment Mixin for all Peer Functionality.
+
+"""
+
 import logging
```
yro: Linter change.

openassessment/xblock/peer_assessment_mixin.py
```diff
@@ -161,12 +166,13 @@ def peer_path_and_context(self, continue_grading):
         path = 'openassessmentblock/peer/oa_peer_unavailable.html'
         finished = False
         problem_closed, reason, start_date, due_date = self.is_closed(step="peer-assessment")
-        user_service = self.runtime.service(self, 'user')
+        user_preferences = get_user_preferences(self.runtime.service(self, 'user'))
```
yro: Refactored the user timezone service to also retrieve the user_language (needed for time localization).

openassessment/xblock/resolve_dates.py
```diff
@@ -24,16 +24,25 @@ class DateValidationError(Exception):
...
+    }
+    retrieved_preferences = user_service.get_current_user().opt_attrs.get('edx-platform.user_preferences')
+
+    if retrieved_preferences is not None:
+        if retrieved_preferences.get('time_zone'):
+            user_preferences['user_timezone'] = retrieved_preferences.get('time_zone')
+        if retrieved_preferences.get('pref-lang'):
+            user_preferences['user_language'] = retrieved_preferences.get('pref-lang')
+
+    return user_preferences
```
yro: Here is the refactored time_zone retrieval method

openassessment/xblock/static/js/fixtures/templates.json
```diff
@@ -1195,5 +1196,22 @@
...
+                        { "text": "", "prompt": { "description": "Prompt 2" }}
+                    ]
+                }
+            },
+            "save_status": "This response has not been saved.",
+            "submit_enabled": true,
+            "submission_due": "2020-01-01T00:00"
+        },
+        "output": "oa_response_date.html"
     }
```
yro: This renders a template for the jasmine test.

openassessment/xblock/static/js/spec/lms/oa_datefactory.js
```diff
@@ -0,0 +1,84 @@
...
+            var testDict = {
+                'Invalid date': false,
+                'invalid date': true,
+                '': false,
+                1: true
+            };
+            Object.keys(testDict).forEach(function(key) {
+                expect(datetimeFactory.isValid(key)).toEqual(testDict[key]);
+            });
+        })
```
yro: Due to the async nature of require, the test had to be structured (as much as possible) to test the logic of the datetime factory, and not the actual calls to moment and moment-timezone. There is a 'load every script' call in the jasmine framework for ora, so one can be reasonably confident that the libraries are loading and that the factory will work.

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -26,6 +26,9 @@ OpenAssessment.ResponseView = function(element, server, fileUploader, baseView,
     this.fileUploaded = false;
     this.announceStatus = false;
     this.isRendering = false;
+    this.dateFactory = new OpenAssessment.DateTimeFactory(this.element);
+    this.announceStatus = false;
+    this.isRendering = false;
 };
```
yro: lines 30 and 31 were to resolve a merge conflict.

openassessment/xblock/static/js/src/lms/oa_peer.js
```diff
@@ -16,6 +16,10 @@ OpenAssessment.PeerView = function(element, server, baseView) {
     this.rubric = null;
     this.isRendering = false;
     this.announceStatus = false;
+    this.dateFactory = new OpenAssessment.DateTimeFactory(this.element);
+    this.isRendering = false;
+    this.announceStatus = false;
+
```
yro: lines 20 and 21 were to resolve a merge conflict.

openassessment/xblock/static/js/src/lms/oa_datefactory.js
```diff
@@ -0,0 +1,90 @@
...
+            );
+        }).call(this, require || RequireJS.require);
+    },
+
+    isValid: function(candidateVariable) {
+        return candidateVariable !== undefined &&
+            candidateVariable !== '' &&
+            candidateVariable !== 'Invalid date' &&
+            candidateVariable !== 'None';
+    }
```
yro: This was structured to separate out the logic and the library (e.g. DateUtils.localize(context);) calls as much as possible for js testing. The library calls would not work in a straightforward manner due to async calls to moment, etc., so this approach was adopted.

openassessment/xblock/static/js/src/lms/oa_datefactory.js
```diff
@@ -0,0 +1,90 @@
...
+        var opFactory = this;
+        var dateToken = 'date';
+        if (opFactory.isValid(el.data('datetoken'))) {
+            dateToken = el.data('datetoken');
+        }
+        return dateToken;
+    },
+
+    generate: function(el) {
+        var opFactory = this;
```
yro: This is a terrible method name. Please help.

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -58,11 +61,10 @@ OpenAssessment.ResponseView.prototype = {
...
                 view.installHandlers();
                 view.setAutoSaveEnabled(true);
-
                 view.isRendering = false;
                 view.baseView.announceStatusChangeToSRandFocus(stepID, usageID, false, view, focusID);
                 view.announceStatus = false;
-
+                view.dateFactory.report();
             }
         ).fail(function() {
```
yro: This call is a bit redundant, but there's no harm done in calling the factory multiple times. It seemed, observationally, to increase the reliability of the functions.

openassessment/xblock/static/js/src/lms/oa_self.js
```diff
@@ -16,6 +16,9 @@ OpenAssessment.SelfView = function(element, server, baseView) {
     this.rubric = null;
     this.isRendering = false;
     this.announceStatus = false;
+    this.dateFactory = new OpenAssessment.DateTimeFactory(this.element);
+    this.isRendering = false;
+    this.announceStatus = false;
 };
```
yro: 20 and 21 to resolve merge conflicts

openassessment/xblock/test/test_openassessment.py
```diff
@@ -159,67 +159,40 @@ def _render_xblock(self, xblock):
...
-        with patch(self.TIME_ZONE_FN_PATH) as time_zone_fn:
-            time_zone_fn.return_value = pytz.timezone('America/Los_Angeles')
-
-            # Set start dates'
-            xblock = self._set_up_start_date(date)
-            resp = self._render_xblock(xblock)
-            self.assertIn(expected_start_date, resp.body)
-
-    @ddt.data(('utc', 'May 1, 2014 00:00 UTC'),
-              ('America/Los_Angeles', 'April 30, 2014 17:00 PDT'))
```
yro: DST is now tested in the dateutils lib.

require-config.js
```diff
@@ -0,0 +1,63 @@
...
+            'edx-ui-toolkit/js/utils/string-utils': './node_modules/edx-ui-toolkit/src/js/utils/string-utils'
+        },
+        shim: {
+            'jquery': {
+                exports: 'jQuery'
+            }
+        }
+
+    })
+}).call(this, require || RequireJS.require, define || RequireJS.define);
```
yro: This is a require-config file, but effectively 'fakes' require for the jasmine tests to pass and is not called by the methods in production. 
From my testing it seemed like it needed to live in a more centralized dir than where I initially put it (openassessment/xblock/static/js/lib/), as it needs to determine the directory structure outside of the normal paths that a standard require config would use. (at least in examples I could find)

openassessment/xblock/resolve_dates.py
```diff
@@ -24,16 +24,25 @@ class DateValidationError(Exception):
 DISTANT_FUTURE = dt.datetime(dt.MAXYEAR, 1, 1, tzinfo=pytz.utc)
 
 
-def get_current_time_zone(user_service):
+def get_user_preferences(user_service):
```
nasthagiri: Nit: Should this be moved to another file since it's no longer date specific?

openassessment/xblock/resolve_dates.py
```diff
@@ -24,16 +24,25 @@ class DateValidationError(Exception):
...
+        'user_language': None,
+        'user_timezone': None
+    }
+    retrieved_preferences = user_service.get_current_user().opt_attrs.get('edx-platform.user_preferences')
+
+    if retrieved_preferences is not None:
+        if retrieved_preferences.get('time_zone'):
+            user_preferences['user_timezone'] = retrieved_preferences.get('time_zone')
+        if retrieved_preferences.get('pref-lang'):
+            user_preferences['user_language'] = retrieved_preferences.get('pref-lang')
```
nasthagiri: Can simplify this by just removing the `if` statements.

yro:
Acceptance tests passing:
http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/1234/console

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
```diff
@@ -26,11 +26,11 @@ <h2 class="staff-info__title ui-staff__subtitle">
...
                 <p>
                     {% if workflow_cancellation.cancelled_by %}
-                        {% blocktrans with removed_by_username=workflow_cancellation.cancelled_by removed_datetime=workflow_cancellation.cancelled_at|timezone:time_zone|date:"F j, Y H:i e" %}
+                        {% blocktrans with removed_by_username=workflow_cancellation.cancelled_by removed_datetime=workflow_cancellation.cancelled_at|date:"Y-m-d H:i" %}
                             Learner submission removed by {{ removed_by_username }} on {{ removed_datetime }}
                         {% endblocktrans %}
                     {% else %}
-                        {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:time_zone|date:"F j, Y H:i e" %}
+                        {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|date:"Y-m-d H:i" %}
                             Learner submission removed on {{ removed_datetime }}
```
nasthagiri: Do these 2 datetimes need to be localized?

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
```diff
@@ -32,12 +32,12 @@ <h5 class="message__title">{% trans "Status" %}</h5>
...
-                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:time_zone|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
-                                Your submission has been cancelled by {{ removed_by_username }} on {{ removed_datetime }}
+                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:"UTC"|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
+ -                                Your submission has been cancelled by {{ removed_by_username }} on {{ removed_datetime }}
                             {% endblocktrans %}
                         {% else %}
-                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:time_zone|date:"N j, Y H:i e" %}
-                                Your submission was cancelled on {{ removed_datetime }}
+                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:"UTC"|date:"N j, Y H:i e" %}
+ -                              Your submission was cancelled on {{ removed_datetime }}
```
nasthagiri: Where is `removed_datetime` localized?

openassessment/xblock/staff_area_mixin.py
```diff
@@ -319,11 +319,14 @@ def get_student_submission_context(self, student_username, submission):
...
             A context dict for rendering a student submission and associated rubric (for staff grading).
         """
+        user_preferences = get_user_preferences(self.runtime.service(self, 'user'))
+
         context = {
             'submission': create_submission_dict(submission, self.prompts) if submission else None,
             'rubric_criteria': copy.deepcopy(self.rubric_criteria_with_labels),
             'student_username': student_username,
-            'time_zone': get_current_time_zone(self.runtime.service(self, 'user')),  # localize for staff user
+            'user_timezone': user_preferences['user_timezone'],  # localize for staff user
```
nasthagiri: The `localize for staff user` comment makes more sense to stay with the call to `get_user_preferences` (now moved above).  Since the comment refers to _which_ user's preferences are used.

openassessment/xblock/static/js/src/lms/oa_peer.js
nasthagiri: L20-21 are a repeat of L17-18.

(Same goes for the other files with the apparent "merge conflict" issues.)

openassessment/xblock/static/js/src/lms/oa_datefactory.js
nasthagiri: Can you explain this further?  Are you saying this file is needed only for test code?  Or, also for production code?
I'm not too happy with duplicating a lot of this logic that is already handled and implemented in `DateUtils`.

.gitignore
yro: @nasthagiri it looks like the 'make install' command from the wiki here: https://openedx.atlassian.net/wiki/display/TNL/ORA+FAQ#ORAFAQ-InitialSetup
creates this dir. I got Christina to thumb this change on https://github.com/edx/edx-ora2/pull/966

openassessment/xblock/static/js/src/lms/oa_peer.js
yro: 💯

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
```diff
@@ -32,12 +32,12 @@ <h5 class="message__title">{% trans "Status" %}</h5>
                     {% trans "Your submission was cancelled. " %}
                     <p>
                         {% if workflow_cancellation.cancelled_by %}
-                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:time_zone|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
-                                Your submission has been cancelled by {{ removed_by_username }} on {{ removed_datetime }}
+                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:"UTC"|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
+ -                                Your submission has been cancelled by {{ removed_by_username }} on {{ removed_datetime }}
```
yro: 👍

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
yro: It isn't -- previously it was a simple datestamp style string (e.g. '2014-12-25 14:46 UTC') - and I was unsure about making a UX change. I will follow up with product about this.

openassessment/templates/openassessmentblock/self/oa_self_assessment.html
```diff
@@ -18,12 +18,16 @@ <h4 class="step__title">
...
+                            <span class="date ora-datetime" data-datetime="{{ start_date }}" data-string="available {date} (in {{ time_until }})" data-timezone="{{ user_timezone }}" data-language="{{ user_language }}"></span>
+                        {% endblocktrans %}
                     </span>
                     {% elif self_due %}
                     <span class="step__deadline">
                         {# Translators: This string displays a date to the user, then tells them the time until that date.  Example: "due August 13th, 2014 (in 5 days and 45 minutes)" #}
-                        {% blocktrans with due_date=self_due|timezone:time_zone|date:"N j, Y H:i e" time_until=self_due|timeuntil %}due <span class="date">{{ due_date }} (in {{ time_until }})</span>{% endblocktrans %}
+                        {% blocktrans with due_date=self_due|date:"Y-m-d H:i" time_until=self_due|timeuntil %}
+                        <span class="date ora-datetime" data-datetime="{{ due_date }}" data-string="due {date} (in {{ time_until }})" data-timezone="{{ user_timezone }}" data-language="{{ user_language }}"></span>
```
yro: 👍

openassessment/templates/openassessmentblock/staff_area/oa_student_info.html
yro: This is similar to the cancelled submission template above in the UX. Will follow up.

openassessment/xblock/resolve_dates.py
yro: @nasthagiri That's not a bad call. There isn't a strong candidate in the dir right now, but I could make it live alone in a 'user_info.py' file or something similar.

openassessment/xblock/resolve_dates.py
yro: 👍  of course!

openassessment/xblock/staff_area_mixin.py
yro: That's solid. 👍

openassessment/xblock/static/js/src/lms/oa_datefactory.js
yro: This is duplicative of code that lives in edx-platform (edx-platform/lms/static/js/dateutil_factory.js), but not of logic within the dateutils proper. The way this is structured here, it *should* be able to stand on its own. It's possible to piggyback the dateutil_factory from edx-platform, but it seemed a little unstable, and less 'self-contained'.

openassessment/xblock/static/js/src/lms/oa_response.js
nasthagiri: These calls seem symptomatic of an underlying issue.  Is there no common place where we can activate the dateFactory code?

yro: I agree it's a little bit redundant. I attempted to get it to fire using oa_base.js (the 'base view' renderer) and couldn't get it to line up. The utils need to fire after rendering, and I was only able to get it to work in the individual views. The redundancy is to insure against instances where course staff has only chosen to render one step of the possible assessment steps (e.g. staff assessment only). This assures that the datefactory fires with the correctly associated element at all times.

openassessment/xblock/static/js/spec/lms/oa_datefactory.js
```diff
@@ -0,0 +1,84 @@
...
+**/
+
+describe('OpenAssessment.DateTimeFactory', function() {
+
+    beforeEach(function() {
+        // Load the DOM fixture
+        loadFixtures('oa_response_date.html');
+
+    });
+    describe('transform', function() {
```
nasthagiri: s/transform/report

openassessment/xblock/static/js/src/lms/oa_datefactory.js
```diff
@@ -0,0 +1,90 @@
...
+ * A helper function to utilize DateUtils.
+ **/
+
+OpenAssessment.DateTimeFactory = function(element) {
+    this.element = element;
+};
+
+OpenAssessment.DateTimeFactory.prototype = {
+
+    report: function() {
```
nasthagiri: Is there a better name for `report`?  Since this isn't really creating a report.

How about renaming `report` to `apply` (or `apply_to_doc`) and `generate` to `apply_to_element`?   Since the former affects the entire document, while the latter affects a single element.

openassessment/templates/openassessmentblock/response/oa_response_cancelled.html
```diff
@@ -32,12 +32,12 @@ <h5 class="message__title">{% trans "Status" %}</h5>
...
                         {% if workflow_cancellation.cancelled_by %}
-                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:time_zone|date:"N j, Y H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
+                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:"UTC"|date:"Y-m-d H:i e" removed_by_username=workflow_cancellation.cancelled_by %}
                                 Your submission has been cancelled by {{ removed_by_username }} on {{ removed_datetime }}
                             {% endblocktrans %}
                         {% else %}
-                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:time_zone|date:"N j, Y H:i e" %}
-                                Your submission was cancelled on {{ removed_datetime }}
+                            {% blocktrans with removed_datetime=workflow_cancellation.cancelled_at|timezone:"UTC"|date:"Y-m-d H:i e" %}
+                               Your submission was cancelled on {{ removed_datetime }}
```
nasthagiri: Is there a ticket for the follow-on work to fix the `removed_datetime` values (and whatever else remains)?  Until then, those dates will continue to render in UTC.

yro: @nasthagiri I consulted with @lizcohen and the feeling was that as this was a communiqué and not a deadline, it should remain UTC.

openassessment/xblock/static/js/spec/lms/oa_datefactory.js
```diff
@@ -0,0 +1,84 @@
...
+        it('has the correct HTML elements', function() {
+            var timeElement = $('.step__title').get(0);
+            var datetimeFactory = new OpenAssessment.DateTimeFactory(timeElement);
+            datetimeFactory.report();
+            $('.ora-datetime', timeElement).each(function() {
```
andy-armstrong: I'm nervous about this as ``report`` is asynchronous because it invokes ``generate`` which in turn uses ``require``. It seems that there might be a timing window here where the factory hasn't completed before these assertions are run.

Thinking it through, I think the three dependencies (jquery, string-utils, and date-utils) should all be pre-loaded both in platform and in the tests, so maybe in that case the require call is synchronous.

openassessment/xblock/test/test_openassessment.py
```diff
@@ -100,8 +100,8 @@ def test_student_view_workflow_error(self, xblock):
         # Expect that the page renders even if the update fails
         self.assertIn("OpenAssessmentBlock", xblock_fragment.body_html())
 
-    @ddt.data(('utc', 'April 1, 2014 00:00 UTC'),
-              ('America/Los_Angeles', 'March 31, 2014 17:00 PDT'))
+    @ddt.data(('utc', '2014-03-31 20:00'),
```
andy-armstrong: Why do the dates now all render as ``2014-03-31`` instead of ``March 31, 2014``. Is this change going to be visible to the user, and if so is this the way product expects the dates to be rendered? I think the latter is much more user friendly so I'd rather see that as the default.

openassessment/xblock/static/js/src/lms/oa_datefactory.js
```diff
@@ -0,0 +1,90 @@
...
+        if (opFactory.isValid(el.data('datetoken'))) {
+            dateToken = el.data('datetoken');
+        }
+        return dateToken;
+    },
+
+    generate: function(el) {
+        var opFactory = this;
+        (function(require) {
+            require([
```
andy-armstrong: It is unfortunate to have the only usage of ``require`` be hidden inside a function like this, but in investigating this with you we saw that it wasn't possible to get ``DateUtils`` to load without RequireJS. I was concerned about bundling, but it doesn't matter since ORA does its own bundling, and the three dependencies will all be pre-loaded in the platform.

openassessment/xblock/static/js/spec/lms/oa_datefactory.js
yro: @andy-armstrong I absolutely agree with you. In practice, the platform has loaded the dependencies well before this is called, however if we decide in the future to deepen the require dependency in edx-ora, there might be some refactoring that can be done.
This seemed to be the best solution in this interim state, and if we do move forward in the future, I'm confident that it'll continue to work as is.

openassessment/xblock/test/test_openassessment.py
yro: The template-rendered dates now are contained within data elements in the html tags. No user-visible date format changes have occurred (indeed, I did a PR against edx-ui-toolkit to add a format constant to assure this would be the case).
This tests only for the data elements.

openassessment/xblock/static/js/src/lms/oa_datefactory.js
yro: 👍

openassessment/xblock/static/js/src/lms/oa_datefactory.js
```diff
@@ -0,0 +1,90 @@
...
+ **/
+
+OpenAssessment.DateTimeFactory = function(element) {
+    this.element = element;
+};
+
+OpenAssessment.DateTimeFactory.prototype = {
+
+    report: function() {
+        var opFactory = this;
```
yro: Absolutely. 👍

openassessment/xblock/static/js/spec/lms/oa_datefactory.js
yro: 👍

nasthagiri:
👍 remember to squash before merging.

PR #961, opened 2017-01-11
Aj/tnl 6242 announce save state

staubina:
### Description

When a user saves their response the aria live region will now provide feedback.

This is the next task in the Story https://openedx.atlassian.net/browse/TNL-5883

This will not be merged until the aj/TNL-5883_announce_state branch has been merged.

### Sandbox
https://staubina-ora2.sandbox.edx.org

### Testing

### Reviewers
- [x] Code review: @cahrens 
- [x] Accessibility review: @cptvitamin 

FYI Product review: @sstack22 
@jlajoie 

### Post-review


staubina:
@cahrens @efischer19 @cptvitamin Please review this PR when you have a chance.

cahrens:
@staubina although the code looks good, when I tried this with VoiceOver, 3 out of 4 times I tried it I didn't hear the sr text.

staubina:
> @staubina although the code looks good, when I tried this with VoiceOver, 3 out of 4 times I tried it I didn't hear the sr text.

I will take another look at this.  I might need to clear the SR region before adding.  Also, there is a lot of chatter on that region.

> Finally, I wonder if the issue is related to the dynamic injection of the sr region. If an sr region is injected with text already in it, I doubt it will be read. I know that this is part of the previous PR, but why not just have the sr-region always present (via the template)?

I followed the example of what was done with the other SR region.  This can be changed, I did not see an issue with either method.

staubina:
@cahrens I have made the changes in #958 and this branch is now rebased with those changes.  I am not seeing the same issue you are with the live region failing to update.  Let me know if the issue persists and I will talk with @cptvitamin .

staubina:
Sandbox updated.

staubina:
@cahrens  I have made the requested  change to the status SR text

staubina:
>If you go this route, you should also call srReadTexts in the fail clause below so that errors are read. 

```
view.baseView.toggleActionError('save', errMsg);
```

The message is already added in the toggleActionError method.

cahrens:
@staubina thanks for the follow up-- I was thinking there was a case where an error message would be shown in that same area, but apparently I was wrong!

cahrens:
👍

cptvitamin:
👍  this rocks.  The feedback was timely and informative.  Loved using it with a screen reader.

staubina:
@cahrens @efischer19 

During the squash process I somehow managed to delete/overwrite my changes.  Please take a look at this pr again.  The functionality is back to what it was before when you reviewed.  I am also updating the Sandbox.

staubina:
The code changes should also be very similar to before.  The Sandbox is updated.

openassessment/xblock/static/js/src/lms/oa_base.js
```diff
@@ -69,6 +69,13 @@ OpenAssessment.BaseView.prototype = {
     },
 
     /**
+     * Clear the text in the Aria live region.
+     */
+    srClear: function() {
+        $('.sr.reader-feedback').html('');
```
cahrens: I would define .sr.reader-feedback as a class variable, since you use it in a couple of methods.

cahrens:
👍

PR #972, opened 2017-02-02
Loosen celery version pin

jcdyer:
Spike to see about upgrading celery for edx-platform


efischer19:
The celery requirement was removed, and all dependencies loosened in https://github.com/edx/edx-ora2/pull/1023

PR #976, opened 2017-02-14
Fix broken acceptance tests

efischer19:
A recent change to edx-platform (see [the ECOM ticket](https://openedx.atlassian.net/browse/ECOM-7041) for background, [the git commit](https://github.com/edx/edx-platform/commit/d141eb15b1687a1c493666c7cbec566b4f7c5c91) for the actual change) has revealed a bug in our testing code.

Basically, the method `get_username_and_email` had a poorly-written regex that was actually extracting the username and *password* from the auto auth page. In all tests that were using the `FullWorkflowMixin` class, we'd try to "log back in" as a previous user with the correct username and this buggy "email". [A quick peek at the auto auth code](https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/views.py#L2079-L2094) will show that the username is used to lookup the user then the new password is updated, so our tests were still functionally correct with the added side-effect of updating a bunch of emails to be identical.

The fix is to just make the regex do what the method name is claiming it does.

## tl;dr:
- recent edx-platform change forced unique email for all users
- buggy regex in edx-ora2 auto_auth logic had been updating multiple test users' emails to be `test_password`

FYI @yro @cahrens

### DIFF
--- a/test/acceptance/auto_auth.py
+++ b/test/acceptance/auto_auth.py
@@ -85,7 +85,7 @@ class AutoAuthPage(PageObject):
         Finds and returns the username and email address of the current user.
         """
         message = self.q(css='BODY').text[0].strip()
-        match = re.search(r'Logged in user (\S+) \(.*\) with password (\S+)', message)
+        match = re.search(r'Logged in user (\S+) \((.*)\).*', message)
         if not match:
             return None
         return match.group(1), match.group(2)


scripts/test-acceptance.sh
```diff
@@ -53,4 +53,4 @@ if [[ "${test_name}" = "accessibility" ]]; then
 fi
 
 echo "Running acceptance tests from ${test_name}.py against the sandbox..."
-nosetests ../acceptance/${test_name}.py --with-xunit --xunit-file=../acceptance/xunit-${test_name}.xml
+nosetests ../acceptance/${test_name}.py:FullWorkflowRequiredTest --with-xunit --xunit-file=../acceptance/xunit-${test_name}.xml
```
efischer19: This is just to confirm the buggy tests pass on Jenkins without needing to wait on a full test run; will not be included in final merge.

Test plan:
- this subset of tests passes
- all tests pass
- PR reviewed and merged
- @yro (or I, whoever's free at that time) can continue with the ongoing release effort

cahrens:
👍 Thanks, Eric!!

yro:
👍

efischer19:
[The tests that were failing passed on Jenkins](http://jenkins.edx.org:8080/view/ora2/job/ora2-acceptance-tests/1269/), although the run didn't complete on account of the daily "rebuild the sandbox" job starting and deactivating the sandbox out from under us.

I'm going to merge this and restart the rebuild job in order to get these changes onto the daily acceptance test run.

PR #977, opened 2017-02-16
Update README dev instructions

efischer19:
Removing the outdated instructions that bit @jzoldak yesterday for clarity when new or unfamiliar devs come into edx-ora2.

I tried to remove everything that is on our FAQ page so we have a single source of truth and don't have to do this again in 6 months. LMK if you have any suggestions.

I've also copied @nedbat's [commit](https://github.com/edx/edx-platform/commit/7523fb2e07661800d5babadf28b946f1d0a50309) updating IRC references in the edx-platform README to refer to Slack instead.

2 of @cahrens @staubina @yro to review at their convenience.


efischer19:
I've been looking at https://github.com/edx/edx-ora2/tree/efischer/update_readme to preview this.

README.rst
```diff
@@ -4,141 +4,10 @@ Open Response Assessment |build-status| |coverage-status|
...
-
-To pull strings from Transifex
-
-.. code:: bash
-
-    ./scripts/i18n-pull.sh
+Installation, Tests, and other Developer Tasks
+==============================================
 
+The previous README instructions have fallen woefully out-of-date, edX engineers now follow the guides on our `internal wiki <https://openedx.atlassian.net/wiki/display/TNL/ORA+FAQ>`_.
```
staubina: Do we want the instructions in the internal Wiki or here?  It seems more logical to have them here.

cahrens: I don't have a problem linking to the wiki, though I would not refer to "internal wiki" since it is public. I find it easier to update the wiki (and more likely that people will actually do so) then to require all changes to go through a PR.

cahrens: There's no reason to mention a "previous README" since the content is being deleted. Just point to the wiki page.

README.rst
```diff
@@ -4,141 +4,10 @@ Open Response Assessment |build-status| |coverage-status|
...
-
-    mkdir orastack
-    cd orastack
-    git clone git@github.com:edx/edx-ora2.git
-
-Create the Vagrant instance:
-
-.. code:: bash
-
-    ln -s ./edx-ora2/Vagrantfile ./
```
cahrens: Should we also delete the vagrant file?

README.rst
```diff
@@ -158,12 +27,14 @@ Reporting Security Issues
...
+======================
+
+You can discuss this code on the `edx-code Google Group`. Alternatively, our real-time conversations are on Slack. You can request a `Slack invitation`_, then join our `community Slack team`_.
 
-You can discuss this code on the
-`edx-code Google Group <https://groups.google.com/forum/#!forum/edx-code>`_ or
-in the `edx-code` IRC channel on Freenode.
+.. _edx-code Google Group: https://groups.google.com/forum/#!forum/edx-code
+.. _Slack invitation: https://openedx-slack-invite.herokuapp.com/
+.. _community Slack team: http://openedx.slack.com/
```
nedbat: Although this is a copy of a change I made elsewhere, these days I think it would be better to point to open.edx.org/getting-help.  Let's do this:
```
You can get help with this code on our `mailing lists`_ or in real-time conversations on `Slack`_.

.. _mailing lists: https://open.edx.org/getting-help
.. _Slack: https://open.edx.org/getting-help
```
(Yes, it's a little weird to link them both to the same place, but I think it's a good compromise between having a content-less link to the page, and too many details here.)

efischer19:
@staubina's comment got squashed in my last commit, but the discussion still ought to be had - do we want the instructions listed here or on our wiki?

IMO, the most important thing is to only have 1 canonical set of instructions, wherever they may be located. In defense of the wiki (how I'm leaning, but open to debate), edits are easier/faster to make, current engineers expect to look there for help, and the github README page is kept very minimal (note that the user docs also link to an outside page rather than spelling everything out here). Other thoughts?

staubina:
@efischer19 I am good with either and confluence does have more editing power.  I think I was also confused by the use of "internal".  Can we remove that part as @cahrens mentioned?

efischer19:
Oh yeah, that's a solid idea @staubina, thanks!

cahrens:
👍 Thanks, Eric!

README.rst
```diff
@@ -4,141 +4,10 @@ Open Response Assessment |build-status| |coverage-status|
...
-
-To pull strings from Transifex
-
-.. code:: bash
-
-    ./scripts/i18n-pull.sh
+Installation, Tests, and other Developer Tasks
+==============================================
 
+EdX engineers now follow the `guides on our wiki <https://openedx.atlassian.net/wiki/display/TNL/ORA+FAQ>`_.
```
jzoldak: nit: Remove the "now"

cahrens: If we remove "now" (which I agree with, but decided not to comment on before), then also remove "EdX engineers". How about just "Follow the ..."?

PR #980, opened 2017-02-23
Criterion Names are poorly defined

efischer19:
This block of code does NOT do what you would think, as I recently learned while investigating a doc question for @catong.

As it turns out, the value you enter in studio as the name of the criterion is a "label", while the "name" is just a list index. I think we should remove this validation, as it causes confusion and isn't actually doing anything useful.

For example, this rubric passes validation despite having duplicate names (from the Studio user's point of view):
screen shot 2017-02-23 at 12 00 15 pm

In a debugger, I see the following as the rubric definition. I think the data we're getting is coming from https://github.com/edx/edx-ora2/blob/master/openassessment/templates/openassessmentblock/edit/oa_edit_criterion.html.
```
(Pdb) import json
(Pdb) print json.dumps(rubric_dict, indent=2)
{
  "prompts": [
    {
      "description": "    Censorship in the Libraries\n\n    'All of us can think of a book that we hope none of our children or any other children have taken off the shelf. But if I have the right to remove that book from the shelf -- that work I abhor -- then you also have exactly the same right and so does everyone else. And then we have no books left on the shelf for any of us.' --Katherine Paterson, Author\n\n    Write a persuasive essay to a newspaper reflecting your views on censorship in libraries. Do you believe that certain materials, such as books, music, movies, magazines, etc., should be removed from the shelves if they are found offensive? Support your position with convincing arguments from your own experience, observations, and/or reading.\n\n    Read for conciseness, clarity of thought, and form.\n"
    }
  ], 
  "criteria": [
    {
      "prompt": "a", 
      "feedback": "required", 
      "label": "a", 
      "order_num": 0, 
      "options": [], 
      "name": "0"
    }, 
    {
      "prompt": "a", 
      "feedback": "required", 
      "label": "a", 
      "order_num": 1, 
      "options": [], 
      "name": "1"
    }
  ]
}
```

@cahrens Can you look over this as a sanity check when you get a chance? No rush.


cahrens:
👍

efischer19:
FYI I'm going to add the requirements update and version bump commits on this PR, then merge and release.

PR #984, opened 2017-03-14
Ignore reset signal for staff grades

efischer19:
Just an idea I had of how we can clean up the ORA/submissions and Robust Grades interaction, wanted to get it down before I forgot. Let's discuss further tomorrow @nasthagiri.


efischer19:
pairs with https://github.com/edx/edx-submissions/pull/54

PR #988, opened 2017-03-22
proof-of-concept

efischer19:
FYI @staubina 

Throwing this up on a branch to save it from the purging fires that are about to claim my devstack.


requirements/base.txt
```diff
@@ -22,10 +18,14 @@ jsonfield==1.0.3
...
 libsass==0.10.0
 loremipsum==1.0.5
+lxml==3.7.3
+numpy==1.12.1
 path.py==8.2.1
 python-dateutil==2.1
 python-memcached==1.48
 pytz
+scikit-learn==0.18.1
+scipy==0.19.0
```
efischer19: This is the results of:

1.  install each entry previously in `wheels.txt` without a version specifier
2. see which version was installed, and pin that here

I had to do this as some of the previously checked-in wheels were for versions that are no longer available via `pip`.

I think quite a few of these are only needed by AI grading, we ought to excise that unused feature from ORA entirely at some point.

efischer19:
I was able to get my devstack working using these changes and [my scripts](https://gist.github.com/efischer19/e5509f94a6fde57ea8be90de32a1296a)

efischer19:
Closing in favor of https://github.com/edx/edx-ora2/pull/989, which handles wheel installation much more cleanly

PR #990, opened 2017-03-23
update edx-submissions requirement

efischer19:
Follow up to https://github.com/edx/edx-submissions/pull/55, to ensure this update gets into edx-platform with the next ORA release


efischer19:
Merging this using @yro's previous edx-submissions thumb

PR #991, opened 2017-03-23
travis tests

efischer19:
FYI @staubina 

I'm just testing a few changes to see how we can get our travis tests run more quickly, using my own PR so as not to disturb the primary. Will not merge here.


staubina:
Change the base to my branch for the PR and if it works we can merge it in before merging mine to master.

efischer19:
So, the takeaways here as I see them are:

- Using `--only-binary` results in much faster travis builds. Compare https://travis-ci.org/edx/edx-ora2/builds/214332881 (this PR, 292 seconds for `make install`) with https://travis-ci.org/edx/edx-ora2/builds/214270438 (@staubina's PR, 441 seconds for `make install`
- Moving to `--only-binary` means updating the version of the requirements in `wheels.txt`, as I other wise ran into this:
```
(edx-ora2) edxapp@vagrant:/edx/src/edx-ora2$ pip install --only-binary=:all: -r requirements/wheels.txt
Collecting lxml==3.4.4 (from -r requirements/wheels.txt (line 5))
  Could not find a version that satisfies the requirement lxml==3.4.4 (from -r requirements/wheels.txt (line 5)) (from versions: 3.6.3, 3.6.4, 3.7.1, 3.7.2, 3.7.3)
No matching distribution found for lxml==3.4.4 (from -r requirements/wheels.txt (line 5))
```
- Updating the version of those files causes AI-grading related tests to fail (see https://travis-ci.org/edx/edx-ora2/builds/214332881 again), so they probably break AI grading in general.
  - Also, updating `numpy` from 1.6.2 is scary, and would necessitate a larger investigation given that it's used at 1.6.2 in edx-platform currently.

I'm going to add my Makefile fix to the main PR, then we can merge it.

Makefile
```diff
@@ -37,7 +35,7 @@ javascript: update-npm-requirements
 sass:
 	python scripts/compile_sass.py
 
-verify-generated-files: javascript sass
+verify-generated-files:
```
efischer19: This is the change I'll be movjng over - it was added for travis testing, but travis also does a `make install` earlier in its build process, so we're just repeating work needlessly

scripts/make-wheels.sh
```diff
@@ -1,15 +0,0 @@
...
-# Since the archives contain binary distributions, however,
-# you MUST run the script within an Ubuntu 12.04 box.
-
-cd `dirname $BASH_SOURCE` && cd ..
-
-pip install --upgrade pip
-pip install wheel
-
-WHEELHOUSE="scripts/data/wheelhouse"
-pip wheel --wheel-dir=$WHEELHOUSE -r requirements/wheels.txt
```
efischer19: Oh, I'm also moving this script deletion over as well

PR #1004, opened 2017-05-24
Instructor Dashboard Fixes

efischer19:
# [EDU-156](https://openedx.atlassian.net/browse/EDUCATOR-156)
# [EDU-118](https://openedx.atlassian.net/browse/EDUCATOR-118)

The fix for EDU-156 was just a small CSS change, we now truncate larger numbers. I also dropped the font size of the header numbers to fit a few more digits. The header numbers can now fit up to 5 digits (<100,000), and the smaller per-problem numbers can go up to 7 (<10,000,000).

EDU-118 requires a bit of context. Initially, there was no staff grading in ORA. Thus, the valid states for a workflow were `self`, `peer`, `waiting`, `done`, and `cancelled`. The difference between `peer` and `waiting` was that `peer` meant the learner needed to grade peers to complete their obligations, and `waiting` meant they'd done all they needed to do, but not enough peers had graded them ([source](https://github.com/edx/edx-ora2/blob/master/openassessment/workflow/models.py#L62)).

Once `staff` was added, things got a bit more complicated. The naive step definition was to say that a given workflow is in state `staff` if that was the next step to complete, and the user had not yet fulfilled their obligations in that step. However, there is *never* anything for the learner to do in a staff step! ([source](https://github.com/edx/edx-ora2/blob/master/openassessment/assessment/api/staff.py#L29)) When this feature was first merged, we sidestepped the issue by naively treating all `waiting` entries for problems that contain a staff assessment as 'staff'. However, it wasn't implemented perfectly, leading to the bug this PR fixes.

Now, I've added a separate column for `waiting`, which will be used when there is no staff grading step present. When staff grading is enabled for a problem, `waiting` is treated as `staff`. This means, as a side-effect, that any problem that has both staff and peer grading will not be able to differentiate between those 2 states, all `waiting` entries will be smashed together. I think this is acceptable, given the fact that staff grades already take precedence over peer grades when both are present.


efischer19:
Screenshot of first commit, fixing layout issues by truncation:
screen shot 2017-05-24 at 4 29 33 pm

efischer19:
Screenshots:
screen shot 2017-05-26 at 1 42 24 pm
screen shot 2017-05-26 at 2 20 24 pm

openassessment/xblock/static/sass/oa/views/_oa-course-items-listing.scss
```diff
@@ -64,22 +64,23 @@
   }
 
   .ora-summary-value {
-    font-size: 2.2em;
+    font-size: 1.5em;
     padding-top: 3px;
+    text-overflow: ellipsis;
+    overflow: hidden;
```
thallada: If this text can be hidden, I think that we should make sure that the `title` HTML attribute should be set on this element. That way, the user can hover over the value that has an ellipsis and get the full value in a tooltip.

openassessment/xblock/static/css/openassessment-ltr.css
```diff
@@ -1,3 +1,3 @@
...
...apper--metadata:last-child,.openassessment .step--peer-assessment .peer-assessment__assessment .question__answers .wrapper--metadata:last-child,.openassessment .step--self-assessment .self-assessment__assessment .question__answers .wrapper--metadata:last-child,.step--student-training .student-training__assessment .question__answers .wrapper--metadata:last-child{margin-right:0}}.openassessment .staff-assessment__assessment .question__answers .answer__tip,.openassessment .step--peer-assessment .peer-assessment__assessment .question__answers .answer__tip,.openassessment .step--self-assessment .self-assessment__assessment .question__answers .answer__tip,.step--student-training .student-training__assessment .question__answers .answer__tip,.openassessment .staff-assessment__assessment .question__answers .answer__points,.openassessment .step--peer-assessment .peer-assessment__assessment .question__answers .answer__points,.openassessment .step--self-assessment .self-assessment__assessment .question__answers .answer__points,.step--student-training .student-training__assessment .question__answers .answer__points{display:block}.openassessment .staff-assessment__assessment .question__answers .
```
thallada: I'm surprised there is built CSS and JS checked into git here. Is there a reason for that? It's usually an anti-pattern because of these giant impossible to read diffs that are created on every little sass/js change.

openassessment/xblock/static/sass/oa/views/_oa-course-items-listing.scss
```diff
@@ -64,22 +64,23 @@
   }
 
   .ora-summary-value {
-    font-size: 2.2em;
+    font-size: 1.5em;
     padding-top: 3px;
+    text-overflow: ellipsis;
```
dsjen: Nit: this css class and the `backgrid` one share same `text-overflow` and `overflow` attributes and it could b pulled into a placeholder selector (e.g. [%ui-box in Insights](https://github.com/edx/edx-analytics-dashboard/blob/aec067b1dddf7bfe10299163b720578fa2cf622a/analytics_dashboard/static/sass/_base.scss#L73) and ["extended" in an actual class](https://github.com/edx/edx-analytics-dashboard/blob/aec067b1dddf7bfe10299163b720578fa2cf622a/analytics_dashboard/static/sass/_base.scss#L355).

openassessment/xblock/static/js/src/lms/oa_course_items_listing.js
```diff
@@ -128,6 +132,10 @@
 
             if (itemId in data) {
                 _.extend(self.oraData[i], data[itemId]);
+                if (self.oraData[i].staff_assessment) {
```
dsjen: Nit: Isn't `self.oraData[i]` the same as `oraItem`?  It seems a bit inconsistently used (or not used!) prior to the PR.

openassessment/xblock/static/js/src/lms/oa_course_items_listing.js
```diff
@@ -151,27 +159,21 @@
...
                 var idx = 0;
-                var realKey = key;
-                if (key === 'waiting' && staffGradeRequired) {
-                    realKey = 'staff';
-                }
-                if (realKey in summaryDataMap) {
-                    idx = summaryDataMap[realKey];
+                if (key in summaryDataMap) {
+                    idx = summaryDataMap[key];
                     if (summaryData[idx].num) {
```
dsjen: Is it possible that this will result in unexpected behavior if `summaryData[idx].num` is `0` vs. `undefined`?  I guess I'm unsure if this `if` statement is intended to account for both of those cases.

dsjen:
Also, thanks for the excellent write up and description of the problem and solution this morning--very helpful!

openassessment/xblock/static/css/openassessment-ltr.css
efischer19: Yeah, it's definitely not ideal - I think it's a case of "it worked this way initially and no one has ever taken the time to fix it".

Ever since I've been working in this repo, I've been treating these giant diffs as "Okay, I see you touched some js and the minified file is different somehow, cool." A while back I added [this check](https://github.com/edx/edx-ora2/blob/master/Makefile#L38) to our travis builds, so you can't pass CI unless the checked-in minified files are correctly updated.

openassessment/xblock/static/sass/oa/views/_oa-course-items-listing.scss
efischer19: Great note! Always good to be in the habit of DRYing things a bit.

openassessment/xblock/static/js/src/lms/oa_course_items_listing.js
efischer19: I don't think it matters here - `num` is a boolean value. `summaryData` gets constructed from `this._columns`, which are statically defined above.

efischer19:
Thanks for the review @edx/educator-dahlia! I'm going to squash-n-merge here once tests pass, then get an ORA release going.

openassessment/xblock/static/css/openassessment-ltr.css
thallada: Ok, I just wanted to point it out. Definitely not something that needs to be fixed in this PR!

openassessment/xblock/static/js/src/lms/oa_course_items_listing.js
dsjen: 👍

PR #1014, opened 2017-06-22
Fix unable to view ORA responses (files) in leaderboard

awaisdar001:
### Description

With the new updates in the ORA file upload, we can now add a description with the uploaded files during submissions. The leaderboard view had a bug, which was not handling file uploads correctly with the descriptions. I have updated the view and fixed the failing tests in this PR. 

### Screenshots after fix
screen shot 2017-07-04 at 2 05 18 pm

### Sandbox
- [ ] https://ora2.sandbox.edx.org
- [ ] https://ora2.sandbox.edx.org/courses/course-v1:edX+DemoX+Demo_Course/courseware/9fca584977d04885bc911ea76a9ef29e/07bc32474380492cb34f76e5f9d9a135/?activate_block_id=block-v1%3AedX%2BDemoX%2BDemo_Course%2Btype%40sequential%2Bblock%4007bc32474380492cb34f76e5f9d9a135

### Reviewers
- [x] @noraiz-anwar 
- [ ] @ssemenova 
 
### FYIs
- [x] @efischer19  
- [x] @cahrens  
 
### Post-review

### DIFF
--- a/openassessment/xblock/leaderboard_mixin.py
+++ b/openassessment/xblock/leaderboard_mixin.py
@@ -76,16 +76,18 @@ class LeaderboardMixin(object):
         for score in scores:
             score['files'] = []
             if 'file_keys' in score['content']:
-                for key in score['content']['file_keys']:
-                    url = ''
-                    try:
-                        url = file_upload_api.get_download_url(key)
-                    except FileUploadError:
-                        pass
-                    if url:
-                        score['files'].append(url)
+                file_keys = score['content'].get('file_keys', [])
+                descriptions = score['content'].get('files_descriptions', [])
+                for idx, key in enumerate(file_keys):
+                    file_download_url = self._get_file_download_url(key)
+                    if file_download_url:
+                        file_description = descriptions[idx] if idx < len(descriptions) else ''
+                        score['files'].append((file_download_url, file_description))
+
             elif 'file_key' in score['content']:
-                score['files'].append(file_upload_api.get_download_url(score['content']['file_key']))
+                file_download_url = self._get_file_download_url(score['content']['file_key'])
+                if file_download_url:
+                    score['files'].append((file_download_url, ''))
             if 'text' in score['content'] or 'parts' in score['content']:
                 submission = {'answer': score.pop('content')}
                 score['submission'] = create_submission_dict(submission, self.prompts)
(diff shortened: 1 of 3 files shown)


openassessment/xblock/leaderboard_mixin.py
```diff
@@ -76,16 +76,18 @@ def render_leaderboard_complete(self, student_item_dict):
...
-                    url = ''
-                    try:
-                        url = file_upload_api.get_download_url(key)
-                    except FileUploadError:
-                        pass
-                    if url:
-                        score['files'].append(url)
+                file_keys = score['content'].get('file_keys', [])
+                descriptions = score['content'].get('files_descriptions', [])
+                for idx, key in enumerate(file_keys):
```
noraiz-anwar: Does it make sense to do this operation like
```python
file_download_urls = list(map(lambda file_download_url: self._get_file_download_url(file_download_url), file_keys))

```
and then

``` python
score['files'] = map(None, file_download_urls, descriptions)
```
OR
``` python
score['files'] = [file for file in izip_longest(file_download_urls,descriptions)]
```

openassessment/xblock/leaderboard_mixin.py
```diff
@@ -76,16 +76,18 @@ def render_leaderboard_complete(self, student_item_dict):
...
+                    file_download_url = self._get_file_download_url(key)
+                    if file_download_url:
+                        file_description = descriptions[idx] if idx < len(descriptions) else ''
+                        score['files'].append((file_download_url, file_description))
+
             elif 'file_key' in score['content']:
-                score['files'].append(file_upload_api.get_download_url(score['content']['file_key']))
+                file_download_url = self._get_file_download_url(score['content']['file_key'])
+                if file_download_url:
+                    score['files'].append((file_download_url, ''))
```
awaisdar001: @noraiz-anwar Before multiple file feature was added, we used to treat/add single files. Single uploads don't have any description as well. When the multi-upload feature was added, this changed the way we use to handle files. 

This method is tackling both keys, `file_key` and `file_keys` for the backward compatibility.

openassessment/xblock/leaderboard_mixin.py
awaisdar001: @noraiz-anwar Let's keep it simple as that, I find it more readable and understandable. Although your suggestion has less LOC but I will vote for readability.

What do you think?

noraiz-anwar: Yeah. I think it is good to keep code readable.

openassessment/xblock/leaderboard_mixin.py
noraiz-anwar: Makes sense.

PR #1019, opened 2017-07-03
This check is no longer valid, removing it as focus is not on the header

staubina:
button

This test is obsolete with the a11y refactoring we did.  It is not failing due to the changes in the UI for the new course outline.

This will also resolve the issues with ORA2 Build failures.


cahrens:
👍

PR #1027, opened 2017-07-21
Bump version of edx-submissions and package version.

doctoryes:
https://openedx.atlassian.net/browse/LEARNER-1977

The submissions version needs to be bumped here as well to get the submissions change over to edx-platform.

@iloveagent57 @efischer19 FYI


doctoryes:
As this change is low-risk, I'm merging.

PR #1028, opened 2017-08-02
Fix empty file description in ORA response - EDUCATOR-1054

awaisdar001:
### Description

With the new updates in the ORA file upload, we can now add a description with the uploaded files during submissions. There was a bug, which was not handling if only spaces are added in the file description. File uploads are successful, but it's just with ` ` as a description the link of the file is not viewable. I have updated the view to clean the description and strip off spaces. 

### Screenshots after fix
screen shot 2017-08-02 at 3 59 12 pm

### Sandbox
- [ ] https://ora2.sandbox.edx.org
- [ ] https://ora2.sandbox.edx.org/courses/course-v1:edX+DemoX+Demo_Course/courseware/3b31cd75db7347d4bff662234ca3f0ff/ff23dff05cbe426fb3c8edca9cf9bad3/1?activate_block_id=block-v1%3AedX%2BDemoX%2BDemo_Course%2Btype%40vertical%2Bblock%40b9d0951bfec8478fb0fa4aac5b42f2c4

### Reviewers
- [ ] @noraiz-anwar 
- [ ] @efischer19 
 
### FYIs
- [ ] @efischer19  
- [ ] @cahrens  
 
### Post-review


cahrens:
@efischer19 When you merged this OSPR, what was the requirement about adding a description? For accessibility, I think that a non-blank description should be required, and that the better solution here might be to require a non-blank description at upload time (although I suppose we would still need to support existing uploaded files without descriptions).

efischer19:
Yeah, that's a good point @cahrens - "only spaces" is a pretty bad way to satisfy a11y. @sstack22, what do you think? I'd vote for disallowing new "blank" descriptions, and a "the user didn't include a description" clickable string to handle existing instances.

sstack22:
@efischer19 , @cahrens  - yeah definitely, the user should be required to enter an actual description.

awaisdar001:
@cahrens @efischer19 @sstack22 Ok, Seems like I would need to change the fix and update the JS to force user to enter Description. If user enters all spaces, do not enable Upload button.

efischer19:
Sounds great @awaisdar001. What happens currently for existing images uploaded without a description? We ought to add some placeholder text there too

awaisdar001:
@efischer19 Files without description do not show up to the users. So learners report File missing on the submissions. I think the `.strip()` will do the trick for the existing submissions. The existing blank or only spaces file description will show up as `View the files associated with this submission: #1`

awaisdar001:
@cahrens @efischer19 Can you please help me figure out why tests are failing on this PR?

efischer19:
To close the loop here - we discussed in the testeng room, the answer is "ora is in a unstable state right now". You should be able to get python tests running by pinning edx-submissions to 2.0.7 temporarily, js tests are flaky and won't be resolved before the 2.0.0 upgrade work is finished (that's currently blocked on broken acceptance tests, but will get wrapped up in the next day or two)

openassessment/xblock/test/test_submission.py
```diff
@@ -212,11 +212,10 @@ def test_remove_all_uploaded_files(self, xblock):
             course_id='test_course',
             anonymous_student_id='test_student',
         )
-
         download_url = api.get_download_url("test_student/test_course/" + xblock.scope_ids.usage_id)
         resp = self.request(xblock, 'download_url', json.dumps(dict()), response_format='json')
         self.assertTrue(resp['success'])
-        self.assertEqual(download_url, resp['url'])
+        self.assertTrue(resp['url'].startswith(download_url))
```
awaisdar001: I updated the `assertEqual`  to `assertTrue ` because the test becomes flaky due to the query string in the `download_url`. Query string parameters are time-dependent (expiration and signature)

efischer19:
(don't forget to squash!)

awaisdar001:
@efischer19 Thank you for the all the help, The tests are passing now. 
I will need a 2nd thumbs up for merge. 

For your question, this PR can wait until the next week.

awaisdar001:
Merging this pr.
FYI: @efischer19

efischer19:
👍 I'll get this deployed in a fresh release early next week

PR #1031, opened 2017-08-03
master *just* passed last night

efischer19:


efischer19:
[This PR fails js tests](https://travis-ci.org/edx/edx-ora2/jobs/260741134). It is "master + 1 random file".

[master by itself passes js tests](https://travis-ci.org/edx/edx-ora2/jobs/260378195). I just reran them *after* the above failure.

PR #1030, opened 2017-08-03
2.0.0 release

efischer19:
- [x] update translations
- [x] actually bump up the required version of `jsonfield` - FYI @MichaelRoytman 
- [x] create a PR on edx-ora2 containing the version bump and translation updates.
- [x] get a green Travis build on said PR https://travis-ci.org/edx/edx-ora2/builds/260763218
- [ ] merge to master
- [ ] green build on master
- [ ] create a release tag on GitHub https://github.com/edx/edx-ora2/releases
- [ ] manually inspect diff at https://github.com/edx/edx-ora2/compare/2.0.0...1.4.9
- [ ] create PR to update `requirements/edx/github.txt` in `edx-platform`
- [ ] make sure that the ORA acceptance tests have run and passed
- [ ] get green build on the edx-platform PR, merge

PR #1032, opened 2017-08-04
EDUCATOR-1090

efischer19:
Needed to get https://github.com/edx/edx-submissions/pull/67 to prod.

There are a large number of changes in master that I do *not* want to include in this release, so this is 1.4.10 compared to the current 2.0.0 version on master.

FYI @jibsheet @feanil @sstack22

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,6 @@ script:
 branches:
     only:
       - master
-      - ora-staff-grading
+      - release_branch/1.4.9
 after_success:
   coveralls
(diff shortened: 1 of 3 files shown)


efischer19:
I've set up this PR against a branch that tracks the 1.4.9 release for easy diff viewing. The 1.4.10 tag will be created from the tip of this branch.

efischer19:
Python tests passed, which are all I was potentially worried about. Js tests are flaky as noted [here](https://github.com/edx/edx-ora2/pull/1030#issuecomment-320077513), and will be fixed with the impending 2.0.0 release.

Merging this patch.

PR #1033, opened 2017-08-05
EDU-1090, take 2 part 2

efischer19:
https://github.com/edx/edx-submissions/pull/68


efischer19:
travis is going to fail due to the flaky js test issue

Merging this anyways, as the only change is edx-submissions and I've manually validated that exists: https://pypi.python.org/pypi/edx-submissions/2.0.9

PR #1034, opened 2017-08-07
Finalize 2.0.1 release

efischer19:
This PR will serve to collect:
- edx-submissions changes, [as we get that back to normal following this past weekend](https://github.com/edx/edx-submissions/pull/69)
- acceptance test cleanup
- fresh translations/version bump

Travis tests are going to break until edx-submissions 2.0.10 exists


openassessment/workflow/migrations/0003_remove_uuid_test.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ('workflow', '0002_remove_django_extensions'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='assessmentworkflow',
+            name='uuid',
+        ),
+    ]
```
efischer19: Worth noting: this migration ends up being unreversible, due to the default value `uuid()` being identical when run a bunch of times at once. There ought to be a better way to do that, maybe I'll mess with the default value in the initial "prepare for removal" work

efischer19: Ew

https://docs.djangoproject.com/en/1.9/howto/writing-migrations/#migrations-that-add-unique-fields

efischer19:
The way the migrations work:
- Going forward: Remove uniqueness constraint, do nothing, drop column
- Going backward: Add column (nullable, no default), `RunPython` to generate unique values, add uniqueness constraint

This 3-part operation won't be needed in this PR, just wanted to call it out for future use when we are ready to drop the column.

efischer19:
@jibsheet Dropping the uuid field looks like it's going to work out. I'm going to split out the changes to actually drop the uuid column into a separate long-lived PR, and update this one to deprecate it for the time being.

jibsheet:
@efischer19 thanks for the explanation, I was wondering where the migration was, but it makes more sense (that you're going to remove it from the api calls and we'll deal with the DB later)

openassessment/xblock/grade_mixin.py
```diff
@@ -374,7 +374,7 @@ def _get_assessment_part(title, feedback_title, part_criterion_name, assessment)
         if len(assessments) > 0:
             first_assessment = assessments[0]
             option = first_assessment['option']
-            if option and option.get('points'):
+            if option and option.get('points', None) != None:
```
efischer19: I think this fixes most of our problems with acceptance tests - `0 points` was being hidden because 0 is falsy.

screen shot 2017-08-10 at 11 02 05 am

efischer19:
We have a successful acceptance test run! https://tools-edx-jenkins.edx.org/job/ora2/job/run-acceptance-tests/155/

I'll get https://github.com/edx/edx-submissions/pull/70 resolved, then we can merge this and finish the 2.0.0 release

efischer19:
I've done 3 things in this PR:

- removed usage of the Workflow model's `uuid` field. This will prevent the serialization change that affected edx-submissions from breaking edx-ora2, and allows for us to potentially remove that column later (https://github.com/edx/edx-ora2/pull/1035 has that work started, but we can't do it all in one go due to blue/green deploying)
- fixed acceptance tests via `unicode`-ing and fixing some incorrect logic in `_get_assessment_part`
- updated edx-submissions to the latest, most correct, last-weekend's-problems-are-in-the-rearview-mirror version

efischer19:
Thanks @jibsheet and @dsjen!

efischer19:
I'm going to tack on a translations update before merging this - no need to review that per ORA's release docs.

efischer19:
The i18n update is taking *forever*, probably due to the massive amount of files that have changed during this process.

Here's the release checklist I'll be following:

- [ ]  update translations (especially if many display strings have changed, otherwise you may be able to skip this step)
- [x] create a PR on edx-ora2 containing the version bump and translation updates. (it's this one)
- [ ] get a green Travis build on said PR (blocked on i18n updates - will update once tests have run on the actual commit to be merged)
- [ ] merge to master, get a green build
- [ ] create a release tag on GitHub https://github.com/edx/edx-ora2/releases
- [ ] manually inspect diff at https://github.com/edx/edx-ora2/compare/0.x.n-1...0.x.n, email contributors (Should be 1.final against 2.0.1, but that diff will be so large as to be unwieldy)
- [ ] create PR to update `requirements/edx/github.txt` in `edx-platform` - https://github.com/edx/edx-platform/pull/15736
- [ ] make sure that the ORA acceptance tests have run and passed (blocked, waiting for i18n updates/merge/new tag)
- [ ] get green build on the edx-platform PR, merge
- [ ] manually test changes on stage (smoke test)

PR #1035, opened 2017-08-09
Drop Workflow.uuid field

efischer19:
This PR will need to remain open for a while, as we want to deprecate usage of the field before running the migration to drop it in the database. Builds on https://github.com/edx/edx-ora2/pull/1034, I'll update this PR to pint at master after that lands.

Note that the migrations to drop the table are a set of 3 files, this allows for the migration to be reversible (since adding unique, non-null columns to a table is not a simple operation)

### DIFF
--- a/openassessment/workflow/migrations/0004_reverse_gen_uuid.py
+++ b/openassessment/workflow/migrations/0004_reverse_gen_uuid.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models, transaction
+import uuid
+
+
+def gen_uuid(apps, schema_editor):
+    workflow_model_class = apps.get_model('workflow', 'AssessmentWorkflow')
+    total_len = workflow_model_class.objects.count()
+    current_index = 0
+    chunk_size = 1000
+    while current_index < total_len:
+        end_chunk = current + chunk_size if total_len - chunk_size >= current_index else total_len
+        with transaction.atomic():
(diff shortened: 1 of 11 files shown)


openassessment/workflow/migrations/0004_reverse_gen_uuid.py
```diff
@@ -0,0 +1,36 @@
...
+
+def gen_uuid(apps, schema_editor):
+    workflow_model_class = apps.get_model('workflow', 'AssessmentWorkflow')
+    total_len = workflow_model_class.objects.count()
+    current_index = 0
+    chunk_size = 1000
+    while current_index < total_len:
+        end_chunk = current + chunk_size if total_len - chunk_size >= current_index else total_len
+        with transaction.atomic():
+            for workflow in workflow_model_class.objects.all()[current_index:end_chunk].iterator():
```
efischer19: Dev note: check whether this change is needed in the recent submissions management command as well

efischer19: Yarp

```
mysql> select count(*) from submissions_submission;
+----------+
| count(*) |
+----------+
|  1767566 |
+----------+
1 row in set (2.98 sec)

mysql> select id from submissions_submission order by id desc limit 1;
+---------+
| id      |
+---------+
| 1768723 |
+---------+
1 row in set (0.00 sec)
```

efischer19:
For posterity: at the time of this comment, this PR has been rebased on top of https://github.com/edx/edx-ora2/pull/1034, excluding the pending i18n updates

PR #1048, opened 2017-09-20
Added validate_translations logic

fysheets:
Updates related to adding `validate_translations` logic to Makefile and updates necessary in related configurations.

`make validate_translations` will run via travis and error if updated source files are not found.


Makefile
```diff
@@ -35,6 +35,28 @@ verify-generated-files:
 	@git diff --quiet || (echo 'Modifications exist locally! Run `make javascript` or `make sass` to update bundled files.'; exit 1)
 
 ################
+#Translations Handling
+################    
+# creates the django-partial.po & django-partial.mo files
+extract_translations:
+	i18n_tool extract -v
```
efischer19: Could you add `.tox` to the ignore directories specified here: https://github.com/edx/edx-ora2/blob/76acce9523b9a1d4b54294872ae45930aacc6f44/openassessment/locale/config.yaml#L76?

efischer19: whoa, new github link expandos

Makefile
```diff
@@ -35,6 +35,28 @@ verify-generated-files:
...
+
+# compiles the *.po & *.mo files
+compile_translations:
+	i18n_tool generate -v
```
efischer19: Unfortunately, this command doesn't take `--ignore`, so it's scanning everything under `edx-ora2/.tox/` as well. From checking the logs, it looks like `edx-ora2/openassessment` contains all the strings we actually care about (`edx-ora2/conf/locale/` is hit as well, but that's just a symlink back to `edx-ora2/openassessment/locale` anyways). Could you try changing this make target to `cd ./openassessment/` first, then change back after?

PR #1049, opened 2017-09-29
Move model imports into methods, enabling django upgrade

pwnage101:
These changes rearrange model imports in order to enable Django 1.9+ support.

https://github.com/edx/edx-platform/pull/16110 confirms these changes do not break edx-platform tests under the existing Django 1.8.

fyi @doctoryes

### DIFF
--- a/openassessment/xblock/course_items_listing_mixin.py
+++ b/openassessment/xblock/course_items_listing_mixin.py
@@ -8,7 +8,6 @@ import json
 from webob import Response
 from xblock.core import XBlock
 
-from openassessment.data import OraAggregateData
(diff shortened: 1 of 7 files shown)


pwnage101:
The django 1.11 update is still in flight, and there are still other satellite repos to address so I don't see the urgency to release right now.

The norm is to squash all the fixup commits, right?

doctoryes:
@pwnage101 Yeah - you could really just squash it all into a single commit in this case.

pwnage101:
Sorry, I missed a whole slew of other submissions.api imports in the other xblock mixin files, so I added those and now re-running platform tests: https://github.com/edx/edx-platform/pull/16110

PR #1055, opened 2017-11-08
Move django-nose to test dependencies

jmbowman:
django-nose is only referenced in the test settings, so it should be safe to only install it when running tests.  This keeps it out of `setup.py`'s `install_requires`, which has been pulling it into edx-platform even though we no longer use nose to run tests there.


jmbowman:
@doctoryes It looks like some of our npm dependencies aren't pinned, the JS test run regenerates the minified files, and those files exist in the repo.  Hence it's choking on a git command after the files change, counting as an uncommitted change to the repo.  @edx/fedx-team, should I just check in what now gets generated or take a stab at cleaning up the dependencies?

jmbowman:
@efischer19 recommended just checking in the updated files, which did resolve the issue (with JS tests still passing).

PR #1058, opened 2017-11-16
WYSIWYG editor for prompts (in studio)

dmitry-viskov:
TynyMCE is used for in studio for HTML blocks.
This PR adds the same functionality for prompts:

wysiwyg_1

wysiwyg_2

@dpowell may provide more details why we need this feature


edx-webhook:
Thanks for the pull request, @dmitry-viskov! I've created [OSPR-2014](https://openedx.atlassian.net/browse/OSPR-2014) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

dpowell:
We currently are using a separate HTML component to provide richer formatting of prompts for ORAs. Having the full prompt attached to the ORA itself will allow us to select an ORA from a Library ensuring that the prompt comes with it (otherwise those two components would have to be somehow "linked" in the library). It also makes versioning of changes to an ORA simpler without having to synchronize changes to multiple components simultaneously. Lastly, reporting on ORA responses that might need to include the prompt itself (for consumers not already familiar with the content of the course) is simplified if the prompt itself is part of the ORA component the learner responses are against.

nedbat:
Forgive my ORA-ignorance: does this change the prompt from a plain text field to an HTML field? Aren't there backward compatibility concerns?

nedbat:
@edx/educator Take a look.

dmitry-viskov:
hi @nedbat 

>Forgive my ORA-ignorance: does this change the prompt from a plain text field to an HTML field? 

yes

>Aren't there backward compatibility concerns?

I can't imagine any backward compatibility concerns but if they exists - we may create toggle to switch between text/html

nedbat:
The backward compatibility issue is that people have created plain text data (prompts) which are now being interpreted as HTML.  Perhaps this only matters if they have something that looks like an HTML tag?

openassessment/xblock/static/js/src/studio/oa_edit_prompts.js
```diff
@@ -21,7 +23,10 @@ OpenAssessment.EditPromptsView = function(element, notifier) {
             removeButtonClass: "openassessment_prompt_remove_button",
             containerItemClass: "openassessment_prompt",
             notifier: notifier,
-            addRemoveEnabled: this.addRemoveEnabled
+            addRemoveEnabled: this.addRemoveEnabled,
+            containerItemsKwargs: {
```
sandroroux: Instead of "containerItemsKwargs", name this field "addWysiwyg" and have it pass it the value of "addWysiwyg" from above.

openassessment/xblock/static/js/src/studio/oa_container_item.js
```diff
@@ -75,9 +75,17 @@ OpenAssessment.ItemUtilities = {
  Returns:
  OpenAssessment.Prompt
  **/
-OpenAssessment.Prompt = function(element, notifier) {
+OpenAssessment.Prompt = function(element, notifier, kwargs) {
     this.element = element;
     this.notifier = notifier;
+    this.kwargs = kwargs || {addWysiwyg: false};
+
+    var elId = $(element).find('textarea').attr('id');
```
sandroroux: The Prompt constructor should not be modifying "element". Looking at the constructor's original code, it expects to receive its "element" from wherever Prompt is instantiated. 

My recommendations:
1. Wrap this code in a helper method (lines 83-88).
2. Modify "element", then pass its modified value into the Prompt constructor. We construct Prompt in "OpenAssessment.EditPromptsView".

openassessment/xblock/static/js/src/studio/oa_container_item.js
```diff
@@ -109,7 +117,16 @@ OpenAssessment.Prompt.prototype = {
      **/
     description: function(text) {
         var sel = $('.openassessment_prompt_description', this.element);
-        return OpenAssessment.Fields.stringField(sel, text);
+        if (this.kwargs.addWysiwyg) {
```
sandroroux: You shouldn't have the prototype of the class introspecting for a subclass's field ("addWysiwyg"). Why not override "description" in the Prompt class rather than modify "description" in the prototype?

openassessment/xblock/static/js/src/studio/oa_edit_prompts.js
```diff
@@ -21,7 +23,10 @@ OpenAssessment.EditPromptsView = function(element, notifier) {
             removeButtonClass: "openassessment_prompt_remove_button",
             containerItemClass: "openassessment_prompt",
             notifier: notifier,
-            addRemoveEnabled: this.addRemoveEnabled
+            addRemoveEnabled: this.addRemoveEnabled,
```
sandroroux: Why did you add "addRemoveEnabled"?

openassessment/templates/openassessmentblock/oa_submission_answer.html
```diff
@@ -8,7 +8,7 @@ <h5 class="submission__answer__part__text__title">{% trans "The question for thi
             </div>
             <article class="submission__answer__part__prompt">
                 <div class="submission__answer__part__prompt__value">
-                    {{ part.prompt.description|linebreaks }}
+                    {% autoescape off %}{{ part.prompt.description }}{% endautoescape %}
```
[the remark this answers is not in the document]
dmitry-viskov: @sandroroux I agree that this part requires additional development. There are few ideas how to implement backward compatibility: 1. To add new option to the ORA settings what type of prompt should be used (simple text or html). By default it will be set as simple text. 2. The most easiest way - to check the first letter in the prompt - if it is `<` we may assert that this field is HTML. Otherwise use old construction `{{ part.prompt.description|linebreaks }}`. What do you think? May be you have more elegant solutions?

openassessment/xblock/static/js/src/studio/oa_container_item.js
```diff
@@ -75,9 +75,17 @@ OpenAssessment.ItemUtilities = {
  Returns:
  OpenAssessment.Prompt
  **/
-OpenAssessment.Prompt = function(element, notifier) {
+OpenAssessment.Prompt = function(element, notifier, kwargs) {
```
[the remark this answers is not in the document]
dmitry-viskov: I'll add comment but i disagree about `kwargs`:
As I may see from the code base - all elements are implemented as the `Container` items: https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/static/js/src/studio/oa_container.js#L75
It is `Prompt`, `RubricOption`, `RubricCriterion` and etc. So i can't just pass some additional parameter to `Prompt` constructor. Of course it will not do anything wrong if I'll pass it to all `ContainerItem`-s but as for my opinion it isn't good solution. Also I may add `if` statement but as for my opinion current solution with `kwargs` is universal and more elegant

openassessment/xblock/static/js/src/studio/oa_edit_prompts.js
dmitry-viskov: @sandroroux it wasn't me :wink:  I've just added comma - i may assume it is the reason of this misunderstanding :smiley:

dmitry-viskov:
hi @sandroroux . Thanks for the review. I've recreated this feature using another approach - I'll add new `WysiwygPrompt` class instead of modifying existing `Prompt`. So in this case we need not `kwargs`, `containerItemsKwargs`, `addWysiwyg` and other tricky variables anymore. Please review. Thanks.

And I've posted variants of possible solutions for the backward compatibility

openassessment/xblock/static/js/src/studio/oa_edit_prompts.js
sandroroux: Ah, my mistake. Thanks for clarifying!

openassessment/templates/openassessmentblock/oa_submission_answer.html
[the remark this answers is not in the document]
sandroroux: Great ideas. I prefer option 1. Do you mind adding an ORA setting for HTML prompts?

openassessment/xblock/static/js/src/studio/oa_edit_prompts.js
```diff
@@ -13,15 +13,25 @@ OpenAssessment.EditPromptsView = function(element, notifier) {
     this.editorElement = $(this.element).closest("#openassessment-editor");
     this.addRemoveEnabled = this.editorElement.attr('data-is-released') !== 'true';
 
+    var addWysiwyg = window.tinyMCE !== undefined;
```
sandroroux: Why do you calculate `addWysiwyg` at this point? Could you check for tinyMCE in `WysiwygPrompt`?

You don't have to pass the `addWysiwyg` to `WysiwygPrompt` if it can check for the value internally.

openassessment/templates/openassessmentblock/oa_submission_answer.html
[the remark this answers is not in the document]
dmitry-viskov: @sandroroux I suggest to add "HTML prompts" selector with options "Yes"/"No" ("No" is default) here (over the "Text Response" option):
ora

Do you agree with such solution?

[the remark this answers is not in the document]
dpowell: Do we need the user to set it? Can't this be a hidden setting we set automatically if using the HTML editor?

openassessment/xblock/static/js/src/studio/oa_edit_prompts.js
dmitry-viskov: @sandroroux i'm not quite sure that understand you, You asked me not to modify "element" in the constructor:

>The Prompt constructor should not be modifying "element". Looking at the constructor's original code, it expects to receive its "element" from wherever Prompt is instantiated.
>
>My recommendations:
>
>Wrap this code in a helper method (lines 83-88).
>Modify "element", then pass its modified value into the Prompt constructor. We construct Prompt in "OpenAssessment.EditPromptsView".

So I've created special function [prepareElement](https://github.com/edx/edx-ora2/pull/1058/files#diff-d3ccf467bd20f391e1bb1d3a4298b1b0R27) where I need to detect is tinyMCE available or not. 

Besides this we should know what type of element should be used: [wysiwyg or textarea](https://github.com/edx/edx-ora2/pull/1058/files#diff-d3ccf467bd20f391e1bb1d3a4298b1b0R19)

openassessment/templates/openassessmentblock/oa_submission_answer.html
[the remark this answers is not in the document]
dmitry-viskov: @sandroroux @dpowell  Another one solution (without changing studio UI):
1. We always use HTML editor for prompts in studio (i.e ORA settings).
2. We just change existing XBlock `prompt` to:
```
    prompt = String(
        default=None,
        scope=Scope.content,
        help="Do not use. For backwards compatibility only."
    )
``` 
3. And add new XBlock property like:
```
    prompt_html = String(
        default=DEFAULT_PROMPT,
        scope=Scope.content,
        help="The HTML prompts to display to a student."
    )
4. If `prompt` is not `None` it means that text prompt type is used - in this case we use `{{ part.prompt.description|linebreaks }}` to display description in the LMS + use encoding html tags and conversion new lines to `<br>` to display it in the ORA settings.
5. If prompt is `None` it means that new html prompt type is used - we use `{% autoescape off %}{{ part.prompt.description }}{% endautoescape %}` to display description in the LMS

File: openassessment/xblock/static/js/src/studio/oa_edit_prompts.js (hunk 7 above)
Comment (sandroroux) [reply openassessment/xblock/static/js/src/studio/oa_edit_prompts.js]: I'm sorry for the confusion. Let me explain myself. 

While I like the idea, I don't think you need to make a `WysiwygPrompt` class. It doesn't specialize `Prompt` enough, and I do not want to deepen the prototype tree. You should revert to your original `Prompt`-based implementation.

From what I see, `Prompt` can become a wysiwyg editor with two changes:
1. Attach tinyMCE to `element`, and pass it to the `Prompt` constructor.
2. Update `Prompt`'s `description` to pull content from the tinyMCE element.

You can eliminate `addWysiwyg` if you compute its value where you use it (at changes 1 and 2 – the bodies of `prepareElement` and `Prompt::description`). Computing it _in situ_ frees you from having to pass it to `Prompt` in a `kwargs` object. Removing branching on `addWysiwyg` will clean up the code.

Please let me know if my explanation doesn't make sense.

File: openassessment/templates/openassessmentblock/oa_submission_answer.html (hunk 5 above)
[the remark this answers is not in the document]
Comment (sandroroux) [reply openassessment/templates/openassessmentblock/oa_submission_answer.html]: @dmitry-viskov, I'll speak with my colleagues about how to best approach this. Many of them are out of the office for the holidays. I'll  respond with a recommendation as soon as I can. Thank you very much for your ideas!

**Comment by dmitry-viskov** [comment]
@sandroroux thanks for the comments. I've updated PR and looking forward your decision about backward compatibility

File: openassessment/xblock/static/js/src/studio/oa_container.js [hunk 8]
```diff
@@ -65,14 +65,19 @@ OpenAssessment.Container = function(ContainerItem, kwargs) {
     this.removeButtonClass = kwargs.removeButtonClass;
     this.containerItemClass = kwargs.containerItemClass;
     this.notifier = kwargs.notifier;
+    this.prepareElement = (typeof kwargs.prepareElement === 'function') ? kwargs.prepareElement : function(el) {
```
sandroroux: @dmitry-viskov, Thanks for working with me on this. I suggest localizing tinyMCE dependencies to `Prompt`. Note that these suggestions reverse my previous comment about modifying `element` outside of `Prompt`. I'm sorry for flip-flopping – it is more important to contain dependencies.

1. In `OpenAssessment.Prompt.prototype`, create a new class property called `TinyMCEEnabled`. Compute `tinyMCEEnabled` as `window.tinyMCE !== undefined`.
2. Move the `prepareElement` definition to `OpenAssessment.Prompt.prototype`. Rename the function as `attachWysiwygToPrompt`. At the top of the function definition, assert that `self.tinyMCEEnabled` is true (the assertion enforces that tinyMCE must be available for the function to be usable).
3. In the `OpenAssessment.Prompt` constructor, assign `self.element` in the following way: 
```
if (self.tinyMCEEnabled)
{
    element = self.attachWysiwygToPrompt(element);
}
self.element = element;
```

dmitry-viskov:
@sandroroux hi. PR is updated

openassessment/xblock/static/js/src/studio/oa_container_item.js
```diff
@@ -109,7 +134,16 @@ OpenAssessment.Prompt.prototype = {
      **/
     description: function(text) {
         var sel = $('.openassessment_prompt_description', this.element);
-        return OpenAssessment.Fields.stringField(sel, text);
+        if (this.tinyMCEEnabled) {
```
sandroroux: Restructure this as:
```
if (!this.tinyMCEEnabled)
{
    return OpenAssessment.Fields.stringField(sel, text);
}

var tinyEl = window.tinyMCE.get(sel.attr('id'));
if (text) {
    tinyEl.setContent(text);
} else {
    return tinyEl.getContent();
}
```

openassessment/xblock/static/js/src/studio/oa_edit.js
```diff
@@ -278,6 +278,66 @@ OpenAssessment.StudioView.prototype = {
     }
 };
 
+/* Get tinyMCE comfig */
+/* jshint unused:false */
+function tinymceCfg() {
```
sandroroux: Did you copy this configuration from elsewhere in the codebase? How did you come up with it?

openassessment/xblock/test/test_submission.py
```diff
@@ -125,7 +125,7 @@ def test_closed_submissions(self, xblock):
     def test_prompt_line_breaks(self, xblock):
```
sandroroux: @dmitry-viskov I would not modify the existing prompts. Our goal is to add your feature while supporting plaintext prompts. The unmodified defaults work in plaintext mode. Please revert `defaults.py` too.

This test should fail until your feature is backwards compatible.

sandroroux:
@dmitry-viskov Thanks for the speedy revisions. We'll discuss backwards compatibility with plaintext early next week. In the meantime, how have you been testing your feature?

openassessment/xblock/static/js/src/studio/oa_edit.js
dmitry-viskov: I took it from the edx-platform repo:
https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/js/src/html/edit.coffee#L61

openassessment/xblock/static/js/src/studio/oa_container_item.js
dmitry-viskov: ok

openassessment/xblock/test/test_submission.py
dmitry-viskov: For now I think that `defaults.py` should be modified anyway because default prompt should be adapted for the HTML editor (I may assume that HTML editor should be default editor). I suggest to return to this question when backwards compatible question will be more clear

dmitry-viskov:
@sandroroux hi

>how have you been testing your feature?

in our project we used manual tests, but probably this feature could be covered with some python/js tests (I mean to cover not WYSIWYG editor but all integration functionality)

openassessment/templates/openassessmentblock/oa_submission_answer.html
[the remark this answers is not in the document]
sandroroux: @dmitry-viskov Is there anything that prevents us from formatting "part.prompt.description" before it's injected into the template?

Where "description" is defined, can we can change plaintext into HTML?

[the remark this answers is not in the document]
dmitry-viskov: @sandroroux yes, we can. But as for my opinion it's not important. it doesn't solve backward compatibility question.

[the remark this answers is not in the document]
sandroroux: @dmitry-viskov Do you mind explaining why it wouldn't solve backwards compatibility? 

As I've understood it, removing `linebreaks` threatens backwards compatibility because plaintext `description`s will not be HTML formatted. Plaintext newlines must be converted into corresponding HTML tags before being rendered in the browser. Can we perform the conversion before the template expands `part.prompt.description`?

Could you point me to the code that generates "part.prompt.description"? The codepath may fork for HTML and plaintext cases. Can we detect between plaintext and HTML at `part.prompt.description`'s assignment?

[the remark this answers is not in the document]
dmitry-viskov: @sandroroux it doesn't matter where you make html encoding - in templates or in views or in other place. The problem is that some old prompts could contain HTML that should be displayed as HTML. line breaks is as particular case of such situation.

>Could you point me to the code that generates "part.prompt.description"? 

https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/data_conversion.py#L104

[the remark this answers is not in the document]
dmitry-viskov: so the main problem is how to determinant which prompt is used - old (created for the plain text representation) or the new one (created for the HTML representation). I've suggested solution with the new XBlock property: https://github.com/edx/edx-ora2/pull/1058#discussion_r158165750 (see above)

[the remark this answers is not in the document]
sandroroux: @dmitry-viskov I see. I thought tinyMCE wouldn't render newline characters as line breaks. Seeing as tinyMCE automatically formats plaintext line-breaks, let's tackle the backwards compatibility issue that you identified: unescaped HTML tags in plaintext prompts.

What happens when tinyMCE renders a prompt with unescaped HTML? How badly does ORA break? Can you post a screenshot of what happens?

In the meantime, I'll look into how common this is in edX courses.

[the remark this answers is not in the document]
dmitry-viskov: >Seeing as tinyMCE automatically formats plaintext line-breaks, let's tackle the backwards compatibility

I think you understood me wrong. tinyMCE doesn't automatically format plaintext line-breaks. But it is not a big problem because 1. when ORA block is published and there are some answers - xblock settings can't be changed, 2. when user will open settings - he will see that prompt is broken in the WYSIWYG editor. The main problem is in the prompt display.

Example 1:

s1

s2

Example 2:

s3

s4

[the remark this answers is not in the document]
dmitry-viskov: @sandroroux

I'm going to implement solution described in: https://github.com/edx/edx-ora2/pull/1058#discussion_r158165750

because as for my opinion this is the only way how to support backward compatibility for all cases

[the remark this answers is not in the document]
sandroroux: @dmitry-viskov Thanks for illustrating the backwards compatibility concerns. I'm sorry for the misunderstanding. Please hold off on implementing a solution.

My colleagues and I are considering how to introduce your feature. We are assessing a few options. I'll let you know how to proceed shortly. 

In the meantime, thanks for being patient with me.

dmitry-viskov:
@sandroroux hi. I've just updated PR with backward compatibility. I've added new XBlock property `prompts_type` (may be `text` or `html`). By default it is `text` and all ORA prompts are displayed as text. After user update ORA settings using WYSIWYG prompts_type will be html and all prompts will be displayed as HTML

dmitry-viskov:
hi @sandroroux . Any news according this PR?

edx-webhook:
Thanks for the pull request, @dmitry-viskov! I've created [OSPR-2164](https://openedx.atlassian.net/browse/OSPR-2164) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

sandroroux:
Hi @dmitry-viskov , I'm sorry for the delay; thanks for your patience. In order to merge this branch, we need to:
1. Ensure that all tests are passing (see the Travis CI build).
2. Write the proper tests for your feature.

I need to confirm your approach for backwards compatibility with our product team. I'll update you as soon as I can.

dmitry-viskov:
hi @sandroroux . Any news according this PR?

sandroroux:
Hi @dmitry-viskov, yes. Please see my comment above for next steps. Please let me know if you have any questions.

dmitry-viskov:
@sandroroux i don't see answer on the backwards compatibility question. I'll start to work on tests and etc only after we deal with this

dmitry-viskov:
hi @sandroroux . Any news according this PR?

sandroroux:
@dmitry-viskov Again, I'm sorry for the delay. 

I've investigated the impact of shifting all ORA prompts to HTML. If shifting ORA to tinyMCE breaks few existing prompts, I'll advocate that we migrate the few broken prompts towards supporting the HTML editor. 

As I draw conclusions from this investigation, I recommend that you fix this branch's test errors and write new tests ensuring the HTML editor works. If I understood your explanation of how tinyMCE handles plaintext, using the tinyMCE editor should not break any existing ORA tests.

The release plan is as follows: either we will roll out your implementation as it stands (in other words, "ORA supports legacy plaintext prompts and new HTML prompts"), or we will remove the logic for supporting plaintext prompts ("ORA prompts are in HTML"). Whichever of the two strategies we deploy shouldn't affect your ability to test your feature. Do you agree? If I'm missing something, please let me know and we'll plan accordingly.

Thank you for your continuing work on this.

dmitry-viskov:
@sandroroux hi. sorry. but i don't understand you at all.

I've suggested solution:
1. We always use HTML editor for prompts in studio (i.e ORA settings).
2. New XBlock property `prompts_type` was added (can be set as `text` or `html`).
3. By default it is `text` and all ORA prompts are displayed as text (backwards compatibility 
 is here - nothing will be broken after update).
4. After user update ORA settings using WYSIWYG - `prompts_type` will be `html` and all prompts will be displayed as HTML.
If you agree with this - I'll change existing tests. I will not change them without final decision because it is significant change and i don't want to do some job twice. Unfortunately I have no much time to re-write some tests twice or more time.

>using the tinyMCE editor should not break any existing ORA tests.

you are not correct. supporting backwards compatibility affects many existing tests in case if we change or add Xblock property

openassessment/templates/openassessmentblock/oa_submission_answer.html
```diff
@@ -8,7 +8,11 @@ <h5 class="submission__answer__part__text__title">{% trans "The question for thi
             </div>
             <article class="submission__answer__part__prompt">
                 <div class="submission__answer__part__prompt__value">
-                    {{ part.prompt.description|linebreaks }}
+                    {% if prompts_type == 'html' %}
+                        {% autoescape off %}{{ part.prompt.description }}{% endautoescape %}
```
bill-filler: This might be cleaner with:
```{{ part.prompt.description|safe }}```
You don't need the autoesape off and on

File: openassessment/templates/openassessmentblock/response/oa_response.html [hunk 13]
```diff
@@ -74,7 +74,11 @@ <h4 class="step__title">
                                 <h5 class="submission__answer__part__text__title">{% trans "The prompt for this section" %}</h5>
                                 <article class="submission__answer__part__prompt">
                                     <div class="submission__answer__part__prompt__copy">
-                                        {{ part.prompt.description|linebreaks }}
+                                        {% if prompts_type == 'html' %}
+                                            {% autoescape off %}{{ part.prompt.description }}{% endautoescape %}
```
bill-filler: Same comment as above: This might be cleaner with:
```{{ part.prompt.description|safe }}```
You don't need the autoesape off and on

File: openassessment/xblock/static/js/src/studio/oa_container_item.js [hunk 14]
```diff
@@ -76,11 +76,43 @@ OpenAssessment.ItemUtilities = {
...
+            /* jshint undef:false */
+            var textarea = $(el).find('textarea');
+            var text = $(textarea).val();
+            var type = $(textarea).data('type');
+            if (text && (type === 'text')) {
+                text = _.escape(text).replace(/(?:\r\n|\r|\n)/g, '<br />');
+                $(textarea).val(text);
+            }
+            var newElId = Date.now() + '-textarea-' + (Math.random() * 100);
+            $(textarea).attr('id', newElId).tinymce(tinymceCfg());
```
bill-filler: Not sure where in the code it would be set, but would be nice to give the html editor focus when it's created so the cursor starts in there. Especially when you create a new prompt the focus should be automatically in the html editor. Not a blocker but would be nice

openassessment/xblock/static/css/openassessment-ltr.css
```diff
@@ -1,3 +1,3 @@
...
...nassessment .submission__feedback__title__copy,.step--student-training .student-training__display__title,.openassessment .openassessment__steps__step .step__title .step__deadline,.staff-info__student .title--sub,.wrapper--xblock .field label,.wrapper--xblock .field .label,.wrapper--xblock .field .field-group__label,.wrapper--xblock .field--radio .option__label,.wrapper--xblock .field--checkbox .option__label,.wrapper--xblock .message .message__content,.openassessment .staff-assessment__assessment .question__answers .wrapper--input .answer__label,.openassessment .step--peer-assessment .peer-assessment__assessment .question__answers .wrapper--input .answer__label,.openassessment .step--self-assessment .self-assessment__assessment .question__answers .wrapper--input .answer__label,.step--student-training .student-training__assessment .question__answers .wrapper--input .answer__label,.openassessment .staff-assessment__assessment .question__answers .answer__tip,.openassessment .step--peer-assessment .peer-assessment__assessment .question__answers .answer__tip,.openassessment .step--self-assessment .self-assessment__assessment .question__answers .answer__tip,.step--student-training .stude
```
bill-filler: does this file need to be committed? or does it get auto-generated when building? did any scss files change?
Same comment applies for the other .css and .min.js files in this PR

bill-filler:
@dmitry-viskov 
This is awesome, thank you for the contribution. I just tested it and it works well. I did make some small comments. Please try to address them and get the tests working so we can land this.
@sandroroux 
I don't see any backward compatibility issues with this PR (unless I'm missing something). Prompts created before this will still work, they will just be rendered in plaintext as the currently are. The next time the instructor edits the prompt, then the text will be displayed in html going forward. @dmitry-viskov is this correct?

dmitry-viskov:
hi @bill-filler . thanks for the review

>Prompts created before this will still work, they will just be rendered in plaintext as the currently are. The next time the instructor edits the prompt, then the text will be displayed in html going forward. @dmitry-viskov is this correct?

yes. that's correct.

sandroroux:
@dmitry-viskov Your current implementation is good to go. Please resolve any failures in **continuous-integration/travis-ci/pr** and add tests for your feature.

Thank you for your additions to ORA, and for working with me through the review process. I'm sorry for how long it took, and appreciate your patience. 

@bill-filler will take over as the engineering reviewer.

openassessment/templates/openassessmentblock/oa_submission_answer.html
dmitry-viskov: yes. you are right. I've fixed this

openassessment/templates/openassessmentblock/response/oa_response.html
dmitry-viskov: fixed

openassessment/xblock/static/js/src/studio/oa_container_item.js
dmitry-viskov: added `focus()` after create element

openassessment/xblock/static/css/openassessment-ltr.css
dmitry-viskov: `.min.js` need to be committed (because they are used when DEBUG == `False` in the server settings). `.css` seems that was committed wrong

bill-filler:
@dmitry-viskov 
Thanks for the changes. We at edX are now working on writing tests for this feature, and will do so in another branch that we'll link to once we have it. Thank you for your help, we will reach out if we have anymore questions but you don't need to write the tests are we are working on it. @hasnain-naveed  @ziafazal

dmitry-viskov:
hi @bill-filler . thanks for the update. I've already added three tests. Hope that they will be helpful too

hasnain-naveed:
@bill-filler yes, I am working on it.

bill-filler:
This PR is has been combined into https://github.com/edx/edx-ora2/pull/1088.
Will close this and continue on development on that branch

bill-filler:
@dmitry-viskov Thanks again for your contributions. We are in the last stages of testing and hope to merge your PR in the coming days. It's now being tracked here https://github.com/edx/edx-ora2/pull/1088

PR #1061, opened 2017-11-30
fix multi-response training essays

noraiz-anwar:
[EDUCATOR-1847](https://openedx.atlassian.net/browse/EDUCATOR-1847)
This PR is made in effort to fix issue with example responses of multiple prompts. In case of multiple prompts, example responses against these prompts were not showing up on studio and lms.
Further details are on mentioned on ticket.  
### Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [x] Code review: @efischer19 
- [x] Code review: @awaisdar001

PR #1062, opened 2017-12-01
update version number to 2.1.5

noraiz-anwar:
Update ora2 version:
Include fix: https://github.com/edx/edx-ora2/pull/1061
Ticket: https://openedx.atlassian.net/browse/EDUCATOR-1847

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='ora2',
-    version='2.1.4',
+    version='2.1.5',
     author='edX',
     url='http://github.com/edx/edx-ora2',
     description='edx-ora2',


setup.py
```diff
@@ -34,7 +34,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='ora2',
-    version='2.1.4',
+    version='2.1.5',
```
awaisdar001: @noraiz-anwar Have you created the release tag?
https://github.com/edx/edx-ora2/releases/new

noraiz-anwar: Not yet. Is this supposed to be created before merging this version update?

noraiz-anwar: @awaisdar001 I have communicated with @efischer19 to confirm that this PR needs to be merged before so that newly created tag can point to this commit.

PR #1065, opened 2017-12-08
EDUCATOR-1817 | be defensive about serializing training examples, log…

iloveagent57:
… an exception when it occurs.

https://openedx.atlassian.net/browse/EDUCATOR-1817

As I mentioned in the ticket, there's a training example that's actually a list when we're expecting a dict, and it's causing a course export to consistently fail.  This PR aims to be a little more defensive about the types of data structures we might see (as we are elsewhere in ORA) and logs an exception if one occurs in the problematic area of code.

### DIFF
--- a/openassessment/xblock/xml.py
+++ b/openassessment/xblock/xml.py
@@ -2,6 +2,7 @@
 Serialize and deserialize OpenAssessment XBlock content to/from XML.
 """
 from uuid import uuid4 as uuid
+import logging
 
 import dateutil.parser
 import defusedxml.ElementTree as safe_etree
@@ -9,6 +10,9 @@ import lxml.etree as etree
 import pytz
 
 from openassessment.xblock.data_conversion import update_assessments_format
+
+
+log = logging.getLogger(__name__)
 
 
 class UpdateFromXmlError(Exception):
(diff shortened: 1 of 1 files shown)


openassessment/xblock/xml.py
```diff
@@ -606,9 +610,21 @@ def serialize_training_examples(examples, assessment_el):
 
         # Answer provided in the example (default to empty string)
         answer_el = etree.SubElement(example_el, 'answer')
-        for part in example_dict.get('answer', {}).get('parts', []):
-            part_el = etree.SubElement(answer_el, 'part')
-            part_el.text = unicode(part.get('text', u''))
+        try:
+            answer = example_dict.get('answer')
```
awaisdar001: @iloveagent57 We are initiating `parts` var here, Why we are not using it anywhere?

openassessment/xblock/xml.py
```diff
@@ -606,9 +610,21 @@ def serialize_training_examples(examples, assessment_el):
...
+            if answer is None:
+                parts = []
+            elif isinstance(answer, dict):
+                parts = answer.get('parts', [])
+            elif isinstance(answer, list):
+                parse = answer
```
awaisdar001: Did you mean `parts` here?

openassessment/xblock/xml.py
```diff
@@ -606,9 +610,21 @@ def serialize_training_examples(examples, assessment_el):
...
+
+            for part in example_dict.get('answer', {}).get('parts', []):
```
awaisdar001: Is this PR is just for logging? because `example_dict.get('answer', {})` will break again if it returns a `list`

iloveagent57: ah, thank you, that should be `for part in parts:`

openassessment/xblock/xml.py
iloveagent57: geez, yup!

PR #1072, opened 2018-01-23
PLAT-1890 Fix allow_file_upload default value

jmbowman:
The `allow_file_upload` boolean xblock field has been defaulting to `None`, which just gets coerced to `False` on initialization after triggering a `ModifyingEnforceTypeWarning`.  Setting this to the more appropriate `False` value as part of a warning-squashing initiative in `edx-platform`.


jmbowman:
@doctoryes I don't think so; the type coercion happens before the field value is even set in memory, so the assigned `None` never actually gets persisted anywhere.  As far as I can tell, it was just a pointlessly bad default.

PR #1073, opened 2018-01-23
Efischer/group access export

efischer19:
# [EDUCATOR-1156](https://openedx.atlassian.net/browse/EDUCATOR-1156)

### DIFF
--- a/openassessment/xblock/lms_mixin.py
+++ b/openassessment/xblock/lms_mixin.py
@@ -2,7 +2,18 @@
 Fields and methods used by the LMS and Studio.
 """
 
-from xblock.fields import DateTime, Float, Scope, String
+from xblock.fields import DateTime, Dict, Float, Scope, String
+
+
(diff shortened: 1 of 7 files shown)


efischer19:
Tested manually, the exported data now contains:

`<openassessment url_name="b24c33ea35954c7889e1d2944d3fe397" text_response="required" allow_latex="False" group_access="{381451918: [1179773159]}">`

efischer19:
Tests are failing b/c I've addressed *writing* the data when it exists, but it is not currently *read* in any meaningful way by ora code. Will fix in the AM.

efischer19:
This still isn't working, and I'm not sure why. The export goes through just fine, but then the `group_access` setting is just silently dropped on import. I'll continue looking.

efischer19:
@yro, can you take another look here? A few things have changed since you last reviewed, and I can now confirm both export *and import* retain the `group_access` setting correctly.

PR #1074, opened 2018-01-25
Do not try to upload/submit when uploaded file is moved or deleted.

attiyaIshaque:
### Description
This PR fixes that all the files which are moved or deleted could not b uploaded and it throws error and will not save *saved file description* value.

### How to Test?

**stage** 

- Go to: https://courses.stage.edx.org/courses/course-v1:edx+aishaq+2016/courseware/6d17fb1d5881426f9c22043a4927240b/5651d34ad39d4a54b0525f9561655d28/?activate_block_id=block-v1%3Aedx%2Baishaq%2B2016%2Btype%40sequential%2Bblock%405651d34ad39d4a54b0525f9561655d28
- Enter text in text prompt.
- Click 'choose files', select a file. DO NOT click "upload files"
- Enter file description.
- On your computer, delete the chosen file or move it - i.e. make it un-upload-able.
- Click Submit. See pop-up asking whether you want to upload file and confirm submission. Click yes for both.
- See a brief flash of red error message in the platform. Then the message clears and see that your submission has been made without the file.
- As an instructor: Download a .csv for the problem location.
- Observe the submission state contains the 'saved_file_description' of the file that was not uploaded.

**sandbox** 
_*In progress.......*_

### Tests
 - [x] Jasmine Test

### Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [x] @efischer19 
- [ ] @Rabia23 

### Post-review
- [x] Rebase and squash commits

PR #1076, opened 2018-01-30
2.1.11 update js

efischer19:
We forgot to check in updates to the minified js that occurred in https://github.com/edx/edx-ora2/pull/1074, so the changes are not seen in production.

We'll get this released and upgraded in edx-platform ASAP.


efischer19:
For posterity, here are the steps needed to do this in devstack:

0. Clone the edx-ora2 repo into `/src`, which is a sibling of the main `/devstack` folder

(the rest of these are assumed to be run inside the lms container, so `make lms-shell` from devstack before continuing)
1. `pip install -e /edx/src/edx-ora2` (I usually follow this with `pip freeze | grep ora2` to confirm)
2. `cd /edx/src/edx-ora2`
3. `virtualenv edx-ora2`
4. `source edx-ora2/bin/activate`
5. `make install`
6. `make javascript`
7. `make sass`

Now you can leave the lms-shell, go back to your ora2 folder and notice the minified files are included in a `git status`. Check them in and proceed from there.

PR #1080, opened 2018-02-20
add support for linkifying submissions

bill-filler:
Add support for linkifying learner ORA submission through use of custom django tag filter. 

NOTE: django urlize tag was considered but couldn't be used as it doesn't provide a way to set the target attribute of the <a> element and we need to be able to set it to _blank to force opening the link in a new tab. 

This PR currently uses the bleach library to linkify, but others could be considered and swapped in just as easily.

### DIFF
--- a/openassessment/templatetags/oa_extras.py
+++ b/openassessment/templatetags/oa_extras.py
@@ -1,0 +1,20 @@
+from django import template
+from django.template.defaultfilters import linebreaks, stringfilter
+from django.utils.safestring import mark_safe
+from django.utils.html import conditional_escape
+from bleach import callbacks
+
+import bleach
+
+register = template.Library()
+
+
+@register.filter(needs_autoescape=True)
+@stringfilter
+def link_and_linebreak(text, autoescape=True):
+
+    if text and len(text) > 0:
+        the_text = conditional_escape(text)
+        return mark_safe(linebreaks(bleach.linkify(the_text, callbacks=[callbacks.target_blank])))
+    else:
+        return text
--- a/openassessment/templates/openassessmentblock/oa_submission_answer.html
+++ b/openassessment/templates/openassessmentblock/oa_submission_answer.html
@@ -1,5 +1,6 @@
 {% spaceless %}
 {% load i18n %}
+{% load oa_extras %}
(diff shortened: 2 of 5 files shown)


openassessment/templatetags/oa_extras.py
```diff
@@ -0,0 +1,20 @@
...
+register = template.Library()
+
+
+@register.filter(needs_autoescape=True)
+@stringfilter
+def link_and_linebreak(text, autoescape=True):
+
+    if text and len(text) > 0:
+        the_text = conditional_escape(text)
+        return(mark_safe(linebreaks(bleach.linkify(the_text, callbacks=[callbacks.target_blank]))))
```
georgebabey: nit: `return mark_safe(....`

hasnain-naveed:
@bill-filler if everything is good, can you give me a thumbs up?

bill-filler:
@hasnain-naveed 
This looks good to me. Only question I have is if you tested performance with a large submission. From ticket:
"test performance of bleach library when a very large submission is created, potentially use a different library if the performance is not acceptable."
We probably need to get some performance numbers with submissions that are very large (like 10-100 pages worth of text) and see how much overhead bleach adds.

hasnain-naveed:
@bill-filler we have used maximum data which is allowed (roughly 100KB)

With "master" branch

master


With "bfiller/linkify-submission" branch

linkify

bill-filler:
Seems acceptable since the input is limited

bill-filler:
@sandroroux 
just a heads up we merged this and wanted to make sure you were aware and had a chance to raise any objects before we release. please comment if you have any concerns or issues

sandroroux:
@bill-filler This looks great. Awesome work! Just one question: why do we linkify the submission rather than allowing the author to create links in the submission?

bill-filler:
@sandroroux 
The main reason we do it in an automated fashion is because the partner who requested this feature wanted it to be seamless with no extra work needed by the learner. Similar to how gmail works. In the future we could consider having a rich text editor for the submissions, like we are doing for the instructor prompt

sandroroux:
Awesome, sounds good. Thanks for the explanation, Bill.

PR #1083, opened 2018-03-07
Updated travis validate translations

Ayub-Khan:
Updated travis to use the i18n-tools validate
to validate translations in pullrequests.
LEARNER-2556

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,10 @@ compile_translations:
 generate_dummy_translations:
 	i18n_tool dummy
 
+# Test translation files
+validate_translations:
+	cd ./openassessment/ && i18n_tool validate -v
+
 # check if translation files are up-to-date
 detect_changed_source_translations:
 	i18n_tool changed
@@ -60,7 +64,7 @@ push_translations:
 	tx push -s
 
 # extract, compile, and check if translation files are up-to-date
-validate_translations: extract_translations compile_translations generate_dummy_translations detect_changed_source_translations
+check_translations_up_to_date: extract_translations compile_translations generate_dummy_translations detect_changed_source_translations
 
 ################
 #Tests and checks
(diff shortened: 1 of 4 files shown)


amangano-edx:
Please validate that this works by creating a test branch, committing a bad translation to it, and confirming that the build fails.

amangano-edx:
Also, please confirm that the correct version of edx-i18n-tools is installed (>= 0.4.0) in the travis environment.

Ayub-Khan:
I have updated to install latest version of edx_i18n_tools and also exisitng translations are failing which shows its working fine. This is blocked by https://github.com/edx/edx-ora2/pull/1091 which will fix those translations.

Ayub-Khan:
Some other translation issues were fixed in https://github.com/edx/edx-ora2/pull/1096

requirements/base.txt
```diff
@@ -1,4 +1,5 @@
 # edX Internal Requirements
+edx-i18n-tools
```
amangano-edx: You probably want to ensure that the version used is >= 0.4.0, which introduced logic to exit from `validate` with a non-zero code when problems are found.

Ayub-Khan: if we dnt specify version it automatically installs latest.

amangano-edx: Does it do that even when it has a copy of an older version already?

Makefile
```diff
@@ -60,7 +64,7 @@ push_translations:
 	tx push -s
 
 # extract, compile, and check if translation files are up-to-date
-validate_translations: extract_translations compile_translations generate_dummy_translations detect_changed_source_translations
+check_translations_up_to_date: extract_translations compile_translations generate_dummy_translations detect_changed_source_translations test_translations
```
amangano-edx: Should `test_translations` still be in here?

Ayub-Khan: no. sorry my bad.

requirements/base.txt
Ayub-Khan: @amangano-edx Im not sure how travis handles that. But before that it was using 0.3.9 now its using 0.4.3 which can be seen here https://travis-ci.org/edx/edx-ora2/jobs/362037159

PR #1090, opened 2018-03-19
Fix margin and orientation of arrows

bill-filler:
Change the orientation of arrows to be consistent with the rest of the
platform. Add a margin between due-date and header label.

WL-1507

Will look like this:
ora-new-open
ora-new-closed


bill-filler:
@hasnain-naveed could you please give this a review?

PR #1091, opened 2018-03-21
Fix translations

Ayub-Khan:
This pr is related to https://github.com/edx/edx-ora2/pull/1083 and will fix the failing travis test added in that pr.
-Only pull atleast 1% translated locales
-updated header
-added push_translations make target
-removed existing translations
-new translations
-add test command to make sure it works (will remove this after validating)


Makefile
```diff
@@ -65,6 +69,7 @@ quality:
 	./node_modules/jscs/bin/jscs $(STATIC_JS)/src --verbose
 	./scripts/run-pep8.sh
 	./scripts/run-pylint.sh
+	cd ./openassessment/ && i18n_tool validate
```
amangano-edx: Where is the `quality` target used?

Again, as mentioned in the other PRs, I feel this should be a part of the `validate_translations` target.

openassessment/locale/en/LC_MESSAGES/django.po
```diff
@@ -1,18 +1,18 @@
...
+# Copyright (C) 2018 edX
+# This file is distributed under the GNU AFFERO GENERAL PUBLIC LICENSE.
+# EdX Team <info@edx.org>, 2018.
 #
-#, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: PACKAGE VERSION\n"
-"Report-Msgid-Bugs-To: \n"
+"Project-Id-Version: edx-ora2\n"
```
amangano-edx: This is set to `edx-platform` in other files. Do we know how that was set previously? Can we fix them to be consistent?

openassessment/locale/en@lolcat/LC_MESSAGES/djangojs.po
```diff
@@ -1,297 +0,0 @@
-# edX translation file.
```
amangano-edx: Do we know what `en@lolcat` is?

Makefile
Ayub-Khan: i will add this in other pr. using this here just for testing will remove before merging.

efischer19: It's used in `tox.ini` here: https://github.com/edx/edx-ora2/blob/master/tox.ini#L27

`validate_translations` is called in that script as well, so I agree with the recommendation above.

openassessment/locale/en@lolcat/LC_MESSAGES/djangojs.po
Ayub-Khan: From my understanding it seems a temp directory which is no longer used because i was not able to find the language page on transifex. and its not been updated in a long time so we can safely remove those.

openassessment/locale/en/LC_MESSAGES/django.po
Ayub-Khan: I checked the some old commits and it seems it was set to edx-platform before. But as it is set to edx-platform in edx-platform repo should not we set it edx-ora2?.

Makefile
Ayub-Khan: @amangano-edx  @efischer19 i will update other pr according to above suggestions.

openassessment/locale/en/LC_MESSAGES/django.po
amangano-edx: I'm not sure what it should be, as the gettext docs aren't clear (https://www.gnu.org/software/gettext/manual/html_node/Header-Entry.html). If edx-platform has been ok using 'edx-platform', then I have no reason to believe that using 'edx-ora2' wouldn't be ok here.

openassessment/locale/en@lolcat/LC_MESSAGES/djangojs.po
amangano-edx: ok

PR #1092, opened 2018-03-22
Don't install anything but openassessment

nedbat:


nedbat:
@efischer19 What do you think?

PR #1102, opened 2018-04-10
Set proper file extension on download

regisb:
This change is required because file keys no longer refer to the file
name. Unfortunately, we do not retrieve the former file name, just the
extension.
This is possible for the filesystem backend. I haven't managed to do so
for the django-storage backend.


regisb:
Sorry, wrong move!

edx-webhook:
Although this pull request is already closed, I've created [OSPR-5497](https://openedx.atlassian.net/browse/OSPR-5497) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #1216, opened 2018-12-11
DEPR-9 Remove DataDog support

jmbowman:
Removing DataDog event logging as per [DEPR-9](https://openedx.atlassian.net/browse/DEPR-9).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ sudo: false
 python:
   - "2.7"
 env:
-  - TOXENV=django18
   - TOXENV=django111
 matrix:
   include:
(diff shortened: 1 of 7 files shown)


jmbowman:
I had to make a couple more changes to get tests to pass:

* Make a fix for boto/Travis brokenness derived from the fix @michaelyoungstrom just [made in edx-certificates](https://github.com/edx/edx-certificates/pull/58)
* Change a query count to account for less work now being done without those dogapi calls

I also dropped testing for Django 1.8 and explictly stated our supported Django version in the Trove classifiers while I was at it.

PR #1224, opened 2019-02-07
Fix file upload for S3 regions requiring Signature version 4

silviot:
Passing the `host` parameter to boto's `connect_s3` method solves the issue and should not introduce new problems. It was suggested [on StackOverflow](https://stackoverflow.com/a/29451280/41387).

Without this fix we get the following error when using an S3 bucket from `eu-central-1` region:

```
File "/edx/app/edxapp/venvs/edxapp/src/ora2/openassessment/fileupload/api.py" in remove_file
  29.     return backends.get_backend().remove_file(key)
File "/edx/app/edxapp/venvs/edxapp/src/ora2/openassessment/fileupload/backends/s3.py" in remove_file
  49.         bucket = conn.get_bucket(bucket_name)
File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/boto/s3/connection.py" in get_bucket
  503.             return self.head_bucket(bucket_name, headers=headers)
File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/boto/s3/connection.py" in head_bucket
  550.                 response.status, response.reason, body)
Exception Value: S3ResponseError: 400 Bad Request
```

PR #1228, opened 2019-03-15
Revert "Fix file upload for S3 regions requiring Signature version 4"

ormsbee:
Reverts edx/edx-ora2#1224

Because this breaks file uploads in our system.


ormsbee:
@nedbat: My plan is to get this revert in, create a new PR bumping the version of ORA2 another minor point, then bring that into edx-platform for Ironwood.

nedbat:
OK, it will be in ironwood.2 I think.

PR #1248, opened 2019-07-02
added logs for information in case of FileUploadError

DawoudSheraz:
### Description
In edx-ora2, we have seen a few instances where for a submission, the file uploads run into some sort of error. We were only able to find out this information from the courseware_studentmodule table, where the state column had the information about the uploaded file description. Currently, our codebase isn't doing anything in case of FileUploadError. Since our initial investigation hinted towards this exception, a simple piece of logging has been added in this PR to get more information, if we were to see such a case in the future.

### DIFF
--- a/openassessment/xblock/submission_mixin.py
+++ b/openassessment/xblock/submission_mixin.py
@@ -250,7 +250,15 @@ class SubmissionMixin(object):
                         except IndexError:
                             pass
                 except FileUploadError:
-                    pass
+                    logger.exception(
+                        u"FileUploadError for student_item: {student_item_dict}"
+                        u" and submission data: {student_sub_data} with file"
+                        "descriptions {files_descriptions}".format(
+                            student_item_dict=student_item_dict,
+                            student_sub_data=student_sub_data,
+                            files_descriptions=files_descriptions 
+                        )
+                    )
                 if key_to_save:
                     student_sub_dict['file_keys'].append(key_to_save)
                     student_sub_dict['files_descriptions'].append(file_description)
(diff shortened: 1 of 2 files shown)


openassessment/xblock/submission_mixin.py
```diff
@@ -250,6 +250,13 @@ def create_submission(self, student_item_dict, student_sub_data, files_descripti
                         except IndexError:
                             pass
                 except FileUploadError:
+                    logger.exception(
+                        u"FileUploadError for {student_item_dict} with {url} and {student_sub_data}".format(
+                            student_item_dict=student_item_dict,
```
awaisdar001:
```
url=url,
student_item_dict=student_item_dict,
student_sub_data=student_sub_data,
```

DawoudSheraz:
Here are the example logs:

1. This will tell the file name in case of FileUploadError
`2019-07-08 09:04:49,653 ERROR 19360 [openassessment.xblock.submission_mixin] [user 2] submission_mixin.py:322 - FileUploadError:Error retrieving upload URL for the data:{u'contentType': u'text/plain', u'filenum': 0, u'filename': u'test.txt'}.`

2. This would show the file descriptions and data
`2019-07-08 09:12:45,034 ERROR 19911 [openassessment.xblock.submission_mixin] [user 2] submission_mixin.py:260 - FileUploadError for student_item: {'course_id': u'course-v1:ed999+ed989+2019_T2', 'item_type': 'openassessment', 'item_id': u'block-v1:ed999+ed989+2019_T2+type@openassessment+block@0633fd77bb9c4557a5a08902e6789af7', 'student_id': '1754fb1a7c65075470a8965ad704ca28'} and submission data: [u'Respo'] with filedescriptions [u'ABC']
`

DawoudSheraz:
@awaisdar001 Please take a look at the sample logs.

awaisdar001:
I can't understand where we are logging `user 2`? We shouldn't log the db keys

DawoudSheraz:
> I can't understand where we are logging `user 2`? We shouldn't log the db keys

It is not being logged. I copied the entire line of the terminal when the log was hit.

awaisdar001:
> > I can't understand where we are logging `user 2`? We shouldn't log the db keys
> 
> It is not being logged. I copied the entire line of the terminal when the log was hit.

got it.

PR #1260, opened 2019-09-09
 Increase assignment upload limit from 10MB to 20MB

kdmccormick:
@edx/masters-devs @dabdul-hathi
Ticket: https://openedx.atlassian.net/browse/EDUCATOR-4635

This is another attempt at https://github.com/edx/edx-platform/pull/21589.

Also see https://github.com/edx/configuration/pull/5399, which ups the limit in NGINX.


openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -41,8 +41,11 @@ OpenAssessment.ResponseView.prototype = {
     // before we can autosave.
     AUTO_SAVE_WAIT: 30000,
 
-    // Maximum size (10 MB) for all attached files.
-    MAX_FILES_SIZE: 10485760,
+    // Maximum size (2^20 bytes, approx. 20MB) for all attached files.
```
iloveagent57: nit: it's actually `20 * 2^20`

kdmccormick: Doh!

PR #1263, opened 2019-09-13
jscs -> eslint

matthugs:
JSCS is very old and is the source of some audit noise. So let's replace it! We should probably also follow up and switch over to https://github.com/edx/eslint-config-edx/tree/master/packages/eslint-config-edx-es5, but for the scope of this I just limited myself to what was necessary to remove the dependency.


matthugs:
@iloveagent57 I absolutely could! The most monstrous commit of all of these was this one https://github.com/edx/edx-ora2/pull/1263/commits/47723d30d4fc9028c42e5da127caf50f0fbb9bcb

PR #1268, opened 2019-10-10
BOM-415

awais786:
- Enabling unit tests on travis. 11 Units tests were failing on master branch fixed all.
- Fixing [BOM-415](https://openedx.atlassian.net/browse/BOM-415). edx-platform was giving error in python3.
- Bump the version.

### DIFF
--- a/openassessment/assessment/models/base.py
+++ b/openassessment/assessment/models/base.py
@@ -109,7 +109,7 @@ class Rubric(models.Model):
         rubric_dict.pop("content_hash", None)
 
         canonical_form = json.dumps(rubric_dict, sort_keys=True)
-        return sha1(canonical_form).hexdigest()
+        return sha1(canonical_form.encode('utf-8')).hexdigest()
 
     @staticmethod
     def structure_hash_from_dict(rubric_dict):
@@ -141,7 +141,7 @@ class Rubric(models.Model):
             for criterion in rubric_dict.get('criteria', [])
         ]
         canonical_form = json.dumps(structure, sort_keys=True)
-        return sha1(canonical_form).hexdigest()
+        return sha1(canonical_form.encode('utf-8')).hexdigest()
 
 
 class Criterion(models.Model):
--- a/openassessment/assessment/models/training.py
+++ b/openassessment/assessment/models/training.py
@@ -137,7 +137,7 @@ class TrainingExample(models.Model):
             'options_selected': options_selected,
             'rubric': rubric.id
         })
-        return sha1(contents).hexdigest()
+        return sha1(contents.encode('utf-8')).hexdigest()
 
     @classmethod
     def cache_key(cls, answer, options_selected, rubric):
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
 matrix:
   include:
     - python: 2.7
+      env: TOXENV=django111
+    - python: 2.7
       env: TOXENV=quality
     - python: 2.7
       env: TOXENV=js
--- a/openassessment/xblock/staff_area_mixin.py
+++ b/openassessment/xblock/staff_area_mixin.py
@@ -391,7 +391,7 @@ class StaffAreaMixin(object):
             'score': workflow.get('score'),
             'workflow_status': workflow.get('status'),
             'workflow_cancellation': workflow_cancellation,
-            'are_grades_frozen': grade_utils.are_grades_frozen()
+            'are_grades_frozen': grade_utils.are_grades_frozen() if grade_utils else None
         })
 
         if peer_assessments or self_assessment or staff_assessment:
--- a/openassessment/xblock/test/test_grade.py
+++ b/openassessment/xblock/test/test_grade.py
@@ -254,7 +254,7 @@ class TestGrade(XBlockHandlerTestCase, SubmitAssessmentsMixin):
 
         # Create all but the last peer assessment of the current user; no peer grade will be available
         graded_by = xblock.get_assessment_module('peer-assessment')['must_be_graded_by']
-        for scorer_sub, scorer_name, assessment in zip(scorer_subs, self.PEERS, PEER_ASSESSMENTS)[:-1]:
+        for scorer_sub, scorer_name, assessment in list(zip(scorer_subs, self.PEERS, PEER_ASSESSMENTS))[:-1]:
             self.create_peer_assessment(
                 scorer_sub,
                 scorer_name,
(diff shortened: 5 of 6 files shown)


awais786:
@feanil @nasthagiri units tests were not running on this repo previously . I just enabled them and there are lots of failures. On my local I have 300's failures due to the following issue. But travis does not show this issue.  @adeelkhan is working on other [repo](https://github.com/open-craft/xblock-poll) and facing same following issue.
```
ERROR: test_unavailable (openassessment.xblock.test.test_leaderboard.TestLeaderboardRender)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/arbisoft/Documents/devstack/edx-ora2/openassessment/xblock/test/base.py", line 114, in _wrapped
    xblock = self.load_scenario(scenario_path)
  File "/Users/arbisoft/Documents/devstack/edx-ora2/openassessment/xblock/test/base.py", line 163, in load_scenario
    self.load_fixture_str(xml_path), self.runtime.id_generator
  File "/Users/arbisoft/Documents/ves/edx-ora2_env_27/lib/python2.7/site-packages/xblock/runtime.py", line 683, in parse_xml_string
    return self.parse_xml_file(io_type(xml), id_generator)
  File "/Users/arbisoft/Documents/ves/edx-ora2_env_27/lib/python2.7/site-packages/xblock/runtime.py", line 688, in parse_xml_file
    usage_id = self._usage_id_from_node(root, None, id_generator)
  File "/Users/arbisoft/Documents/ves/edx-ora2_env_27/lib/python2.7/site-packages/xblock/runtime.py", line 717, in _usage_id_from_node
    block_class = self.mixologist.mix(self.load_block_type(block_type))
  File "/Users/arbisoft/Documents/ves/edx-ora2_env_27/lib/python2.7/site-packages/xblock/runtime.py", line 604, in load_block_type
    return XBlock.load_class(block_type, self.default_class, self.select)
  File "/Users/arbisoft/Documents/ves/edx-ora2_env_27/lib/python2.7/site-packages/xblock/plugin.py", line 114, in load_class
    selected_entry_point = select(identifier, all_entry_points)
  File "/Users/arbisoft/Documents/ves/edx-ora2_env_27/lib/python2.7/site-packages/xblock/plugin.py", line 40, in default_select
    raise PluginMissingError(identifier)
PluginMissingError: openassessment
-------------------- >> begin captured stdout << ---------------------
Loading scenario from data/leaderboard_unavailable.xml
```

feanil:
Looks like there are only 11 failures in CI, what's the difference is there some setup you're missing?  Can you try to fix the 11 errors we're seeing in CI?

openassessment/assessment/models/base.py
```diff
@@ -109,7 +109,7 @@ def content_hash_from_dict(rubric_dict):
         rubric_dict.pop("content_hash", None)
 
         canonical_form = json.dumps(rubric_dict, sort_keys=True)
-        return sha1(canonical_form).hexdigest()
+        return sha1(canonical_form.encode('utf-8')).hexdigest()
```
awais786: Need this change for this ticket
https://openedx.atlassian.net/browse/BOM-415
We have few other places also in this repo where sha1 is calling without encode. I just updated this for time being.

openassessment/assessment/models/training.py
```diff
@@ -137,7 +137,7 @@ def calculate_hash(answer, options_selected, rubric):
             'options_selected': options_selected,
             'rubric': rubric.id
         })
-        return sha1(contents).hexdigest()
+        return sha1(contents.encode('utf-8')).hexdigest()
```
awais786: Although this change not require for bom-415 but i did it to avoid any future issue. Please let me know if not require.

openassessment/assessment/models/base.py
```diff
@@ -141,7 +141,7 @@ def structure_hash_from_dict(rubric_dict):
             for criterion in rubric_dict.get('criteria', [])
         ]
         canonical_form = json.dumps(structure, sort_keys=True)
-        return sha1(canonical_form).hexdigest()
+        return sha1(canonical_form.encode('utf-8')).hexdigest()
```
awais786: Although this change not require for bom-415 but i did it to avoid any future issue. Please let me know if not require.

awais786:
@feanil Please review this PR.

openassessment/assessment/models/base.py
noraiz-anwar: Do we have made sure that `canonical_form` doesn't contain any non ASCII characters?
For example:
`'å®ßisøƒ†'.encode("utf-8")` would fail on Python 2.

noraiz-anwar: The same question for other places where `.encode("utf-8")` is used.

awais786: @noraiz-anwar 

Python's json module simply takes the safe route and escapes non-ascii characters. This is a valid way of representing such characters in json, and any conforming parser will resurrect the proper Unicode characters when parsing the string:
```
import json
json.dumps({'Crêpes': 5})
'{"Cr\\u00eapes": 5}'
json.loads('{"Cr\\u00eapes": 5}')
{'Crêpes': 5}
```
https://stackoverflow.com/questions/35582528/python-encoding-and-json-dumps

Ayub-Khan: @noraiz-anwar Its highly unlikely that we will have ASCII characters in canonical_form as that is the rubric data. But @awais786 to be on safe side you can add the check 
```
if not isinstance(canonical_form, bytes):
    canonical_form = canonical_form.encode('utf-8')
```
before the sha1 call.

openassessment/assessment/models/base.py
Ayub-Khan: same as above.

openassessment/assessment/models/base.py
noraiz-anwar: That is interesting. So, basically we need bytes for `.hexdigest()` and for that, we are doing `.encode("utf-8")`

noraiz-anwar:
@Ayub-Khan thanks for the input.
`.encode('utf-8')` will be a no-op in python 2. Encoding bytes would simply return bytes.
So, I believe we can simply skip `if not isinstance(canonical_form, bytes):` condition to avoid an extra check that we will be eventually removing while completely moving on python 3.

openassessment/assessment/models/base.py
Ayub-Khan: @noraiz-anwar yes.

PR #1273, opened 2019-10-25
BOM Project Python3 Compatibility

Ayub-Khan:
-Added py3 tox and travis tests
-Test fixes for python3
-Bump xblock-sdk version
-Removed depricated pep8 infavor of pycodestyle
-Fixed Quality


Ayub-Khan:
if we modernize workbench lib this fixes all tests for py3.

openassessment/assessment/test/test_training.py
```diff
@@ -163,20 +163,21 @@ def test_similar_training_examples_different_answer(self):
...
-        mock_example = mock.MagicMock(TrainingExample)
-        mock_get.side_effect = [TrainingExample.DoesNotExist, mock_example]
-        mock_create.side_effect = IntegrityError
-
-        # Expect that we get the mock example back
-        # (proves that the function tried to retrieve the object again after
-        # catching the integrity error)
-        examples = deserialize_training_examples(self.EXAMPLES[:1], self.RUBRIC)
-        self.assertEqual(examples, [mock_example])
+    if six.PY2:
```
feanil: Why are we not running this test in python 3?

openassessment/fileupload/views_filesystem.py
```diff
@@ -21,7 +21,7 @@ def filesystem_storage(request, key):
     """
     Uploading and download files to the local filesystem backend.
     """
-    if isinstance(key, six.text_type):
+    if six.PY2 and isinstance(key, six.text_type):
```
feanil: Why do we need to only do this in Py2, should this be safe in both py2 and py3?

Ayub-Khan:
fixing quality caused some test failures will fix them tomorrow. 
xblock-sdk was cusing some issues made this pr for those.

openassessment/assessment/test/test_training.py
Ayub-Khan: This raises a pickle deserialization issues for mock objects which is a known py3 issue.

openassessment/fileupload/views_filesystem.py
Ayub-Khan: This is used for file handling. this is used to get the path and open files. open works for both bytes and unicode type path in both py2 and py3 but in py3 if we keep it bytes we will have to change on a lot of other place and make all other strings bytes which are hardcoded and unicode by default.

openassessment/assessment/test/test_training.py
feanil: Can we fix that by changing the type of object our mock is returning so that it's serializable.  I'd like to be able to keep this test in python 3.

openassessment/fileupload/views_filesystem.py
feanil: In that case can we just always leave this string as unicode and not do this conversion?

openassessment/management/tests/test_collect_ora2_data.py
```diff
@@ -69,14 +69,14 @@ def setUp(self):
...
-            '{""text"": ""This is a response to a question. #dylan""}',
+            u'{""text"": ""This is a response to a question. #dylan""}',
             "Assessment #1 -- scored_at: 2014-10-07 20:37:54 -- type: T -- scorer_id: hash -- feedback: Test",
             "Assessment #1 -- Content: Unclear recommendation (5)",
             "2014-10-07 21:35:47+00:00",
             "10",
             "20",
             "Completed test assessments.",
-            "\xf0\x9d\x93\xa8\xf0\x9d\x93\xb8\xf0\x9d\x93\xbe",
+            "𝓨𝓸𝓾",
```
feanil: Should all these strings be prefixed with a `u` instead of just the ones you have?

openassessment/xblock/test/test_data_conversion.py
```diff
@@ -22,7 +22,7 @@ class DataConversionTest(TestCase):
         ('[{"description": "Test prompt."}]', [{'description': 'Test prompt.'}]),
     )
     @ddt.unpack
-    def test_create_prompts_list(self, input, output):
+    def test_create_prompts_list(self, input, output):  # pylint: disable=redefined-builtin
```
iloveagent57: Nit: rename `input` to `prompt_input` (or similar).

openassessment/management/tests/test_collect_ora2_data.py
Ayub-Khan: thats submission api data. user inputs feedback, and raw_answer could be unicode but changing to other data causes problems.

openassessment/assessment/test/test_assessment_models.py
```diff
@@ -181,13 +181,13 @@ def test_check_all_criteria_assessed(self, data):
...
-        self.assertTrue(data['expected_error'] == error)
+        self.assertEqual(data['expected_error'], error)
 
-    def _create_data_structures_with_criterion_properties(
-        self,
-        has_option_selected=True,
-        has_zero_options=True,
-        has_feedback=True
+    def _create_data_structures_with_criterion_properties(  # pylint: disable=invalid-name
+            self,
```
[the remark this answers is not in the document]
Ayub-Khan: linter : For readability.

PR #1275, opened 2019-10-25
EDUCATOR 4751: disallow uploading files with empty/missing descriptions

nsprenkle:
Resolves [EDUCATOR-4751](https://openedx.atlassian.net/browse/EDUCATOR-4751): learners can upload files to ORA blocks without descriptions.

How? Disable/enable logic for submit button now additionally checks that files pending upload contain non-empty descriptions.

FYI: @edx/masters-devs

### DIFF
--- a/openassessment/xblock/static/js/spec/lms/oa_response.js
+++ b/openassessment/xblock/static/js/spec/lms/oa_response.js
@@ -627,7 +627,7 @@ describe("OpenAssessment.ResponseView", function() {
             return $(view.element).find('.file__upload').first();
         }
         function expectFileUploadButton(disabled) {
-            view.checkFilesDescriptions();
+            view.collectFilesDescriptions();
             expect(getFileUploadField().is(':disabled')).toEqual(disabled);
         }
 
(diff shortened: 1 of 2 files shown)


openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -707,10 +721,13 @@ OpenAssessment.ResponseView.prototype = {
             }
         });
 
-        $(this.element).find('.file__upload').prop('disabled', isError);
+        this.uploadEnabled(!isError);
```
Andytr1: - Please update the comment to reflect the return value.
- Now that the function returns a value, 'checkFilesDescriptions' name seems a bit odd. Perhaps 'filesDescriptionsValid' would be a better name?

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -256,6 +262,10 @@ OpenAssessment.ResponseView.prototype = {
         return this.baseView.buttonEnabled('.submission__save', enabled);
     },
 
+    uploadEnabled: function(enabled) {
```
Andytr1: nit: the name is off putting to me because we are use this function to toggle functionality. I would suggest 'toggleEnableUpload' .

iloveagent57: This function can act as a getter or a setter, depending on the number of args provided, much like `saveEnabled()` above (or lots of other JS functions in this project) - 
```js
view.uploadEnabled(true); // enables the upload button.
view.uploadEnabled(); // checks whether the upload button is enabled.
```

openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -715,7 +715,42 @@ describe("OpenAssessment.ResponseView", function() {
          expect(view.hasAllUploadFiles()).toEqual(false);
          expect(view.baseView.toggleActionError).toHaveBeenCalledWith('upload',
             "Your file " + file[0].name + " has been deleted or path has been changed.");
+     });
 
+    it("prevents user from uploading or responses when file descriptions are missing", function() {
```
iloveagent57: Really minor nit: "prevents user from uploading or _submitting_..."

openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -715,7 +715,42 @@ describe("OpenAssessment.ResponseView", function() {
...
+        expect(view.submitEnabled()).toBe(false);
+        expect(view.uploadEnabled()).toBe(false);
+
+        $(firstDescriptionField1).val('test1');
+        $(firstDescriptionField2).val('test2');
 
+        // user finishes adding descriptions which enables submit button
+        view.checkSubmissionAbility(true);
+        expect(view.submitEnabled()).toBe(true);
+        expect(view.uploadEnabled()).toBe(true);
```
iloveagent57: Nice test, very thorough...

openassessment/xblock/static/js/src/lms/oa_response.js
iloveagent57: In addition to checking validity, it also has the side-effect of enabling the upload button if the descriptions are valid and storing the file description values in `this.filesDescriptions`.  `checkFileDescriptions` and `fileDescriptionsValid` basically connote the same idea ("I'm going to return false if something is wrong with the descriptions").  Maybe `collectFileDescriptions`?  When it returns `true`, they will have been collected, and when it returns `false`, nothing will have been collected.

PR #1279, opened 2019-10-30
add delete button for uploaded files

jansenk:
Add delete button to downloaded files so that they can be deleted individually.

Files with the delete buttons
The middle two files have been deleted
Once the files have been submitted, they can't be deleted


openassessment/xblock/static/js/src/oa_server.js
```diff
@@ -531,6 +531,26 @@ if (typeof OpenAssessment.Server === "undefined" || !OpenAssessment.Server) {
...
+        /**
+         * Sends request to server to remove specific uploaded file.
+         */
+        removeUploadedFile: function(filenum) {
+            var url = this.url('remove_uploaded_file');
+            return $.Deferred(function(defer) {
+                $.ajax({
+                    type: "POST",
+                    url: url,
+                    data: JSON.stringify({'filenum': filenum}),
```
Andytr1: nit: what about calling this file fileId, fileIndex or fileNumber (spelled out)

openassessment/xblock/submission_mixin.py
```diff
@@ -370,6 +370,29 @@ def remove_all_uploaded_files(self, data, suffix=''):  # pylint: disable=unused-
...
+            filenum = -1
+
+        removed = file_upload_api.remove_file(self._get_student_item_key(filenum))
+        if removed:
+            saved_file_descriptions = json.loads(self.saved_files_descriptions)
+            saved_file_descriptions[filenum] = None
+            self.saved_files_descriptions = json.dumps(saved_file_descriptions)
+
+            saved_files_names = json.loads(self.saved_files_names)
+            saved_files_names[filenum] = None
```
Andytr1: It appears that we are simply assigning the value of None to the array position that corresponds to the removed file index. However, it would be better to remote that element from the array altogether:

```
saved_file_descriptions.pop(filenum)
saved_files_names.pop(filenum)

```

openassessment/xblock/static/js/src/oa_server.js
jansenk: filenum is used other places in this file so I just wanted to be consistent

jansenk: as well as in other files in this general workflow

openassessment/xblock/submission_mixin.py
jansenk: We would then have to adjust the URLs of all of the remaining elements, replacing it with None lets us just pass over it and ignore it without having to modify the other files in any way

openassessment/xblock/submission_mixin.py
```diff
@@ -552,19 +575,28 @@ def submission_path_and_context(self):
...
+                            file_name = saved_files_names[i]
+                        except IndexError:
+                            pass
+                        file_urls.append((file_url, file_description, file_name))
+                    else:
+                        break
+            else:
+                for i, (file_description, file_name) in enumerate(zip(saved_files_descriptions, saved_files_names)):
+                    if file_description is None:
+                        file_urls.append((None, None, None))
```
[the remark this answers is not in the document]
iloveagent57: Oh, wait, yeah, we have to pass it to the front-end so that we can later use the correct index if we want to delete the file, right?  If that's the case, would you mind adding a comment here to explain that necessity?

openassessment/templates/openassessmentblock/oa_uploaded_file.html
```diff
@@ -13,21 +13,28 @@ <h5 class="{{ class_prefix }}__display__title">
...
+                {% elif file_upload_type == "pdf-and-image" or file_upload_type == "custom" %}
+                    <a href="{{ file_url }}" class="submission__answer__file submission--file" target="_blank">
+                        {% if file_description %}
+                        {{ file_description }} ( {{file_name}} )
+                        {% else %}
+                        {% trans "View the files associated with this submission:" %} #{{ forloop.counter }}
+                        {% endif %}
+                    </a>
+                {% endif %}
+                {% if submit_enabled %}
```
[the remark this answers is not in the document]
jansenk: This template is rendered in many different contexts, including post-submission, the case where the assignment is closed, peer grading, leaderboards, etc, so we need to make sure we are only showing it when the submission is open, and the student has not yet submitted their answer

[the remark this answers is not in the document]
iloveagent57: So if there's a required text response for the block, will the delete buttons appear for me if I have not entered anything into the text box?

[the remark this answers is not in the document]
jansenk: Oh, you're right. I'll have to change how I do that

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -767,11 +779,28 @@ OpenAssessment.ResponseView.prototype = {
...
 
+    /**
+     Remove a previously uploaded file.
+
+     */
+    removeUploadedFile: function(filenum) {
+        var view = this;
+        return this.server.removeUploadedFile(filenum).done(function() {
+            var sel = $('.step--response', view.element);
+            var block = sel.find('.submission__answer__file__block__'+filenum);
```
Andytr1: trivial nit: spaces before and after '+'

openassessment/xblock/submission_mixin.py
```diff
@@ -370,6 +370,29 @@ def remove_all_uploaded_files(self, data, suffix=''):  # pylint: disable=unused-
...
+        if removed:
+            saved_file_descriptions = json.loads(self.saved_files_descriptions)
+            saved_file_descriptions[filenum] = None
+            self.saved_files_descriptions = json.dumps(saved_file_descriptions)
+
+            saved_files_names = json.loads(self.saved_files_names)
+            saved_files_names[filenum] = None
+            self.saved_files_names = json.dumps(saved_files_names)
+
+        return {'success': removed, 'removed_num': 1 if removed else 0}
```
Andytr1: nit: This function always removes a single file, what is the value in returning "removed_num" ? Why not just return success?

PR #1282, opened 2019-11-01
EDUCATOR-4724 upload multiple files

Andytr1:
Allow users to upload new files without deleting existing ones. The maximum file limit is set for 20 for now.


openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -644,6 +636,15 @@ OpenAssessment.ResponseView.prototype = {
             }
         }
 
+        if (this.getSavedFileCount() + files.length > this.data.MAXIMUM_FILE_UPLOAD_COUNT ) {
```
jansenk: `$('.submission__answer__file__block').length;` gets you the count of all files, deleted and not, which is useful for keeping track of indices, but I think you need to make this `$('.submission__answer__file__block').filter(':parent').length` for checking MAX_FILES

Andytr1: How is it different?

jansenk: yeah, we're leaving an empty div for deleted files. `:parent` only includes non-empty elements

openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -583,6 +580,65 @@ describe("OpenAssessment.ResponseView", function() {
...
+    });
+
+    it("tests that file upload number can't exceed maximum", function() {
+        spyOn(view.baseView, 'toggleActionError').and.callThrough();
+        spyOn(view, 'getSavedFileCount').and.returnValue(20);
+        view.textResponse = 'optional';
+        view.fileUploadResponse = 'required';
+        expect(view.submitEnabled()).toBe(false);
+        // Upload files
+        var files = [{type: 'image/jpeg', size: 1024, name: 'picture1.jpg', data: ''},
```
iloveagent57: :) This can be created more efficiently with a for-loop.

openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -583,6 +580,65 @@ describe("OpenAssessment.ResponseView", function() {
...
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''},
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''},
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''},
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''},
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''},
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''},
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''},
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''},
+                     {type: 'image/jpeg', size: 1024, name: 'picture2.jpg', data: ''}];
+        view.prepareUpload(files, 'image', ['i1', 'i2','i1', 'i2','i1', 'i2','i1', 'i2','i1', 'i2','i1', 'i2',
```
iloveagent57: same here

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -928,7 +916,8 @@ OpenAssessment.ResponseView.prototype = {
             } else {
                 file = $('<a />', {
                     href: url,
-                    text: view.filesDescriptions[filenum] + ' (' + view.fileNames[filenum] + ')',
+                    text: view.filesDescriptions[filenum - view.fileCountBeforeUpload] +
```
iloveagent57: Nit: this is probably easier to read as:
```js
var description = view.filesDescriptions[filenum - view.fileCountBeforeUpload];
var fileName = view.fileNames[filenum - view.fileCountBeforeUpload];
...
text: description + ' (' + filename + ')';
```

openassessment/xblock/static/js/src/lms/oa_response.js
Andytr1: done

openassessment/xblock/static/js/spec/lms/oa_response.js
Andytr1: copy and paste got away from me :)

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -836,14 +825,13 @@ OpenAssessment.ResponseView.prototype = {
 
         sel.find('.file__upload').prop('disabled', true);
 
-        promise = view.removeUploadedFiles();
-        promise = promise.then(function() {
-            return view.saveFilesDescriptions();
-        });
+        promise = view.saveFilesDescriptions();
 
+        view.fileCountBeforeUpload = view.getSavedFileCount();
```
jansenk: well, so here I believe you do need the total count of all files, not just the current count of non-deleted files

Andytr1: I disagree. Do you want to chat?

Andytr1: since we limit the user based on the number of "real"/non deleted files - it doesn't make sense to take deleted files into account.

Andytr1: Per our conversation, changes made.

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -766,22 +767,11 @@ OpenAssessment.ResponseView.prototype = {
     },
 
     /**
-     Remove previously uploaded files.
-
+     Returns the number of file blocks rendered on the page
```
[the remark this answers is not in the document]
Andytr1: done

openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -583,6 +581,76 @@ describe("OpenAssessment.ResponseView", function() {
...
+        // Delete the first file
+        view.removeUploadedFile(0);
+        view.checkSubmissionAbility(true);
+        expect(view.submitEnabled()).toBe(true);
+        files = [{type: 'image/jpeg', size: 1024, name: 'picture3.jpg', data: ''}];
+        view.prepareUpload(files, 'image', ['i3']);
+        view.uploadFiles();
+        view.checkSubmissionAbility(true);
+        expect(view.fileNames.length).toEqual(1);
+        expect(view.filesDescriptions.length).toEqual(1);
```
iloveagent57: This doesn't assert anything about the state of the previously uploaded files.  The issue we ran into during manual testing wasn't that uploads didn't work, it's that the file description of the newly uploaded file was something unexpected.

Andytr1: The root cause of the issue was that the fileNames collection wasn't being cleared and we were appending elements to it. The fix in the code, ensured that the fileNames collection is cleared after each upload and its length should be the same as fileDescriptions collection which is the same as the number of files selected for upload. Hence my approach to testing it this way. 
Another reason for this approach: we "spy" on the saveFilesDescriptions method - which is where the fileNames collection is supposed to be . populated. I feel this testing is sufficient and full fills the purpose.

iloveagent57: OK, that's a fair enough argument.  Could you leave a comment to the effect of
```ensure that the fileNames collection is cleared after each upload, and that its length is the same as the fileDescriptions collection (and the number of files selected for upload). ```
before these two assertions?

Comment (Andytr1) [reply openassessment/xblock/static/js/spec/lms/oa_response.js]: UT updated

Comment (iloveagent57) [reply openassessment/xblock/static/js/spec/lms/oa_response.js]: How about
```js
// Ensure that view.fileNames and view.fileDescriptions only contain data about the newest
// set of uploaded files.
expect(view.fileNames).toEqual(['picture3.jpg']);
expect(view.filesDescriptions).toEqual(['i3']);
```

PR #1283, opened 2019-11-04
Quality enhancements

Ayub-Khan:
-Used pylintrc_tweaks for modifications 
-updated pylintrc
-fixed quality accordingly
-enabled missing-docstring pylint check
-added missing docstrings to fix quality
-bump-version
-bump xblock-sdk version

Remaining changes of https://github.com/edx/edx-ora2/pull/1273


requirements/base.txt
```diff
@@ -3,7 +3,7 @@ edx-i18n-tools
 edx-submissions>=2.0.12
 git+https://github.com/edx/django-rest-framework.git@1ceda7c086fddffd1c440cc86856441bbf0bd9cb#egg=djangorestframework==3.6.3
 git+https://github.com/edx/XBlock.git@xblock-1.0.1#egg=XBlock==1.0.1
-git+https://github.com/edx/xblock-sdk.git@8918a3f1d2d4fae3a806a29d1fee9428f4dbe37f#egg=xblock-sdk==0.1.5
+git+https://github.com/edx/xblock-sdk.git@v0.1.6#egg=xblock-sdk==0.1.6
```
feanil: Even if there are tags, we should try to uses hashes here for security. Since tags can still be moved.  You should just use the hash of the commit that matches this tag here.

PR #1287, opened 2019-11-06
Jkantor/size2

jansenk:
track file size


openassessment/templates/openassessmentblock/oa_uploaded_file.html
```diff
@@ -12,8 +12,10 @@ <h5 class="{{ class_prefix }}__display__title">
 
     <div class="{{ class_prefix }}__display__file {% if not file_urls %}is--hidden{% endif %} submission__{{ file_upload_type }}__upload" data-upload-type="{{ file_upload_type }}">
         <div class="submission__answer__files">
-        {% for file_url, file_description, file_name in file_urls %}
-            <div class="submission__answer__file__block submission__answer__file__block__{{ forloop.counter0 }}" {% if not file_url %} deleted {% endif %}>
+        {% for file_url, file_description, file_name, file_size in file_urls %}
```
Andytr1: Do we want to display file size to the user?

jansenk: I think we should, I also have it down at line 34

openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -506,7 +506,7 @@ describe("OpenAssessment.ResponseView", function() {
         var files = [{type: 'image/jpeg', size: 22000000, name: 'huge-picture.jpg', data: ''}];
         view.prepareUpload(files, 'image');
         expect(view.baseView.toggleActionError).toHaveBeenCalledWith('upload',
-            'File size must be 20MB or less.');
+            'Total file upload size must be 20MB or less.');
```
Andytr1: I think this case is no longer relevant.

openassessment/templates/openassessmentblock/oa_uploaded_file.html
Andytr1: I may have missed it, but the business requirements don't ask for it. Also, what is the value of that since we are only limiting the user to 500MB / single file. There is no cumulative limit.

jansenk: Oh, I totally forgot about that. You're right.

openassessment/xblock/static/js/spec/lms/oa_response.js
jansenk: Yeah, you're right. I'll also have to change some of the other validation

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -584,20 +584,18 @@ OpenAssessment.ResponseView.prototype = {
...
         var fileType = null;
         var errorCheckerTriggered = false;
 
         for (var i = 0; i < files.length; i++) {
-            totalSize += files[i].size;
             ext = files[i].name.split('.').pop().toLowerCase();
             fileType = files[i].type;
 
-            if (totalSize > this.MAX_FILES_SIZE) {
+            if (files[i].size > this.MAX_FILE_SIZE) {
```
iloveagent57: Oh, cool.  I like this change - makes it easier to reason about and explain.

openassessment/xblock/submission_mixin.py
```diff
@@ -200,6 +189,56 @@ def save_submission(self, data, suffix=''):  # pylint: disable=unused-argument
         else:
             return {'success': False, 'msg': self._(u"This response was not submitted.")}
 
+    def get_saved_file_metadata(self):
```
iloveagent57: I would love (like, with the fury of a thousand suns) to have created (in a separate PR) a `file_upload_api.py` module that abstracts all of this away, and let's us treat file uploads as objects that have a simple interface.  Something like
```python
class FileUpload:
    def __init__(self, **kwargs):
        self.description = blah
        self.size = blah
        self.index = 3
        # etc.

    @classmethod
    def file_uploads_for_ora_block(cls, block):
        for thing in zipped_metadata:
            # etc.
```

I know I'm really waving hands here, hopefully you get the idea.  Would you be up for stubbing out a ticket and helping me groom this on Tuesday?
BTW, I think what you're doing here is a good first step.

openassessment/xblock/submission_mixin.py
```diff
@@ -200,6 +189,56 @@ def save_submission(self, data, suffix=''):  # pylint: disable=unused-argument
...
+                    except IndexError:
+                        pass
+                    file_metadata.append((file_url, file_description, file_name, file_size))
+                else:
+                    break
+        else:
+            zipped_metadata = zip(saved_files_descriptions, saved_files_names, saved_files_sizes)
+            for i, (file_description, file_name, file_size) in enumerate(zipped_metadata):
+                if file_description is None:
+                    # We are passing Nones to the template because when files are deleted, we still want to
```
iloveagent57: In our future world of being able to track indices of file objects more easily, we could just pass an index to the front-end to let it know how to correctly create the download/upload URLs.

openassessment/xblock/submission_mixin.py
```diff
@@ -200,6 +189,56 @@ def save_submission(self, data, suffix=''):  # pylint: disable=unused-argument
...
+                    # represent them as empty elements in order to preserve the indices and thus urls of the
+                    # remaining files.
+                    file_metadata.append((None, None, None, 0))
```
iloveagent57: Also, is this the important part - the resolution of the bug you discovered yesterday?

openassessment/xblock/submission_mixin.py
jansenk: I would also love that, i'm in

openassessment/xblock/submission_mixin.py
jansenk: In a sense, but I'd say https://github.com/edx/edx-ora2/pull/1287/files?file-filters%5B%5D=.py#diff-df7d00321b610b1545d1563022b59fadR297-R329 (in submission_mixin) is the important part (actually using this existing code in the submissions process)

openassessment/xblock/test/test_submission.py
```diff
@@ -190,6 +191,86 @@ def test_download_url(self, xblock):
...
+
+        data = {'fileMetadata': file_metadata, 'itsamee': 'magrio'}
+        resp = self.request(
+            xblock,
+            'save_files_descriptions',
+            json.dumps(data),
+            response_format='json'
+        )
+        self.assertTrue(resp['success'])
+        resp = json.loads(self.request(xblock, 'remove_uploaded_file', json.dumps({'filenum': 2})))
```
iloveagent57: Nit: looking at line 244, I think adding the `response_format='json'` kwarg lets you remove the `json.loads()` call.

openassessment/xblock/test/test_submission.py
```diff
@@ -190,6 +191,86 @@ def test_download_url(self, xblock):
...
+        self._create_uploaded_files(5, xblock.scope_ids.usage_id)
+        xblock.xmodule_runtime = Mock(
+            course_id='test_course',
+            anonymous_student_id='test_student',
+        )
+        file_metadata = [
+            self._create_entry('File Number ' + str(i) , 'file_' + str(i), 1000) for i in range(5)
+        ]
+        expected_file_metadata = list(file_metadata)
+        expected_file_metadata.pop(2)
```
iloveagent57: Nit: create a variable like `file_index_to_delete = 2`, then you can use it here and in line 247 where you make the request to delete the file.

openassessment/xblock/test/test_submission.py
```diff
@@ -190,6 +191,86 @@ def test_download_url(self, xblock):
...
+
+    @mock_s3
+    @override_settings(
+        AWS_ACCESS_KEY_ID='foobar',
+        AWS_SECRET_ACCESS_KEY='bizbaz',
+        FILE_UPLOAD_STORAGE_BUCKET_NAME="mybucket"
+    )
+    @scenario('data/file_upload_scenario.xml', user_id='bob')
+    def test_delete_and_submit(self, xblock):
+        """ Test that after deleting a file, the remaining files are currectly submitted  """
```
iloveagent57: Nit: "correctly"

PR #1289, opened 2019-11-06
EDUCATOR-4779 | Disable text-shadow and box-shadow on the delete butt…

iloveagent57:
…on during active/hover/focus states.

https://openedx.atlassian.net/browse/EDUCATOR-4779

Before:
Screen Shot 2019-11-06 at 2 40 33 PM

After:
Screen Shot 2019-11-06 at 3 56 20 PM


jansenk:
haha i thought this looked janky before thank you for fixing it 🙇

PR #1293, opened 2019-11-13
New file upload API

iloveagent57:
[EDUCATOR-4783](https://openedx.atlassian.net/browse/EDUCATOR-4783): Create python API abstraction layer over file uploads

FYI: @edx/masters-devs

### DIFF
--- a/openassessment/fileupload/api.py
+++ b/openassessment/fileupload/api.py
@@ -1,13 +1,20 @@
 """
 The File Upload application is designed to allow the management of files
-associated with submissions. This can be used to upload new files and provide
-URLs to the new location.
-
+associated with submissions. This can be used to upload new files, manage
+URLs of existing files, and delete files.
 """
 
-from __future__ import absolute_import
+from __future__ import absolute_import, unicode_literals
+
+import json
+import logging
+
+from openassessment.fileupload.exceptions import FileUploadError
 
 from . import backends
+
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
 def get_upload_url(key, content_type):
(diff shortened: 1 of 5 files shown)


openassessment/fileupload/api.py
```diff
@@ -29,3 +31,160 @@ def remove_file(key):
...
+        self.item_id = student_item_dict.get('item_id')
+
+    @property
+    def exists(self):
+        return self.key is not None
+
+    @property
+    def download_url(self):
+        if self.exists:
+            return get_download_url(self.key)
```
Andytr1: I know this POC - but what happens if there is no key - i.e: should we return a NULL/empty string?

openassessment/xblock/submission_mixin.py
```diff
@@ -44,11 +33,11 @@ class SubmissionMixin(object):
 
     """
 
-    ALLOWED_IMAGE_MIME_TYPES = ['image/gif', 'image/jpeg', 'image/pjpeg', 'image/png']
+    ALLOWED_IMAGE_MIME_TYPES = ['image/gif', 'image/jpeg', 'image/pjpeg', 'image/png']  # pragma: no cover
 
-    ALLOWED_FILE_MIME_TYPES = ['application/pdf'] + ALLOWED_IMAGE_MIME_TYPES
+    ALLOWED_FILE_MIME_TYPES = ['application/pdf'] + ALLOWED_IMAGE_MIME_TYPES  # pragma: no cover
```
Andytr1: nice. I like the explicit pragma.

openassessment/xblock/submission_mixin.py
```diff
@@ -64,6 +53,10 @@ class SubmissionMixin(object):
         'htm', 'html',
     ]
 
+    @cached_property
```
Andytr1: I am curious, what is the reason behind the use of cached_property ?

openassessment/fileupload/api.py
```diff
@@ -5,11 +5,19 @@
 
 """
 
-from __future__ import absolute_import
```
Andytr1: nit for header comments: Instead of "and provide URLs to the new location" consider: 
"and manage URLs of uploaded files."

openassessment/fileupload/api.py
```diff
@@ -29,3 +37,299 @@ def remove_file(key):
...
+    except ValueError:
+        return []
+
+
+class FileUpload(object):
+    """
+    A layer of abstraction over the various components of file
+    data stored as ORA XBlock user-scoped fields.
+    """
+    def __init__(self, name=None, description=None, size=None, index=None, descriptionless=False, **student_item_dict):
```
Andytr1: Is the "descriptionless" parameter there to handle the boundary condition that we fixed? I am trying to think if there is a slightly better / descriptive (pardon the pun) name ?

openassessment/fileupload/api.py
```diff
@@ -29,3 +37,299 @@ def remove_file(key):
...
+        self.item_id = student_item_dict.get('item_id')
+        self.descriptionless = descriptionless
+
+    @property
+    def exists(self):
+        return (self.description is not None) or self.descriptionless
+
+    @property
+    def download_url(self):
+        if self.exists:
```
Andytr1: What happens if self.exists == False ?

openassessment/fileupload/api.py
```diff
@@ -29,3 +37,299 @@ def remove_file(key):
...
+        """
+        Returns:
+            A dictionary representation of the FileUpload.
+        """
+        attrs = ('description', 'name', 'size', 'course_id', 'student_id', 'item_id', 'descriptionless')
+        return {
+            key: getattr(self, key, None) for key in attrs
+        }
+
+    def __eq__(self, other):
```
Andytr1: I like the thinking ahead. Do we foresee a need to compare two file uploads

openassessment/fileupload/api.py
```diff
@@ -29,3 +37,299 @@ def remove_file(key):
...
+        Returns:
+            True if self's dict representation equals other's dict representation,
+            False otherwise.
+        """
+        return self._to_dict() == other._to_dict()  # pylint: disable=protected-access
+
+
+class FileUploadManager(object):
+    """
+    Manages the various CRUD operations of file uploads
```
Andytr1: nit: remove "various" : Manages CRUD operations...

openassessment/fileupload/api.py
```diff
@@ -29,3 +37,299 @@ def remove_file(key):
...
+        """
+        Given file upload data as list fields, return a list of FileUploads constructed from those fields
+        """
+        file_fields_by_key = {
+            'name': names,
+            'description': descriptions,
+            'size': sizes,
+        }
+
+        if not descriptions:
```
Andytr1: Just checking that this is required to handle the old boundary case - i.e: we can't throw an exception if descriptions are null?

openassessment/xblock/submission_mixin.py
iloveagent57: "The @cached_property decorator caches the result of a method with a single self argument as a property. The cached result will persist as long as the instance does, so if the instance is passed around and the function subsequently invoked, the cached result will be returned."  So it's an easy way of not having a new `FileUploadManager` instantiated each time `self.upload_manager` is referenced.  Not a big deal here, because `FileUploadManager` doesn't store much state.

openassessment/fileupload/api.py
iloveagent57: Yes it is.  I couldn't think of a better one, but lemme know if you have suggestions.

openassessment/fileupload/api.py
iloveagent57: Then `download_url` returns `None`, which is what we want - deleted files should have null as a download URL.

openassessment/fileupload/api.py
iloveagent57: This only exists for the sake of unit tests at the moment.

openassessment/fileupload/api.py
iloveagent57: Hmm, I don't think so, but I'll make sure that we have a unit test that covers that condition.

iloveagent57: 👍 Added a unit test and fixed a bug.

PR #1294, opened 2019-11-13
Ensure that all file metadata field lists are of same length during all operations

iloveagent57:
This fixes a "migration" issue when new uploaded files get staged alongside old uploaded files:
- Old uploads contain the `fileDescription` and `fileName` metadata fields
- New uploads contain an additional `fileSize` metadata field

Added new functions to handle validating/filling out the metadata object so the two metadata schemas can exist together, filling nonexistent metadata values with empty/null placeholders

FYI: @edx/masters-devs


openassessment/xblock/test/test_submission.py
```diff
@@ -445,6 +451,56 @@ def test_open_saved_response(self, xblock):
             }
         )
 
+        self.assertEqual(2, len(xblock.get_file_descriptions()))
```
iloveagent57: Nit: we should assert that the context of the list equals what we expect.  We should also do the same for the file names and the file sizes.

openassessment/xblock/submission_mixin.py
```diff
@@ -239,60 +243,143 @@ def get_saved_file_metadata(self):
...
+    def delete_safe_normalized_file_metadata(self, index):
+        """
+        Given a file index to remove, null out its metadata in our stored file metadata fields
+
+        Args:
+            index: file index to remove
+
+        Returns: newly updated file metadata fields
+        """
+        existing_file_descriptions, existing_file_names, existing_file_sizes = self.get_safe_normalized_file_metadata()
```
iloveagent57: Nit: `existing_file_descriptions` etc. is a bit of a misnomer at the end of this function, since we modify and return it.  Maybe `stored_file_descriptions` etc. instead?

PR #1298, opened 2019-11-18
dump response without escaping to render special characters

DawoudSheraz:
### Description
This PR is updating the way ORA submission data is collected by dumping the student response. This change is to ensure that special characters are rendered in the CSV and aren't represented in the form of code points.
Following is the sample output before and after the change:

#### Before
`"{u'file_keys': [], u'files_sizes': [], u'files_name': [], u'files_descriptions': [], u'parts': [{u'text': u'\u95a2\u4fc2\u306e\u8a71\u984c\u306f\u3001\u30c8 '}]}"`

#### After
`"{""file_keys"": [], ""files_sizes"": [], ""files_name"": [], ""files_descriptions"": [], ""parts"": [{""text"": ""関係の話題は、ト ""}]}"` 

### Sandbox
 - https://dawoudsheraz.sandbox.edx.org/courses/course-v1:arb+123+2019_T2/course/

### Reviewers
 - [x] @awaisdar001 
 - [x] @Ayub-Khan 

### Testing Instructions

1. Visit the test ORA.
2. Create a submission with some Unicode data
3. Go to the Instructor tab with a staff account and navigate to the `Data Download` section.
4. Click generate ora data
5. Since the report isn't uploaded on the S3 and is saved locally, SSH into the sandbox(I will be giving the SSH access).
6. Navigate to `/tmp/edx-s3/grades` folder. Move into the folder with some hash as its title.
7. The report should be there. Viewing the report will render the special characters

### DIFF
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -478,7 +478,7 @@ class OraAggregateData(object):
                 submission['student_item'],
                 student_item['student_id'],
                 submission['submitted_at'],
-                submission['answer'],
+                json.dumps(submission['answer'], ensure_ascii=False),
                 assessments_cell,
                 assessments_parts_cell,
                 score.get('created_at', ''),
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -478,6 +478,7 @@ class OraAggregateData(object):
                 submission['student_item'],
                 student_item['student_id'],
                 submission['submitted_at'],
+                #  Dumping required to render special characters in CSV
                 json.dumps(submission['answer'], ensure_ascii=False),
                 assessments_cell,
                 assessments_parts_cell,
--- a/openassessment/tests/test_data.py
+++ b/openassessment/tests/test_data.py
@@ -6,6 +6,7 @@ Tests for openassessment data aggregation.
 from __future__ import absolute_import, print_function
 
 import csv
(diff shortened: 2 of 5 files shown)


openassessment/tests/test_data.py
```diff
@@ -495,6 +495,25 @@ def test_collect_ora2_data(self):
...
 
+    @ddt.data(
+        u'ゅせ第1図', u'ぞひのぽ。', u"ГЂіи lіиэ ъэтшээи"
+    )
+    def test_collect_ora2_data_with_special_characters(self, answer):
+        """
+        Scenario: Verify the data collection for ORA2 works with special or non-ascii characters.
+
+        Given the submission object
+        Then update its answers with a non-ascii value
```
DawoudSheraz: `answer`

openassessment/data.py
```diff
@@ -478,7 +478,7 @@ def collect_ora2_data(cls, course_id):
                 submission['student_item'],
                 student_item['student_id'],
                 submission['submitted_at'],
-                submission['answer'],
+                json.dumps(submission['answer'], ensure_ascii=False),
```
Ayub-Khan: Do add a comment why we are doing this.

PR #1299, opened 2019-11-19
EDUCATOR-4780: Add JS/CSS regeneration as pre-commit step

nsprenkle:
## Changes
[EDUCATOR-4780](https://openedx.atlassian.net/browse/EDUCATOR-4780): Add javascript compiling as a pre-commit step to ORA2 repo

- Add a new make target, `install-git-hooks`, to redirect repo git hooks to the new versioned `hooks` folder

## Etc.
- Add VSCode .gitignore entry
- Update NPM dependencies
- Documentation updates
- Regenerate JS/CSS as a side-effect of the commit command

## Resources
- https://codeinthehole.com/tips/tips-for-using-a-git-pre-commit-hook/
- https://www.tygertec.com/git-hooks-practical-uses-windows/

FYI: @edx/masters-devs


nsprenkle:
> Very late in the process for me to throw in these particular pennies, but have we considered doing this check via travis rather than before someone can commit? Pre-commit hooks that take a lot of time encourage larger commits, which doesn't really help folks use version control well. I think a better choke-point for this Dev Exp.-wise would probably be a blocking travis check.

Thanks @matthugs, that's totally valid! Can you explain the workflow you're proposing a bit more?

nsprenkle:
Pending further discussion about how best to alert/automate this (e.g. check to notify developer on commit instead of adding to a commit, Travis, etc.)

PR #1304, opened 2019-11-26
Pusher POC for realtime collaboration

iloveagent57:
To use:
* specify PUSHER_APP_ID, PUSHER_KEY, PUSHER_SECRET in `settings/private.py` (this file is not tracked by git)
* ~Figure out how the correct URL to give to the Pusher client on the JS side (i.e. it's an xblock endpoint that can be accessed at `<something>/pusher_authenticate`.~


nsprenkle:
Closing as POC only

PR #1308, opened 2019-12-03
Show Team info in Team-enabled ORA

jansenk:
For team-enabled ora in team-enabled course, show a box with team name and team members (including yourself)

Screen Shot 2019-12-04 at 11 30 13 AM


openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -64,6 +69,30 @@ describe("OpenAssessment.ResponseView", function() {
...
+            } else {
+                return successPromiseWithResult(this.listTeamsResult)
+            };
+        }
+        
+        this.teamDetailError = false
+        this.teamDetailResult = {
+            course_id: 'CourseID',
+            description: "A team!",
+            membership: [{user: 'user1'}, {user: 'user2'}, {user: 'user3'}]
```
Andytr1: What is the benefit of having a key in a dictionary with a single value? Would something like: membership:` ['user1', 'user2','user3']` make more sense?

jansenk: This is a mock of what we will recieve from the teams API. There are more keys here but currently I'm only using the one so that's all I mocked

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -59,25 +59,78 @@ OpenAssessment.ResponseView.prototype = {
...
+    generateTeamString: function(teamName, usernames) {
+        var userString = usernames.join('</b>, <b>')
+        var teamMsg;
+        if (usernames.length >= 1) {
+            userString = '<b>' + userString + '</b>'
+            teamMsg = 'Your teammates are: ' + userString
+        } else {
+            teamMsg = 'You have no teammates'
+        }
+        return 'You are on team <b>' + teamName + '</b>.' + teamMsg
```
Andytr1: Trivial nit: What about wording of: Team: <teamName>. Teammates: <list>

openassessment/xblock/static/js/src/oa_server.js
```diff
@@ -612,6 +612,43 @@ if (typeof OpenAssessment.Server === "undefined" || !OpenAssessment.Server) {
             $.ajax({
                 type: "POST", url: url, data: payload, contentType: jsonContentType
             });
-        }
-    };
+        },
+
+        /**
```
Andytr1: I would imagine the comments will get filled out with details of return values, etc..

openassessment/xblock/static/js/src/oa_server.js
```diff
@@ -612,6 +612,43 @@ if (typeof OpenAssessment.Server === "undefined" || !OpenAssessment.Server) {
...
+
+        listTeams: function(courseId) {
+            var teamsUrl = 'http://www.courses.edx.org/api/team/v0/teams/';
+            return $.Deferred(function(defer) {
+                $.ajax({
+                    type: "GET",
+                    url: teamsUrl,
+                    data: JSON.stringify({course_id: courseId}),
+                    contentType: jsonContentType
+                }).done(function(data) {
```
Andytr1: I am confused - why can't we have multiple teams in a given course?

jansenk: from the ticket: 

"Use this view to determine the team identifier for the current user and course - you will never get back a list of length > 1 with this call (this will change in the medium-term future): "

"JS unit tests - required.  Cover the basic cases: no team, one team, what if I erroneously get more than 1 team back from the API endpoint?  What if the API returns a non-200 status code response?  What if the waffle switch is not enabled."

jansenk:
Yay the python3 test was just flaky apparently

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -941,4 +952,66 @@ OpenAssessment.ResponseView.prototype = {
...
+     * @param {array} usernames - The list of users on the team
+     */
+    generateTeamStringHTML: function(teamName, usernames) {
+        var userString = usernames.join('</strong>, <strong class="emphasis">');
+        var teamMsg;
+        if (usernames.length >= 1) {
+            userString = '<strong class="emphasis">' + userString + '</strong>';
+            teamMsg = 'Team Members: ' + userString;
+        } else {
+            teamMsg = 'You have no teammates';
```
iloveagent57: I think this is an impossible state.  If you're the only one on the team, `usernames` will only contain you.  If you're not in the team, `getTeamInfo()` will return null and you wouldn't expect to see any team info rendered.

openassessment/xblock/submission_mixin.py
```diff
@@ -286,6 +286,14 @@ def create_submission(self, student_item_dict, student_sub_data):
 
         return submission
 
+    @XBlock.json_handler
+    def get_student_username(self, data, suffix):  # pylint: disable=unused-argument
```
iloveagent57: Nit: it would be good to have a new unit test for this handler.

jansenk: I thought that too but I'm not exactly sure how I would meaningfully test it, i feel like any test would mock away basically everything

openassessment/xblock/static/js/src/lms/oa_response.js
jansenk: That ... is definitely true lol

openassessment/xblock/submission_mixin.py
iloveagent57: Yeah, that's fine though - you just want something to check that there's a `/get_student_username` endpoint available from this XBlock, and that it returns some username in its data.

iloveagent57:
One last thing: I think we mentioned in the ticket that the Team name should actually be a link to the  team's page (for example: http://localhost:18000/courses/course-v1:edX+DemoX+Demo_Course/teams/#teams/Topic1ID/the-a-team-b66bd283a50b4d93a990b6f49c823086) But if that blows up the scope of this ticket, we can create a new ticket to capture that work.

PR #1309, opened 2019-12-03
ran make upgrade on repo

jinder1s:


requirements/base.txt
```diff
@@ -4,17 +4,16 @@
...
-certifi==2019.9.11        # via requests
+certifi==2019.11.28       # via requests
 chardet==3.0.4            # via requests
 defusedxml==0.5.0
 django-model-utils==3.0.0
 django==1.11.26
 git+https://github.com/edx/django-rest-framework.git@1ceda7c086fddffd1c440cc86856441bbf0bd9cb#egg=djangorestframework
 edx-i18n-tools==0.4.8
-edx-submissions==3.0.2
+edx-submissions==3.0.3
```
iloveagent57: Just a reminder that you'll have to make sure edx-submissions gets upgraded to 3.0.3 in edx-platform, too, when you upgrade ora2 there.

setup.py
```diff
@@ -26,7 +26,7 @@ def load_requirements(*requirements_paths):
     """
     requirements = set()
     for path in requirements_paths:
-        with open(path) as reqs:
+        with open(os.path.expanduser(path)) as reqs:
```
jmbowman: Why was this needed?

setup.py
```diff
@@ -57,7 +57,7 @@ def load_requirements(*requirements_paths):
     packages=find_packages(include=['openassessment*'], exclude=['*.test', '*.tests']),
     include_package_data=True,
     # Todo - this should be loading 'requirements/base.in' but Tox is having an issue with it
-    install_requires=load_requirements('requirements/base.txt'),
+    install_requires=load_requirements('requirements/base.in'),
     tests_require=load_requirements('requirements/test.txt'),
```
jmbowman: Should probably also switch to `test.in` while you're at it.  And then I think you could drop `requirements/*.txt` from `MANIFEST.in`.

setup.py
jinder1s: This was one of the things I used during debugging, will remove it

setup.py
iloveagent57: @jinder1s I think if you add `requirements/*.in` to the MANIFEST.in file, tox won't balk at you switching this `install_requires` line to look at `base.in`.

PR #1310, opened 2019-12-10
Mattcarter/educator 4806

mattcarter:
Draft for review. Partly as the quality tests are reporting errors on passing statements on my devstack and that shouldn't block this


openassessment/assessment/models/base.py
```diff
@@ -828,3 +832,22 @@ def _check_all_criteria_assessed(cls, rubric_index, selected_criteria, criteria_
         if missing_criteria:
             msg = u"Missing selections for criteria: {missing}".format(missing=', '.join(missing_criteria))
             raise InvalidRubricSelection(msg)
+
+
+@python_2_unicode_compatible
```
Andytr1: I am curious, what is this attribute fore?

mattcarter: the decorator is to support python2 & python3 in a single codebase.  Basically, using it for QOL reasons

PR #1312, opened 2019-12-11
add student module support for upload info

DawoudSheraz:
### Description
We have had various instances where the file upload information in `submissions` was missing. Because of that, during the peer assessment or staff viewing the submission, the files' information is not visible. However, there is evidence that files were uploaded. Some file artifacts, like description, were present in the student module. The learners themselves could see the link to their files. There isn't enough information on why it is happening. In an attempt to streamline our behavior, this PR is introducing a workaround that will allow the staff members to see the uploaded files. By using the service introduced in https://github.com/edx/edx-platform/pull/22425, when staff renders a learner's submission and file upload was either optional/required, the files data will be obtained from CSM under certain conditions. The workaround will be hidden behind the waffle switch/course override **openresponseassessment.user_state_upload_data**. 

Following is the log which will tell if the CSM was queried for the upload information for a particular user:
```
staff_area_mixin.py:272 - Checking student module for upload info for user: edx in block: block-v1:myOrg+pd379+2019_T2+type@openassessment+block@fc057c43a31d43468e10fab58151a368
```

### Testing Instructions
**Note: Since the sandbox creation has been down for a few days, the reviewers will have to test it out on their local machines**

1. Fetch this platform branch on your local: https://github.com/edx/edx-platform/pull/22495
2. Install the requirements to bring in the changes.
3. Create an ORA problem that requires file upload.
4. Submit and verify entries that have been created in both CSM and submission.
5. Now, go to MySQL and update the submission's raw_answer for the latest entry like following:
`{"file_keys":[],"files_sizes":[],"parts":[{"text":"Hello 2"}],"files_descriptions":[],"files_names":[]}`
     -  This is an attempt to mimic the behavior happening on the production where upload information is empty

6. View the learner's submission using `Manage Individual Learners`. Note the file information isn't present(view the submission in courseware as the service is made part of runtime in module_render).
7. Now, either enable waffle switch/course override **openresponseassessment.user_state_upload_data**.
8. View submission again through `Manage Individual Learners`. The file link will be visible and the file will be downloadable.
9. Also, confirm the logs have been hit when using CSM.

### Tests
 - They will be added in a follow-up PR

### Reviewers
 - [x] @awaisdar001 
 - [ ] @iloveagent57

### DIFF
--- a/openassessment/xblock/openassessmentblock.py
+++ b/openassessment/xblock/openassessmentblock.py
@@ -38,6 +38,7 @@ from openassessment.xblock.workflow_mixin import WorkflowMixin
 from openassessment.xblock.xml import parse_from_xml, serialize_content_to_xml
 from webob import Response
 from xblock.core import XBlock
+from xblock.exceptions import NoSuchServiceError
 from xblock.fields import Boolean, Integer, List, Scope, String
 from xblock.fragment import Fragment
 
@@ -91,6 +92,7 @@ def load(path):
 
 @XBlock.needs("i18n")
 @XBlock.needs("user")
+@XBlock.needs("user_state")
 class OpenAssessmentBlock(MessageMixin,
                           SubmissionMixin,
                           PeerAssessmentMixin,
@@ -351,6 +353,53 @@ class OpenAssessmentBlock(MessageMixin,
             A unique id for (user, course) pair
         """
         return self.runtime.service(self, 'user').get_anonymous_user_id(username, course_id)
+
+    def is_user_state_service_available(self):
+        """
+        Check if the user state service is present in runtime.
+        """
+        try:
+            self.runtime.service(self, 'user_state')
+            return True
+        except NoSuchServiceError:
+            return False
+
+    def get_user_state(self, username):
+        """
+        Get the student module state for the given username for current ORA block.
+
+        Arguments:
+            username(str): username against which the state is required in the current block.
+
+        Returns:
+            user state, if found, else empty dict
+        """
+        if self.is_user_state_service_available():
+            user_state_service = self.runtime.service(self, 'user_state')
+            return user_state_service.get_state_as_dict(username, str(self.location))  # pylint: disable=no-member
+        return {}
+
+    def should_use_user_state(self, upload_urls):
+        """
+        Return a boolean if the user state is used for additional data checks.
+
+        User state is utilized when all of the following are true:
+        1. user state service is available(which is only part of courseware)
+        2. The waffle flag/switch is enabled
+        3. the file upload is required or optional
+        4. the file data from submission is missing information
+        """
+        return not any(upload_urls) \
+            and self.is_user_state_service_available() \
+            and self.user_state_upload_data_enabled() \
+            and self.file_upload_response
+
+    def get_student_item_dict_from_username(self, username):
+        """
+        Get the item dict for a given username in the parent course of block.
+        """
+        anonymous_user_id = self.get_anonymous_user_id(username, self.course_id)
+        return self.get_student_item_dict(anonymous_user_id=anonymous_user_id)
 
     def get_student_item_dict(self, anonymous_user_id=None):
         """Create a student_item_dict from our surrounding context.
--- a/openassessment/xblock/staff_area_mixin.py
+++ b/openassessment/xblock/staff_area_mixin.py
@@ -266,6 +266,12 @@ class StaffAreaMixin(object):
         if submission:
             context["file_upload_type"] = self.file_upload_type
             context["staff_file_urls"] = self.get_download_urls_from_submission(submission)
+            if self.should_use_user_state(context["staff_file_urls"]):
+                logger.info(u"Checking student module for upload info for user: {username} in block: {block}".format(
+                    username=student_username,
+                    block=str(self.location)
+                ))
+                context['staff_file_urls'] = self.get_files_info_from_user_state(student_username)
 
         if self.rubric_feedback_prompt is not None:
             context["rubric_feedback_prompt"] = self.rubric_feedback_prompt
--- a/openassessment/xblock/submission_mixin.py
+++ b/openassessment/xblock/submission_mixin.py
@@ -446,6 +446,38 @@ class SubmissionMixin(object):
                 urls.append((file_download_url, '', ''))
         return urls
 
+    def get_files_info_from_user_state(self, username):
+        """
+        Returns the files information from the user state for a given username.
+
+        If the files information is present in the user state, return a list of following tuple:
+        (file_download_url, file_description, file_name)
+
+        Arguments:
+            username(str): user's name whose state is being check for files information.
+        Returns:
+            List of files information tuple, if present, else empty list.
+        """
+
+        files_info = []
+        user_state = self.get_user_state(username)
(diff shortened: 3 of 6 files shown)


openassessment/xblock/staff_area_mixin.py
```diff
@@ -266,6 +266,12 @@ def get_student_submission_context(self, student_username, submission):
         if submission:
             context["file_upload_type"] = self.file_upload_type
             context["staff_file_urls"] = self.get_download_urls_from_submission(submission)
+            if self.file_upload_response and not any(context['staff_file_urls']):
+                logger.info(u"Checking student module for upload info for user: {username} in block: {block}".format(
```
awaisdar001: What do you think a log after calling `self.get_files_info_from_user_state(student_username)` where you cal also log if the state was restored from CSM service?

openassessment/xblock/submission_mixin.py
```diff
@@ -438,6 +438,38 @@ def get_download_urls_from_submission(self, submission):
...
+            username(str): user's name whose state is being check for files information.
+        Returns:
+            List of files information tuple, if present, else empty list.
+        """
+
+        files_info = []
+        user_state = self.get_user_state(username)
+        item_dict = self.get_student_item_dict_from_username(username)
+        if u'saved_files_descriptions' in user_state:
+            # pylint: disable=protected-access
```
awaisdar001: this will disable all errors after this line, so you don't have to disable it again on L#460

openassessment/xblock/staff_area_mixin.py
DawoudSheraz: Do you mean checking before & after CSM call and logging both data values?

openassessment/xblock/submission_mixin.py
DawoudSheraz: ok

DawoudSheraz:
> Are you going to update the ORA version in this PR as well?

Yes. Just looking at the code to see if it can be refactored further. I still have to add `WaffleSwitch` to be able to turn the feature on/off globally. 
I was also thinking of adding the tests in a follow-up PR, given the severity of the issue.

openassessment/xblock/staff_area_mixin.py
awaisdar001: not necessarily logging the data value, but just a flag that files were missing and were found / not found from csm; or something like that.

awaisdar001:
> > Are you going to update the ORA version in this PR as well?
> 
> Yes. Just looking at the code to see if it can be refactored further. I still have to add `WaffleSwitch` to be able to turn the feature on/off globally.
> I was also thinking of adding the tests in a follow-up PR, given the severity of the issue.

sounds good!!

openassessment/xblock/waffle_mixin.py
```diff
@@ -51,16 +53,27 @@ def _course_waffle_flag(flag_name):
         CourseWaffleFlag = import_course_waffle_flag()  # pylint: disable=invalid-name
         return CourseWaffleFlag(WAFFLE_NAMESPACE, flag_name)  # pylint: disable=feature-toggle-needs-doc
 
-    def team_submissions_enabled(self):
+    def is_flag_enabled(self, flag):
```
awaisdar001: 👍

openassessment/xblock/waffle_mixin.py
```diff
@@ -8,6 +8,8 @@
 
 TEAM_SUBMISSIONS = 'team_submissions'
```
awaisdar001: nice work in this file 👍

openassessment/xblock/openassessmentblock.py
```diff
@@ -352,6 +352,34 @@ def get_anonymous_user_id(self, username, course_id):
         """
         return self.runtime.service(self, 'user').get_anonymous_user_id(username, course_id)
 
+    def is_user_state_service_available(self):
+        """
+        Check if the user state service is present in runtime.
+        """
+        return self.runtime._services.get('user_state') is not None  # pylint: disable=protected-access
```
iloveagent57: I think if you decorate the `OpenAssessmentBlock` class with `@XBlock.needs('user_state')`, you'll be able to call `self.runtime.service(self, 'user_state')` without having to look at the private `_services` attribute.

openassessment/xblock/openassessmentblock.py
```diff
@@ -352,6 +352,34 @@ def get_anonymous_user_id(self, username, course_id):
...
+        """
+        Get the student module state for the given username for current ORA block.
+
+        Arguments:
+            username(str): username against which the state is required in the current block.
+
+        Returns:
+            user state, if found, else empty dict
+        """
+        user_state_service = self.runtime._services.get('user_state')  # pylint: disable=protected-access
```
iloveagent57: Same idea here with `@XBlock.needs('user_state')`.

openassessment/xblock/staff_area_mixin.py
```diff
@@ -252,6 +252,21 @@ def get_student_submission_context(self, student_username, submission):
...
+            User state is utilized when all of the following are true:
+            1. user state service is available(which is only part of courseware)
+            2. The waffle flag/switch is enabled
+            3. the file upload is required or optional
+            4. the file data from submission is missing information
+            """
+            return self.is_user_state_service_available() \
+                and self.user_state_upload_data_enabled() \
+                and self.file_upload_response \
+                and not any(context['staff_file_urls'])
```
iloveagent57: Nit: I know this works because `context` is already defined in the local scope in which this function is executed, but it would be clearer if this helper function simply took `staff_file_urls` as a parameter, so that you would call this function as `if should_use_user_state(context['staff_file_urls']):`

iloveagent57: Also, I'd recommend making the `not any(staff_file_urls)` the first thing we check, since we don't have to check anything else if that list contains at least one file URL.

openassessment/xblock/waffle_mixin.py
iloveagent57: Really minor nit: `is_feature_enabled` would be a better name, since it's answering a question about whether a flag or a switch is active.

openassessment/xblock/openassessmentblock.py
iloveagent57: Also, it's worth mentioning that this platform PR: https://github.com/edx/edx-platform/pull/22495 is required to test locally.

openassessment/xblock/openassessmentblock.py
DawoudSheraz: Sure.

openassessment/xblock/openassessmentblock.py
DawoudSheraz: @iloveagent57 Already mentioned in the ticket's description.

openassessment/xblock/staff_area_mixin.py
DawoudSheraz: Ok. I will address both comments.

DawoudSheraz:
@iloveagent57 Please take a second look. As for the unit tests, I am going to add them to a follow-up PR.

openassessment/xblock/openassessmentblock.py
iloveagent57: D'oh, sorry I missed that.

PR #1313, opened 2019-12-13
Jkantor/teams service

jansenk:
Redo the teams display box to use the new teams service.


openassessment/xblock/openassessmentblock.py
```diff
@@ -1154,25 +1155,40 @@ def publish_event(self, data, suffix=''):  # pylint: disable=unused-argument
         self.runtime.publish(self, event_name, data)
         return {'success': True}
 
-    def get_username(self, anonymous_user_id):
+    def get_real_user(self, anonymous_user_id):
```
Andytr1: nit: why is there "real" in get_real_user ? Why not call it get_user ?

openassessment/xblock/submission_mixin.py
```diff
@@ -637,3 +642,22 @@ def submission_path_and_context(self):
...
+        user = self.get_real_user(student_item_dict['student_id'])
+        if not user:
+            return None
+        team = teams_service.get_team(user, student_item_dict['course_id'])
+        if not team:
+            return None
+        return {
+            'team_name': team.name,
+            'team_topic': team.topic_id,
+            'team_description': team.description,
```
iloveagent57: Nit: we're not using these anywhere yet.

openassessment/xblock/openassessmentblock.py
iloveagent57: "real" as opposed to "anonymous", plus it passes through to `self.xmodule_runtime.get_real_user`, so it keeps the naming consistent.

openassessment/xblock/submission_mixin.py
```diff
@@ -637,3 +642,22 @@ def submission_path_and_context(self):
             path = 'openassessmentblock/response/oa_response_submitted.html'
 
         return path, context
+
+    def get_team_info(self):
+        teams_service = self.runtime.service(self, 'teams')
+        if not teams_service:
+            return None
```
iloveagent57: Do you think it's worth logging a warning or something here?

iloveagent57: actually, I wonder if it's also worth raising an exception here (and catching in the caller) to help distinguish between the case where the service isn't available/can't find the user, and where we have the service and the user, but the user isn't on a team yet - because we'll need to distinguish that latter case when we do the work to hide content from teamless users.

iloveagent57:
Also, don't forget to bump the version in `setup.py` and `package.json`

openassessment/xblock/submission_mixin.py
```diff
@@ -603,6 +605,16 @@ def submission_path_and_context(self):
             context['save_status'] = self.save_status
             context['enable_delete_files'] = True
 
+            if self.teams_enabled:
+                try:
+                    team_info = self.get_team_info()
+                    if team_info:
+                        context.update(team_info)
+                except ObjectDoesNotExist:
+                    logger.error('User associated with anonymous_user_id can not be found')
```
iloveagent57: Nit: it'd be helpful to put the anonymous_user_id and/or `self.location` in this error message.

openassessment/xblock/submission_mixin.py
```diff
@@ -637,3 +649,26 @@ def submission_path_and_context(self):
             path = 'openassessmentblock/response/oa_response_submitted.html'
 
         return path, context
+
+    def get_team_info(self):
+        """
+        Returns dict with team info from the team service or None if th user is not on a team
```
iloveagent57: s/th/the/

PR #1315, opened 2019-12-18
DJango22 support

Ayub-Khan:
-Added Django22 support
-Updated Requiremetns
-Updated tox and travis files
-Fixed tests for django22
-Improved Quality
-New Translations


jmbowman:
It looks like the one failing test on two workers is due to the HTML rendering having switched from `<br />` to `<br>` in Django 2.1.  And currently the quality job is failing on a PEP 8 whitespace violation.

openassessment/locale/en/LC_MESSAGES/djangojs.po
```diff
@@ -307,10 +307,6 @@ msgstr ""
 msgid "Thumbnail view of "
 msgstr ""
 
-#: openassessment/xblock/static/js/openassessment-lms.min.js:1
-msgid "Unable to load team detail"
-msgstr ""
```
jmbowman: Any idea why this was removed from these two files?  It still seems to be in `djangojs.po` for all the other languages.  Rebasing might help, Transifex just updated all the translations 18 hours ago.

openassessment/xblock/test/test_submission.py
```diff
@@ -132,7 +133,10 @@ def test_prompt_line_breaks(self, xblock):
         # Verify that prompts with multiple lines retain line breaks
         # (backward compatibility in case if prompt_type == 'text')
         resp = self.request(xblock, 'render_submission', json.dumps(dict()))
-        expected_prompt = u"<p><br />Line 1</p><p>Line 2</p><p>Line 3<br /></p>"
+        if django.VERSION[0] >= 2 and django.VERSION[1] > 0:
```
jmbowman: I suspect this logic will break for Django 3.0.  You should be able to use `django.VERSION >= (2, 1)` instead.

package.json
```diff
@@ -27,7 +27,7 @@
     "karma-spec-reporter": "^0.0.31",
     "phantomjs-prebuilt": "^2.1.14",
     "sinon": "^7.5.0",
-    "uglify-js": "^3.7.1"
+    "uglify-js": "^3.7.2"
```
jmbowman: Why did this need to change?

openassessment/tests/test_data.py
```diff
@@ -247,7 +247,7 @@ class TestOraAggregateData(TransactionCacheResetTest):
     def _build_criteria_and_assessment_parts(self, num_criteria=1, feedback=""):
         """ Build a set of criteria and assessment parts for the rubric. """
         rubric = RubricFactory()
-        criteria = [CriterionFactory(rubric=rubric, order_num=n - 1) for n in range(num_criteria)]
+        criteria = [CriterionFactory(rubric=rubric, order_num=n + 1) for n in range(num_criteria)]
```
jmbowman: Why did this need to change?

Ayub-Khan: quality.

package.json
Ayub-Khan: this was auto-generated by some thing.

openassessment/locale/en/LC_MESSAGES/djangojs.po
Ayub-Khan: I was not able to find this line in code which means this was removed from code. en is our base language so it is updated when we extract translations and eo is our dummy language its also updated locally. Once this is merged this will be pulled by transifex and will be removed from transifex. Other languages are updated when we pull new languages from transifex. we do not update them locally.

PR #1316, opened 2019-12-18
Run travis builds on tag publish, so that we can get PyPI deploys working

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-4853


mattcarter:
Looks good to me!

PR #1322, opened 2019-12-23
EDUCATOR-4844

Andytr1:
https://openedx.atlassian.net/browse/EDUCATOR-4844

PR #1323, opened 2019-12-24
ORA File names not rendered

AhtishamShahid:
In edX-Platform ORA file uploads don't show the names of the files for following: [`PROD-1093`](https://openedx.atlassian.net/browse/PROD-1093)
1. when viewed through the Manage Individual Learners section. 
2. peer assessment
3. Other than the above-mentioned artifacts, there can be other places where the name isn’t visible, The name data is in the database. This ticket is supposed to find out why the uploads' names aren't shown. 

**Issue :**
DB records are for file names are present with 2 different keys ie `files_name` and `files_names`

**Fix:**
This issue is fixed by updating the code to check for both attributes while data composition.

### DIFF
--- a/openassessment/xblock/submission_mixin.py
+++ b/openassessment/xblock/submission_mixin.py
@@ -432,7 +432,7 @@ class SubmissionMixin(object):
         if 'file_keys' in submission['answer']:
             file_keys = submission['answer'].get('file_keys', [])
             descriptions = submission['answer'].get('files_descriptions', [])
-            file_names = submission['answer'].get('files_name', [])
+            file_names = submission['answer'].get('files_name', submission['answer'].get('files_names', []))
             for idx, key in enumerate(file_keys):
                 file_download_url = self._get_url_by_file_key(key)
                 if file_download_url:
--- a/openassessment/xblock/test/test_staff_area.py
+++ b/openassessment/xblock/test/test_staff_area.py
@@ -12,6 +12,7 @@ from six.moves import range, zip
 import six.moves.urllib.error  # pylint: disable=import-error
 import six.moves.urllib.parse  # pylint: disable=import-error
 import six.moves.urllib.request  # pylint: disable=import-error
+import ddt
 
 from openassessment.assessment.api import peer as peer_api
 from openassessment.assessment.api import self as self_api
@@ -64,6 +65,7 @@ class NullUserService(object):
         return MagicMock(opt_attrs={})
 
 
+@ddt.ddt
 class TestCourseStaff(XBlockHandlerTestCase):
     """
     Tests for course staff debug panel.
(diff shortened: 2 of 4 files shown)


openassessment/xblock/test/test_staff_area.py
```diff
@@ -781,3 +781,44 @@ def _create_submission(item, values, types):
         peer_api.on_start(submission["uuid"])
         workflow_api.create_workflow(submission["uuid"], types)
         return submission
+
+    @scenario('data/self_only_scenario.xml', user_id='Bob')
+    def test_staff_area_student_info_image_submission_names(self, xblock):
+        # Bug fix , In some places files_name as a key is used and in other places files_names is used
```
awaisdar001: The docstring can be improved, but rest looks good to me, You would also need to update `version` in `setup.py`. 

e.g. 
https://github.com/edx/edx-ora2/pull/1322/files

openassessment/xblock/submission_mixin.py
```diff
@@ -432,7 +432,7 @@ def get_download_urls_from_submission(self, submission):
         if 'file_keys' in submission['answer']:
             file_keys = submission['answer'].get('file_keys', [])
             descriptions = submission['answer'].get('files_descriptions', [])
-            file_names = submission['answer'].get('files_name', [])
+            file_names = submission['answer'].get('files_name', submission['answer'].get('files_names', []))
```
DawoudSheraz: Can you confirm from the DB if the `files_name` key is used in any submission? Given the usage of `files_names` in other code places, I am assuming this could have been an unintentional mistake.

openassessment/xblock/test/test_staff_area.py
DawoudSheraz: +1 for that

openassessment/xblock/submission_mixin.py
AhtishamShahid: I have verified in Prod database, records are present in DB by both `files_names` and `files_name`

openassessment/xblock/test/test_staff_area.py
AhtishamShahid: Updated versions

DawoudSheraz: @AhtishamShahid The docstring format is still incorrect. Please see some examples on edx-platform (https://github.com/edx/edx-platform/pull/22298/). Add the scenario in the docstring to give the clear intent of the test case.

openassessment/xblock/test/test_staff_area.py
```diff
@@ -781,3 +781,50 @@ def _create_submission(item, values, types):
         peer_api.on_start(submission["uuid"])
         workflow_api.create_workflow(submission["uuid"], types)
         return submission
+
+    @scenario('data/self_only_scenario.xml', user_id='Bob')
+    def test_staff_area_student_info_image_submission_names(self, xblock):
```
DawoudSheraz: `image`? Not sure why this word is in test name.

openassessment/xblock/test/test_staff_area.py
```diff
@@ -781,3 +781,50 @@ def _create_submission(item, values, types):
...
+        """
+        This test checks if file name is rendered on Manage-individual-learner template ie oa_student_info.html.
+
+        Context: In table submission_submissions raw_answer column in db is long text and used as json field ,
```
DawoudSheraz: Better but still needs improvement. We don't have to give back-end/data persistency information in the docstring. They need to explain what is being tested. In our case, we are testing that the file name will be rendered for both `file_names` and `files_names`. Make this test a ddt test and check it for both these keys.

openassessment/xblock/test/test_staff_area.py
AhtishamShahid: Fixed test name.

openassessment/xblock/test/test_staff_area.py
AhtishamShahid: Update test and added ddt for both keys

openassessment/xblock/test/test_staff_area.py
```diff
@@ -781,3 +783,46 @@ def _create_submission(item, values, types):
         peer_api.on_start(submission["uuid"])
         workflow_api.create_workflow(submission["uuid"], types)
         return submission
+
+    @ddt.data('files_names', 'files_name')
+    @scenario('data/self_only_scenario.xml', user_id='Bob')
+    def test_file_name_is_rendered_on_template(self, xblock, key):
+        """
+        This test checks if file name is rendered on Manage-individual-learner template ie oa_student_info.html.
```
DawoudSheraz: something like `Validates the file name is visible to staff when viewing individual submission`. Again, no need to mention the HTML file or the button name.

PR #1326, opened 2019-12-30
add logs for diagnostic

DawoudSheraz:
### Description
This PR is adding a few logs to identify where the URL generation is going wrong in user state workaround. Following are the sample logs:

```
submission_mixin.py:488 - URLWorkaround: no URL for description optional & key 1eb2f9df643698e6195e05f093a830be/course-v1:myOrg+pd379+2019_T2/block-v1:myOrg+pd379+2019_T2+type@openassessment+block@fc057c43a31d43468e10fab58151a368 for user:edx
```

```
api.py:66 - URLWorkaround: Safe Load failed for data field:["optional"] with type:<class 'str'>
```

```
api.py:66 - URLWorkaround: Safe Load failed for data field:["test.txt"] with type:<class 'str'>

```

### Reviewers
 - [x] @awaisdar001 

### Post Review
 - [x] Squash
 - [x] Version bump

### DIFF
--- a/openassessment/fileupload/api.py
+++ b/openassessment/fileupload/api.py
@@ -51,7 +51,7 @@ def get_student_file_key(student_item_dict, index=0):
     return key_template.format(index=index, **student_item_dict)
 
 
-def _safe_load_json_list(field):
+def _safe_load_json_list(field, log_error=False):
     """
     Tries to load JSON-ified string,
     returns an empty list if we try to load some non-JSON-encoded string.
@@ -59,6 +59,11 @@ def _safe_load_json_list(field):
     try:
         return json.loads(field)
     except ValueError:
+        if log_error:
+            logger.exception("URLWorkaround: Safe Load failed for data field:{field} with type:{type}".format(
+                field=field,
+                type=type(field)
+            ))
         return []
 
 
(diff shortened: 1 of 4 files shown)


openassessment/fileupload/api.py
```diff
@@ -51,14 +51,19 @@ def get_student_file_key(student_item_dict, index=0):
...
-def _safe_load_json_list(field):
+def _safe_load_json_list(field, log_error=False):
     """
     Tries to load JSON-ified string,
     returns an empty list if we try to load some non-JSON-encoded string.
     """
     try:
         return json.loads(field)
     except ValueError:
+        if log_error:
```
awaisdar001: I guess it should help if we log the exception every time, so I don't think the param passed here is useful. Please log the exception here.

DawoudSheraz: @awaisdar001 This method also acts as a backward-compatible fix. The file API was recently added with some additional file info(filename & sizes) that don't exist in submissions prior to that. Since data is non-existent, the `ValueError` raised handles it by returning an empty list and not interrupting regular workflow. Logging every time will flood the logs, most of whom will be unnecessary. This is precisely the reason I went with logging during the user state workaround, to give an exact piece of information, when activated.

DawoudSheraz:
> Sounds good.
> are you going to update the version in this PR or the one after?

Updating in this PR.

PR #1329, opened 2020-01-02
add user state upload info unit tests

DawoudSheraz:
### Description
This PR is adding the unit tests for the changes added in https://github.com/edx/edx-ora2/pull/1312. At that time, the changes had to be deployed due to increased escalations and tests were delayed for a follow-up PR. Other than the unit tests addition, a new dependency `testfixtures` has been made a part of the test environment. This willl allow to validate if certain logs were hit during the running of test suites.

### DIFF
--- a/openassessment/xblock/test/test_staff_area.py
+++ b/openassessment/xblock/test/test_staff_area.py
@@ -23,6 +23,10 @@ from openassessment.xblock.data_conversion import prepare_submission_for_seriali
 from openassessment.xblock.test.base import XBlockHandlerTestCase, scenario
 from submissions import api as sub_api
 
+FILE_URL = 'www.fileurl.com'
+SAVED_FILES_DESCRIPTIONS = ['file1', 'file2']
+SAVED_FILES_NAMES = ['file1.txt', 'file2.txt']
+
 STUDENT_ITEM = dict(
     student_id="Bob",
     course_id="test_course",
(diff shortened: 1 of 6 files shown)


openassessment/xblock/test/test_staff_area.py
```diff
@@ -737,6 +758,72 @@ def test_staff_delete_student_state(self, xblock):
         resp = xblock.render_student_info(request)
         self.assertIn("response was not found", resp.body.decode('utf-8').lower())
 
+    @patch('openassessment.xblock.waffle_mixin.WaffleMixin.user_state_upload_data_enabled')
+    @scenario('data/file_upload_missing_scenario.xml', user_id='Bob')
+    @log_capture()
+    def test_staff_area_student_upload_info_from_user_state(self, xblock, logger, waffle_patch):
```
iloveagent57: The order of these arguments doesn't seem right - shouldn't it be `self, logger, xblock, waffle_patch`?

openassessment/xblock/test/test_staff_area.py
```diff
@@ -737,6 +758,72 @@ def test_staff_delete_student_state(self, xblock):
...
+        """
+        waffle_patch.return_value = False
+        self._setup_xblock_and_create_submission(xblock)
+        __, context = xblock.get_student_info_path_and_context('Bob')
+        self.assertFalse(any(context['staff_file_urls']))
+
+    @patch('openassessment.xblock.waffle_mixin.WaffleMixin.user_state_upload_data_enabled')
+    @scenario('data/file_upload_missing_scenario.xml', user_id='Bob')
+    @log_capture()
+    def test_student_userstate_not_used_when_upload_info_in_submission(self, xblock, logger, waffle_patch):
```
iloveagent57: Same question here.

openassessment/xblock/test/test_staff_area.py
DawoudSheraz: https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/test/base.py#L118-L120
The `xblock` argument is always the first argument after self. I will be updating the code to match the argument order with decorator applied order.

DawoudSheraz:
@iloveagent57 Update the code. Please take a second look.

openassessment/xblock/test/test_staff_area.py
iloveagent57: Ah, good to know.  Thanks for updating the order of the decorators, I think that makes it less confusing.

DawoudSheraz:
No need to update version as there are no functionality changes.

PR #1319, opened 2019-12-20
EDUCATOR-4839: Team-enabled ORAs should not be actionable by teamless users

jansenk:
@edx/masters-devs 
[EDUCATOR-4839](https://openedx.atlassian.net/browse/EDUCATOR-4839)
Hide ORA and display message if the ora is team-based and the user is not on a team.
IF the user is course staff or if we are rendering the XBlock in studio preview, use some dummy values to render the page.

ALSO, refactored existing and new teams functionality into TeamsMixin.py and moved testing into test_teams.py

Learner not on a team:
Screen Shot 2019-12-20 at 11 58 42 AM

Course staff not on a team
Screen Shot 2019-12-20 at 11 01 21 AM

PR #1331, opened 2020-01-03
EDUCATOR-4839: Team-enabled ORAs should not be actionable by teamless users

jansenk:
(Recreation of https://github.com/edx/edx-ora2/pull/1319)

@edx/masters-devs 
[EDUCATOR-4839](https://openedx.atlassian.net/browse/EDUCATOR-4839)
Hide ORA and display message if the ora is team-based and the user is not on a team.
IF the user is course staff or if we are rendering the XBlock in studio preview, use some dummy values to render the page.

ALSO, refactored existing and new teams functionality into TeamsMixin.py and moved testing into test_teams.py

Learner not on a team:
Screen Shot 2019-12-20 at 11 58 42 AM

Course staff not on a team
Screen Shot 2019-12-20 at 11 01 21 AM


openassessment/xblock/message_mixin.py
```diff
@@ -49,7 +49,9 @@ def render_message(self, data, suffix=''):  # pylint: disable=unused-argument
 
         # Render the instruction message based on the status of the workflow
         # and the closed status.
-        if status == "done" or status == "waiting":
+        if self.teams_enabled and self.should_hide_team_ora():
+            path, context = 'openassessmentblock/message/oa_message_no_team.html', {}
```
iloveagent57: Does this template file already exist?

openassessment/xblock/team_mixin.py
```diff
@@ -0,0 +1,78 @@
...
+        returns true if the student is on a team, false if the student is not on a team
+        or if an exception is raised while looking up team
+        """
+        try:
+            team = self.team
+        except ObjectDoesNotExist:
+            return False
+        return bool(team)
+
+    def should_hide_team_ora(self):
```
nsprenkle: I think this might read better if we just had something like a `valid_access_to_team_assessment` function that combines this and `teams_enabled`. That way the `if` checks would say "if we are trying an invalid access to a team assessment, don't show it"

openassessment/xblock/message_mixin.py
jansenk: good catch, forgot to copy it over

openassessment/xblock/team_mixin.py
nsprenkle: ... but just a stylistic thought. Feel free to use or disregard.

jansenk: No, I think that you're right, the broader description of what the function does is a better name than this specific use case of the logic

openassessment/xblock/team_mixin.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        anonymous_user_id = self.get_anonymous_user_id_from_xmodule_runtime()
+        user = self.get_real_user(anonymous_user_id)
+        if not user:
+            logger.error(u'{}: User lookup for anonymous_user_id {} failed'.format(self.location, anonymous_user_id))
+            raise ObjectDoesNotExist()
+        team = self.teams_service.get_team(user, self.course_id)
+        return team
+
+    def has_team(self):
```
Andytr1: nit: would assigned_to_team make more sense?

openassessment/xblock/test/test_team.py
```diff
@@ -0,0 +1,119 @@
...
+        # This is required because 'name' is a reserved property for mocks
+        self.team.configure_mock(name='TeamName')
+        self.team.users.all.return_value = [
+            mock.MagicMock(username='UserA'),
+            mock.MagicMock(username='UserB'),
+            mock.MagicMock(username='UserC'),
+        ]
+
+    def get_team(self, user, course_id):  # pylint: disable=unused-argument
+        if self.has_team:
```
Andytr1: consider `return self.team if self.has_team else None`

openassessment/xblock/team_mixin.py
jansenk: the wording might be misleading since users can sometimes join teams themselves

jansenk: "some way or another, this block (user-course-item) has a team"

PR #1332, opened 2020-01-03
Control when learners can delete files as members of a team

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-4854
https://openedx.atlassian.net/browse/EDUCATOR-4852
https://openedx.atlassian.net/browse/EDUCATOR-4795

This also deletes the `FileUpload.url_descriptor_tuple()` method and transitions its responsibility to two new methods: `file_descriptor_tuples()` and `team_file_descriptor_tuples()`.

*Testing Instructions*
* You'll have to checkout this edx-platform branch if it hasn't been merged yet: https://github.com/edx/edx-platform/pull/22834
* Set up a team-enabled ORA block that accepts file uploads as part of submission.

*Pictures*
Staff user's view, after sharing 2 files themselves and the verified user (on the same team) having shared 1 file:
staff-learner-shared-files

...and the same team, but from the verified user's perspective:
verified-learner-shared-files

TODO:
- [x]  Bump version.
- [x] Handful of outstanding unit tests.


openassessment/fileupload/api.py
```diff
@@ -53,6 +53,50 @@ def get_student_file_key(student_item_dict, index=0):
...
+        Boolean indicating if the file with the given key can be deleted by the current user.
+    """
+    if not teams_enabled:
+        return True
+
+    if not shared_file:
+        try:
+            shared_file = SharedFileUpload.by_key(key)
+        except SharedFileUpload.DoesNotExist:
+            # TODO: is it reasonble to say "fine, you can delete it, ugh!"
```
jansenk: Should we at least log a message? Not like that will really do anything ... maybe we should forcibly delete the file in this state xD

openassessment/fileupload/api.py
```diff
@@ -173,15 +210,63 @@ class FileUploadManager(object):
...
     def __init__(self, openassessment_xblock):
         self.block = openassessment_xblock
+        self._cached_shared_uploads = None
 
     def get_uploads(self):
         """
         Returns:
             A list of FileUpload objects associated with an instance of an O.A. Block.
+            This will include FileUpload objects corresponding to existing files
+            *and* files that have been deleted.
```
jansenk: Are you sure? I'm pretty sure by default _file_uploads_from_list_fields ignores deleted files.

openassessment/fileupload/api.py
```diff
@@ -173,15 +210,63 @@ class FileUploadManager(object):
...
+        for upload in self.get_uploads():
+            show_delete_button = can_delete_file(
+                self.block.teams_enabled,
+                upload.key,
+                team_id,
+                shared_file=shared_uploads_by_key.get(upload.key)
+            )
+            url_tuples.append((upload.download_url, upload.description, upload.name, show_delete_button))
+
+        return url_tuples
```
jansenk: I think this will still show files from old teams

openassessment/fileupload/api.py
```diff
@@ -251,16 +336,19 @@ def create_shared_upload(self, fileupload):
...
         """
         Given a file index to remove, null out its metadata in our stored file metadata fields
 
         Args:
-            index: file index to remove
-
-        Returns: newly updated FileUpload records.
+            index (integer): file index to remove
+            delete_shared_upload (boolean): If true, will also cause any related ``SharedFileUpload`` records
+                to be deleted. Defaults to False.
```
jansenk: Why would the default be false? We're calling the API to actually delete the file here no matter what, why would we ever want to keep the SharedFileUpload pointing at nothing?

openassessment/fileupload/tests/test_api_functions.py
```diff
@@ -0,0 +1,130 @@
...
+from openassessment.assessment.models.base import SharedFileUpload
+from openassessment.fileupload import api
+
+
+@pytest.fixture
+def shared_file_upload_fixture():
+    """
+    Helper method to create a SharedFileUpload object and
+    delete it on exit.
+    """
```
jansenk: neat

openassessment/xblock/submission_mixin.py
```diff
@@ -386,16 +396,7 @@ def _get_student_item_key(self, num=0):
...
-        num = int(num)
-        if num > 0:
-            student_item_dict['num'] = num
-            return u"{student_id}/{course_id}/{item_id}/{num}".format(
-                **student_item_dict
-            )
-        return u"{student_id}/{course_id}/{item_id}".format(
-            **student_item_dict
-        )
+        return file_upload_api.get_student_file_key(self.get_student_item_dict(), index=num)
```
jansenk: 👍

openassessment/xblock/team_mixin.py
```diff
@@ -41,6 +41,10 @@ def team(self):
         team = self.teams_service.get_team(user, self.course_id)
         return team
 
+    @cached_property
```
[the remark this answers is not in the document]
iloveagent57: probably not, but actually, I just need to pluck the `team_id` off of the block's `team` property, so I don't need this at all.

openassessment/fileupload/api.py
iloveagent57: yeah, added a log statement

openassessment/fileupload/api.py
iloveagent57: oh, hmm, I reasoned myself into this to convince myself that we didn't have a bug in the file display template, but now I think we might

openassessment/fileupload/api.py
iloveagent57: 👍

openassessment/fileupload/api.py
```diff
@@ -174,15 +219,156 @@ class FileUploadManager(object):
...
+            return self._uploads_owned_by_current_user_and_team(user_uploads)
+
+        return user_uploads
+
+    def get_team_uploads(self):
+        if self.block.is_team_assignment():
+            return self._uploads_owned_by_teammates()
+        return []
+
+    def _uploads_owned_by_current_user_and_team(self, user_uploads):
```
jansenk: Maybe rename this function? Based on the name I would assume this would also return teammates shared uploads

openassessment/fileupload/api.py
```diff
@@ -231,10 +417,13 @@ def append_uploads(self, *new_uploads):
                     existing_file_sizes
                 )
             }
-            for new_file_upload in new_file_uploads:
-                if new_file_upload.key not in existing_file_upload_key_set:
-                    self.create_shared_upload(new_file_upload)
 
+            if self.block.has_team():
```
jansenk: wanna just add this to the `if` on 411?

openassessment/templates/openassessmentblock/oa_team_uploaded_files.html
```diff
@@ -0,0 +1,38 @@
...
+{% if file_upload_type %}
+  <div class="{{ class_prefix }}__display__file {% if not team_file_urls %}is--hidden{% endif %} submission__{{ file_upload_type }}__upload" data-upload-type="{{ file_upload_type }}">
+        {% if team_file_urls %}
+        <h5 class="submission__answer__part__text__title">
+          {% trans "Files that were uploaded by your teammates:" %}
+        </h5>
+        {% endif %}
+
+        <div class="submission__answer__files">
+        {% for file_url, file_description, file_name, show_delete_button in team_file_urls %}
```
Andytr1: Just checking: since these files are team-mate owned, the delete button will never be displayed?
Also, I may have missed, but would it make sense to have a unit tests that checks for correct rendering of the page?

openassessment/xblock/waffle_mixin.py
```diff
@@ -66,6 +68,7 @@ def is_feature_enabled(self, flag):
 
         return False
 
+    @cached_property
```
Andytr1: What is the benefit of making it a cached property?

openassessment/fileupload/api.py
```diff
@@ -153,6 +195,9 @@ def __eq__(self, other):
         return self._to_dict() == other._to_dict()  # pylint: disable=protected-access
 
 
+FileDescriptor = namedtuple('FileDescriptor', ['download_url', 'description', 'name', 'show_delete_button'])
```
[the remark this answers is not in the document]
iloveagent57: 👨‍🍳 💋

openassessment/fileupload/api.py
iloveagent57: 👍

openassessment/fileupload/api.py
iloveagent57: 👍

openassessment/templates/openassessmentblock/oa_team_uploaded_files.html
iloveagent57: yeah, exactly, you can't delete files that someone else is the owner/original-uploader of

openassessment/xblock/waffle_mixin.py
iloveagent57: waffle switches are _really_ slowly changing, so if we've looked up the value of a switch and flag once during the lifetime of an ORA block, it shouldn't be necessary to do it again.

PR #1334, opened 2020-01-06
EDUCATOR-4810: Team submission generates submission records for all team members

nsprenkle:
## Summary
- Work for [EDUCATOR-4810](https://openedx.atlassian.net/browse/EDUCATOR-4810): Team submission generates submission records for all team members
- Version bump to `2.6.6`
- Dependency bump
- Add pytest_cache to .gitignore
- Add missing error message surfacing in `submit` function

## Dependencies
Relies on workaround in `edx-platform` to allow for anonymous user ID lookups by teammates.
See [edx-platform #22868](https://github.com/edx/edx-platform/pull/22868)

## To-do
- [x] Determine missing edge cases and error-handling (deliver as follow-on story)
- [x] Finish updating tests
- [x] Link `edx-platform` PR
- [x] Leverage team mixin changes from open PR, #1331
- [x] Verify that a submission by one team member locks further responses in the UI from all teammates
- [x] Get team shared files for upload
- [x] Submit using anonymized usernames
- [x] Fix "unable to load" on first page load bug

FYI @edx/masters-devs


openassessment/xblock/submission_mixin.py
```diff
@@ -250,8 +276,24 @@ def save_files_descriptions(self, data, suffix=''):  # pylint: disable=unused-ar
 
         return {'success': True, 'msg': u''}
 
+    def create_team_submission(self, student_item_dict, student_sub_data):
+        """ A student submitting for a team should generate matching submissions for every member of the team """
+        team_dict = self.get_team_info()
+
```
iloveagent57: Somewhere in here, you'll want to grab the keys of all files that are shared amongst this team.

openassessment/xblock/submission_mixin.py
```diff
@@ -250,8 +276,24 @@ def save_files_descriptions(self, data, suffix=''):  # pylint: disable=unused-ar
...
+        if team_dict is not None:
+            submissions = []
+
+            team_usernames = team_dict["team_usernames"]
+
+            # submissions are identical except for the student ID
+            for student_id in team_usernames:
```
iloveagent57: Here you'll want user ids (or maybe anonymized ids?) instead of usernames.  You'll have to modify the `get_team_info()` method to get it.  We can chat about how to do this.

openassessment/xblock/test/test_submission.py
```diff
@@ -423,6 +423,55 @@ def test_get_team_info(self, xblock):
...
+        xblock.get_team_info.return_value = {'team_name': 'Red Squadron',
+                                             'team_usernames': team_usernames,
+                                             'team_url': 'rebel_alliance.org'}
+
+        # when the learner submits an open assessment response
+        response = self.request(
+            xblock, 'submit', self.SUBMISSION, response_format='json')
+
+        # then the submission is successful for all members of a team
+        self.assertEquals(len(team_usernames), len(response))
```
mattcarter: This is the complaint that the quality tests are raising.

openassessment/xblock/submission_mixin.py
iloveagent57: ...and then jam them into each `student_sub_data` dict.

openassessment/xblock/test/test_submission.py
```diff
@@ -423,6 +423,55 @@ def test_get_team_info(self, xblock):
...
+    @scenario('data/basic_scenario.xml', user_id='Red Five')
+    def test_team_submission(self, xblock):
+        """ If teams are enabled, a submission by any member should submit for each member of the team """
+
+        # given a learner is on a team
+        xblock.teams_enabled = True
+        xblock.get_team_info = Mock()
+
+        team_usernames = ['Red Leader', 'Red Two', 'Red Five']
+        xblock.get_team_info.return_value = {'team_name': 'Red Squadron',
```
Andytr1: I like the variable naming!

openassessment/xblock/submission_mixin.py
nsprenkle: This wraps the standard `create_submission` call which I believe already covers that part but please confirm

nsprenkle: All it does is modifies the `student_item_dict` to change the student IDs

iloveagent57: That `self.file_manager.get_uploads()` call in `create_submission` doesn't do anything with `SharedFileUploads` yet - we'll have to take care of that in this PR (although it's highly likely that Jansen will need to do very similar work for the next ticket he picks up).

iloveagent57: That `self.file_manager.get_uploads()` call in `create_submission` doesn't do anything with `SharedFileUploads` yet - we'll have to take care of that in this PR (although it's highly likely that Jansen will need to do very similar work for the next ticket he picks up).

openassessment/xblock/submission_mixin.py
```diff
@@ -113,9 +116,35 @@ def submit(self, data, suffix=''):  # pylint: disable=unused-argument
...
+                if self.teams_enabled:
+                    # a submission for a team generates matching submissions for all members
+                    submissions = self.create_team_submission(student_item_dict, student_sub_data)
+
+                    result = []
+
+                    for submission in submissions:
+                        status = True
+                        status_tag = submission.get('student_item')
+                        status_text = submission.get('attempt_number')
```
Andytr1: Consider making this into a helper function, since the same block is also duped below

openassessment/xblock/submission_mixin.py
```diff
@@ -250,24 +277,32 @@ def save_files_descriptions(self, data, suffix=''):  # pylint: disable=unused-ar
 
         return {'success': True, 'msg': u''}
 
+    def create_team_submission(self, student_item_dict, student_sub_data):
+        """ A student submitting for a team should generate matching submissions for every member of the team """
```
Andytr1: nit: missing period at the end of sentence.

openassessment/xblock/submission_mixin.py
```diff
@@ -250,24 +277,32 @@ def save_files_descriptions(self, data, suffix=''):  # pylint: disable=unused-ar
...
+            # submissions are identical except for the student ID
+            for student_id in team_usernames:
+                student_item_dict["student_id"] = student_id
+                submissions.append(self.create_submission(student_item_dict, student_sub_data))
+
+        return submissions
+
     def create_submission(self, student_item_dict, student_sub_data):
-        """ Creates submission for the submitted assessment response. """
+        """ Creates submission for the submitted assessment response or a list for a team assessment """
```
Andytr1: nit: period at the end of comment.

openassessment/xblock/submission_mixin.py
```diff
@@ -288,6 +323,42 @@ def create_submission(self, student_item_dict, student_sub_data):
 
         return submission
 
+    def collect_files_for_submission(self, student_sub_dict):
+        """ collect files from CSM for individual submisisons or SharedFileUpload for team submisisons """
```
Andytr1: nit: Proper capitalization and punctuation :)

openassessment/xblock/submission_mixin.py
```diff
@@ -288,6 +323,42 @@ def create_submission(self, student_item_dict, student_sub_data):
...
+                student_sub_dict['file_keys'].append(file_upload.key)
+                student_sub_dict['files_descriptions'].append(file_upload.description)
+                student_sub_dict['files_names'].append(file_upload.name)
+                student_sub_dict['files_sizes'].append(file_upload.size)
+
+        return student_sub_dict
+
+    def get_team_files(self):
+        """ Team files, unlike individual submissions, are gathered from the SharedFileUpload table
+            and identified by their team ID, course ID, and item ID
```
Andytr1: nit: missing period

openassessment/xblock/submission_mixin.py
```diff
@@ -288,10 +327,46 @@ def create_submission(self, student_item_dict, student_sub_data):
...
+                student_sub_dict['files_descriptions'].append(team_file_upload.description)
+                student_sub_dict['files_names'].append(team_file_upload.item_id)
+                student_sub_dict['files_sizes'].append(team_file_upload.size)
+
+        else:
+            for file_upload in self.file_manager.get_uploads():
+                student_sub_dict['file_keys'].append(file_upload.key)
+                student_sub_dict['files_descriptions'].append(file_upload.description)
+                student_sub_dict['files_names'].append(file_upload.name)
+                student_sub_dict['files_sizes'].append(file_upload.size)
```
Andytr1: What about resolving the duplication of the two blocks by creating a method like:
`def create_submission_entry(self,keys,descriptions,names.sizes):` which would set the dictionary to the data supplies depending on the block?

openassessment/xblock/submission_mixin.py
```diff
@@ -113,9 +116,25 @@ def submit(self, data, suffix=''):  # pylint: disable=unused-argument
...
+                # a submission for a team generates matching submissions for all members
+                if self.teams_enabled:
+                    submissions = self.create_team_submission(student_sub_data)
+
+                    results = []
+
+                    for submission in submissions:
+                        results.append(self._create_submission_response(submission))
+
+                    return results
```
iloveagent57: You can shorten this whole thing to:
```python
return [self._create_submission_response(submission) for submission in submissions]
```

openassessment/xblock/submission_mixin.py
```diff
@@ -148,13 +168,22 @@ def submit(self, data, suffix=''):  # pylint: disable=unused-argument
                 logger.exception(msg)
                 status_tag = 'EUNKNOWN'
                 status_text = self._(u'API returned unclassified exception.')
-            else:
-                status = True
-                status_tag = submission.get('student_item')
-                status_text = submission.get('attempt_number')
```
iloveagent57: Why did you delete this?

openassessment/xblock/submission_mixin.py
```diff
@@ -250,24 +279,34 @@ def save_files_descriptions(self, data, suffix=''):  # pylint: disable=unused-ar
...
+            submissions = []
+
+            # submissions are identical except for the student item
+            for anonymous_user_id in team_anonymous_user_ids:
+                student_item_dict = self.get_student_item_dict(anonymous_user_id=anonymous_user_id)
+                submissions.append(self.create_submission(student_item_dict, student_sub_data))
+
+        # Since we don't get the right UUID through the above loop,
+        # null it out and fetch it from the DB during the next page load
+        self.submission_uuid = None
```
iloveagent57: I think you could get the correct submission_uuid if you make that change I talked about in your edx-platform PR - the one about `get_team_anonymous_user_ids()` returning a map from plain user id to their anon. id.

openassessment/xblock/submission_mixin.py
```diff
@@ -288,10 +327,46 @@ def create_submission(self, student_item_dict, student_sub_data):
 
         return submission
 
+    def collect_files_for_submission(self, student_sub_dict):
```
iloveagent57: I like that you made this a helper method.  Since it's meant to be "private" (i.e., only this class uses it directly), you should preface the method name with a `_`.

openassessment/xblock/submission_mixin.py
iloveagent57: This could be shortened as:
```python
if self.teams_enabled:
    uploads = self.get_team_files()
else:
    uploads = self.get_uploads()
for upload in uploads:
    student_sub_dict['file_keys'].append(upload.file_key)
    student_sub_dict['files_descriptions'].append(upload.description)
    student_sub_dict['files_names'].append(upload.item_id)
    student_sub_dict['files_sizes'].append(upload.size)
```

openassessment/xblock/submission_mixin.py
iloveagent57: So, in light of our conversation on your other PR... :)
I think you can get the anon. id of the current user at any time using `self.xmodule_runtime.anonymous_student_id`, so you could check that against the current anon. id in your loop, and if they match, set `self.submission_uuid` to the current submission's UUID.

openassessment/xblock/submission_mixin.py
nsprenkle: That's now moved to `_create_submisison_response()`, which is called in both paths of the `try:` block earlier in the funciton.

The only remaining codepath is exception handling which explicitly overrides those tags so I thought it was redundant.

openassessment/xblock/submission_mixin.py
nsprenkle: There's slight key differences between the two upload types (e.g. `file_upload.name` correlates to `team_file_upload.item_id`) which requires the difference

iloveagent57: ah, right, that's unfortunate.  I've actually got some code in my open PR that will make this easier, so let's just leave it for now, we can do a quick follow-on PR later.

mattcarter: if you're tolerant to a healthy dose of _clever_, you could avoid altering the model objects and use the builtin [`getattr`](https://docs.python.org/3/library/functions.html#getattr) to make compound statements when populating the `student_sub_dict`.

openassessment/xblock/submission_mixin.py
```diff
@@ -113,9 +115,20 @@ def submit(self, data, suffix=''):  # pylint: disable=unused-argument
 
         status_tag = 'ENOMULTI'  # It is an error to submit multiple times for the same item
         status_text = self._(u'Multiple submissions are not allowed.')
+
         if not workflow:
             try:
-                submission = self.create_submission(student_item_dict, student_sub_data)
+
+                # a submission for a team generates matching submissions for all members
+                if self.teams_enabled:
```
iloveagent57: I think you might actually want `self.is_team_assignment()` here (which also checks `self.team_submissions_enabled`).

openassessment/xblock/submission_mixin.py
```diff
@@ -250,24 +273,36 @@ def save_files_descriptions(self, data, suffix=''):  # pylint: disable=unused-ar
...
+            for anonymous_user_id in team_anonymous_user_ids:
+                student_item_dict = self.get_student_item_dict(anonymous_user_id=anonymous_user_id)
+                submission = self.create_submission(student_item_dict, student_sub_data)
+
+                if anonymous_user_id == submitter_anonymous_user_id:
+                    self.submission_uuid = submission["uuid"]
+
+                submissions.append(submission)
+
+        return submissions
```
iloveagent57: If, somehow, you get here and `self.has_team()` returns false, `return submissions` will result in a `NameError`, because `submissions` is only defined inside the if block.

openassessment/xblock/submission_mixin.py
```diff
@@ -75,8 +75,10 @@ def submit(self, data, suffix=''):  # pylint: disable=unused-argument
             suffix (str): Not used in this handler.
 
         Returns:
-            (tuple): Returns the status (boolean) of this request, the
+            (tuple | [tuple]): Returns the status (boolean) of this request, the
                 associated status tag (str), and status text (unicode).
+                This becomes an array of similarly structured tuples in the event
```
iloveagent57: A thought for future refactoring: should this method always return a list of tuples, to make processing more consistent from the client's perspective?

openassessment/xblock/submission_mixin.py
```diff
@@ -250,24 +273,36 @@ def save_files_descriptions(self, data, suffix=''):  # pylint: disable=unused-ar
...
+
+            # submissions are identical except for the student item
+            for anonymous_user_id in team_anonymous_user_ids:
+                student_item_dict = self.get_student_item_dict(anonymous_user_id=anonymous_user_id)
+                submission = self.create_submission(student_item_dict, student_sub_data)
+
+                if anonymous_user_id == submitter_anonymous_user_id:
+                    self.submission_uuid = submission["uuid"]
+
+                submissions.append(submission)
```
iloveagent57: I like how this came together.

openassessment/xblock/submission_mixin.py
```diff
@@ -250,24 +273,36 @@ def save_files_descriptions(self, data, suffix=''):  # pylint: disable=unused-ar
...
+
+        return submissions
+
     def create_submission(self, student_item_dict, student_sub_data):
-        """ Creates submission for the submitted assessment response. """
+        """ Creates submission for the submitted assessment response or a list for a team assessment. """
```
iloveagent57: Same note here about potential refactoring.

openassessment/xblock/submission_mixin.py
```diff
@@ -288,10 +323,31 @@ def create_submission(self, student_item_dict, student_sub_data):
...
+    def _collect_files_for_submission(self, student_sub_dict):
+        """ Collect files from CSM for individual submisisons or SharedFileUpload for team submisisons. """
+
+        if not self.file_upload_type:
+            return
+
+        for field in ('file_keys', 'files_descriptions', 'files_names', 'files_sizes'):
+            student_sub_dict[field] = []
+
+        if self.teams_enabled:
```
iloveagent57: probably `self.is_team_assignment()`.

openassessment/xblock/submission_mixin.py
```diff
@@ -288,10 +323,31 @@ def create_submission(self, student_item_dict, student_sub_data):
...
+            uploads = self.file_manager.get_team_uploads()
+        else:
+            uploads = self.file_manager.get_uploads()
+        for upload in uploads:
+            student_sub_dict['file_keys'].append(upload.key)
+            student_sub_dict['files_descriptions'].append(upload.description)
+            student_sub_dict['files_names'].append(upload.name)
+            student_sub_dict['files_sizes'].append(upload.size)
```
iloveagent57: Yeah, this came together much more nicely

openassessment/xblock/test/test_submission.py
```diff
@@ -360,6 +360,99 @@ def test_get_student_username(self, xblock):
...
+
+        xblock.xmodule_runtime = Mock(
+            course_id='test_course',
+            anonymous_student_id='r5'
+        )
+
+        mock_team = {'team_id': 'rs-04',
+                     'team_name': 'Red Squadron',
+                     'team_usernames': ['Red Leader', 'Red Two', 'Red Five'],
+                     'team_url': 'rebel_alliance.org'}
```
iloveagent57: Nit: break things over multiple lines as
```python
mock_team = {
    'team_id': 'rs-04',
    'team_name': 'Red Squadron',
    'team_usernames': ['Red Leader', 'Red Two', 'Red Five'],
    'team_url': 'rebel_alliance.org'
}
```

openassessment/xblock/test/test_submission.py
```diff
@@ -360,6 +360,99 @@ def test_get_student_username(self, xblock):
...
+    @scenario('data/basic_scenario.xml', user_id='Red Five')
+    def test_team_submission(self, xblock):
+        """ If teams are enabled, a submission by any member should submit for each member of the team """
+
+        # given a learner is on a team
+        mock_team = self.setup_mock_team(xblock)
+
+        # when the learner submits an open assessment response
+        response = self.request(
+            xblock, 'submit', self.SUBMISSION, response_format='json')
```
iloveagent57: Also tiny nit (which I wont' repeat below): put the closing paren on its own line.

openassessment/xblock/test/test_submission.py
```diff
@@ -774,6 +867,35 @@ def test_open_saved_response_old_format(self, xblock):
...
+        """ When a submission is created for a team, we create identical submissions for each learner.
+            Since we can't save submission info to other learners state, we need to query the database
+            on page load to see if a submisison has been created by a team member.
+        """
+        SubmissionTest.setup_mock_team(xblock)
+        submissions = xblock.create_team_submission(
+            ('A man must have a code', 'A man must have an umbrella too.')
+        )
+        # get submission from Xblock
+        submission = [_ for _ in submissions if _['uuid'] == xblock.submission_uuid][0]
```
iloveagent57: You should only use `_` as a variable name when you really don't need it.  In here, you need it for the conditional portion of your list comprehension.  Maybe just `sub` instead?

openassessment/xblock/workflow_mixin.py
```diff
@@ -126,13 +126,47 @@ def get_workflow_info(self, submission_uuid=None):
...
-                return {}
+            submission_uuid = self.try_get_submission_uuid()
+
+        if submission_uuid is None:
+            return {}
         return workflow_api.get_workflow_for_submission(
             submission_uuid, self.workflow_requirements()
         )
 
+    def try_get_submission_uuid(self):
```
iloveagent57: Just `get_submission_uuid()` is fine - it's pretty common to have methods that return either "the thing" or None, and we call them `get_the_thing()`.

openassessment/xblock/submission_mixin.py
nsprenkle: Could we remove the `self.has_team()` check since that's a path that can't be hit currently? Or have an overabundance of caution? Throw an exception? Return `None`?

openassessment/xblock/submission_mixin.py
nsprenkle: Admittedly, the result is surfaced to the client but never actively read/used. I'd be even happier to return a dict or something more useful than a tuple so that it would be more helpful if the client ever decided to use that info.

PR #1335, opened 2020-01-06
EDUCATOR-4845: Team ORA File Submission: Create new SharedFileUpload records when new files are uploaded

jansenk:
[EDUCATOR-4845](https://openedx.atlassian.net/browse/EDUCATOR-4845)

When files are uploaded in a Team ORA, create SharedFileUpload records for the uploads as well.
Also add migration to add name column to SharedFileUpload


openassessment/assessment/models/base.py
```diff
@@ -848,6 +848,7 @@ class SharedFileUpload(TimeStampedModel):
     history = HistoricalRecords()
     description = models.TextField(default=u"", blank=True)
     size = models.BigIntegerField(default=0, blank=True)
+    name = models.CharField(max_length=255, default=u"")
```
Andytr1: Would it make sense to call it file_name instead?

openassessment/xblock/openassessmentblock.py
```diff
@@ -614,11 +614,14 @@ def _create_fragment(self, template, context_dict, initialize_js_func, additiona
...
             "FILE_TYPE_WHITE_LIST": self.white_listed_file_types,
             "MAXIMUM_FILE_UPLOAD_COUNT": self.MAX_FILES_COUNT,
-            "TEAM_ASSIGNMENT": self.teams_enabled and self.team_submissions_enabled()
+            "TEAM_ASSIGNMENT": self.is_team_assignment()
         }
         fragment.initialize_js(initialize_js_func, js_context_dict)
         return fragment
 
+    def is_team_assignment(self):
+        return self.teams_enabled and self.team_submissions_enabled()
```
Andytr1: Nice

openassessment/fileupload/api.py
```diff
@@ -216,7 +218,38 @@ def append_uploads(self, *new_uploads):
         new_sizes = existing_file_sizes + sizes_to_add
         self._set_file_sizes(new_sizes)
 
-        return self._file_uploads_from_list_fields(new_descriptions, new_names, new_sizes)
+        new_file_uploads = self._file_uploads_from_list_fields(new_descriptions, new_names, new_sizes)
+
+        if self.block.is_team_assignment():
+            existing_file_upload_key_set = {
```
iloveagent57: Another way to do this would be `existing_file_upload_key_set = {upload.key for upload in self.get_uploads()}` and you just have to invoke it prior to line 213.

openassessment/fileupload/tests/test_file_management.py
```diff
@@ -0,0 +1,168 @@
...
+    def get_student_item_dict(self):
+        return dict(
+            student_id=self.student_id,
+            item_id=self.item_id,
+            course_id=self.course_id,
+            item_type='openassessment'
+        )
+
+
+def uploadDict(name, desc, size):
```
iloveagent57: you have strangely transitioned to camel case here and in some parameters below.

openassessment/fileupload/api.py
iloveagent57: ...not that it really helps that much, but maybe makes it a tiny bit prettier...

jansenk: I don't think that will work. Since we've already updated the size/descriptions/names that would include the new files as well.

PR #1336, opened 2020-01-09
upgrade requirements and remove some pins after django 3 upgrade

jansenk:


requirements/constraints.txt
```diff
@@ -8,18 +8,16 @@ bleach<=2.1.4                      # Constrained by edx-platform/edx-enterprise/
 boto<=2.39.0                       # Constrained by edx-platform and its dependencies.
 defusedxml<=0.5.0                  # Constrained by edx-platform and its dependencies.
 django-model-utils==3.0.0          # Constrained by edx-platform and its dependencies.
+Django<1.12                        # Constrained by edx-platform
```
jansenk: Django was previously not constrained and was only constrained by python version. 
The test environment requirements are defined in tox.ini and I am slightly concerned about this conflicting.
That said, since the LMS pins django here, I thought this was definitely important to do

PR #1338, opened 2020-01-09
Don't rely on iterator count for uploaded file index.

mattcarter:
The filenum value for a file may be different than its order in the set.  Have to account for files that have been deleted

### DIFF
--- a/openassessment/templates/openassessmentblock/oa_uploaded_file.html
+++ b/openassessment/templates/openassessmentblock/oa_uploaded_file.html
@@ -12,15 +12,15 @@
 
     <div class="{{ class_prefix }}__display__file {% if not file_urls %}is--hidden{% endif %} submission__{{ file_upload_type }}__upload" data-upload-type="{{ file_upload_type }}">
         <div class="submission__answer__files">
-        {% for file_url, file_description, file_name in file_urls %}
+        {% for file_url, file_description, file_name, file_index in file_urls %}
             <div class="submission__answer__file__block submission__answer__file__block__{{ forloop.counter0 }}" {% if not file_url %} deleted {% endif %}>
             {% if file_url %}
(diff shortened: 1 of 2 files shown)


jansenk:
Could you toss a test into test_file_managers where you create some files and then delete some and assert that the indices are still correct?

openassessment/templates/openassessmentblock/oa_uploaded_file.html
```diff
@@ -12,15 +12,15 @@ <h5 class="{{ class_prefix }}__display__title">
 
     <div class="{{ class_prefix }}__display__file {% if not file_urls %}is--hidden{% endif %} submission__{{ file_upload_type }}__upload" data-upload-type="{{ file_upload_type }}">
         <div class="submission__answer__files">
-        {% for file_url, file_description, file_name in file_urls %}
+        {% for file_url, file_description, file_name, file_index in file_urls %}
             <div class="submission__answer__file__block submission__answer__file__block__{{ forloop.counter0 }}" {% if not file_url %} deleted {% endif %}>
```
jansenk: this line here as well as 28 also use the forloop counter. i vote we completely replace it.

mattcarter:
I'm not sure how we can test the template w/o bokchoy.  Thoughts?

iloveagent57:
You can write a python unit test that calls the `render_submission()` method and inspects the HTML that it returns.  For instance, if you expect to have some files with indices 1 and 5, you could assert that the returned HTML contains the string `"class=submission__answer__file__block submission__answer__file__block__1"` and `"class=submission__answer__file__block submission__answer__file__block__5"`.  I would guess there's other places in this repo where we do that type of unit test, it's a fairly common way of testing Django templates.

mattcarter:
Work done in #1340

PR #1339, opened 2020-01-10
EDUCATOR-4858 setup: track teamset id rather than name 

jansenk:
Some changes that were needed for [EDUCATOR-4858](https://openedx.atlassian.net/browse/EDUCATOR-4858)

I thought they were decent enough to chunk into their own PR.

- Track the ID of the teamset/topic rather than name since that's what we'll need for database lookups
- Changes to the template to reflect the above and to change the `value`s of the dropdowns
- A bugfix i thought I'd throw in here rather than opening a new ticket

### DIFF
--- a/openassessment/xblock/openassessmentblock.py
+++ b/openassessment/xblock/openassessmentblock.py
@@ -269,10 +269,10 @@ class OpenAssessmentBlock(MessageMixin,
         help="Whether team submissions are enabled for this case study.",
     )
 
-    selected_teamset_name = String(
+    selected_teamset_id = String(
         default=u"",
         scope=Scope.settings,
-        help="The name of the selected teamset.",
+        help="The id of the selected teamset.",
     )
 
     @property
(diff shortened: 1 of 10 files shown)


openassessment/xblock/openassessmentblock.py
```diff
@@ -1186,6 +1186,8 @@ def get_real_user(self, anonymous_user_id):
 
         """
         if hasattr(self, "xmodule_runtime"):
+            if self.xmodule_runtime.get_real_user is None:  # pylint: disable=no-member
```
jansenk: I was running into this on studio, when the page is rendered as you are adding assignments. I assumed that was what `is_studio_preview` is, but that check was returning false? Anyways, the above line was true but `self.xmodule_runtime.get_real_user` was `None` so the call was blowing up. 🤷‍♂

mattcarter:
👍

PR #1344, opened 2020-01-17
EDUCATOR-4870: Change 'teamless' message to include teamset name

jansenk:
[EDUCATOR-4870](https://openedx.atlassian.net/browse/EDUCATOR-4870)
@edx/masters-devs 

When a user tries to view a team-based ora for a teamset they are not in a team for, they are shown a screen telling them that they must be on a team before they can view this ORA.

This PR adds in several references to the name of the teamset that the user must be in, since in the future users will be able to be on multiple teams in a course, and getting a message saying "you aren't on a team" when you may be on several teams is unhelpful.

There is also a slight rework of some of the "mocking" in test_teams, since it turns out that some of the tests were just testing "mocked" behavior rather than actual behavior.

- [ ] https://github.com/edx/edx-platform/pull/22893 must be merged before this can be merged

Screen Shot 2020-01-17 at 11 03 15 AM

### DIFF
--- a/openassessment/templates/openassessmentblock/message/oa_message_no_team.html
+++ b/openassessment/templates/openassessmentblock/message/oa_message_no_team.html
@@ -3,7 +3,13 @@
 <div id="openassessment__message__{{ xblock_id }}" class="message--warning openassessment__message message">
     <h4 class="message__title">{% trans "Team membership is required to view this assignment" %} </h4>
     <div class="message__content">
-        <p> {% trans "This is a team assignment for team set {{teamset_name}}. You are currently not on a team in team set {{teamset_name}}. You must be on a team in team set {{teamset_name}} to access this team assignment." %}</p>
+        <p> 
+            {% blocktrans %}
+            This is a team assignment for team set "{{ teamset_name }}".
+            You are currently not on a team in team set "{{teamset_name}}".
+            You must be on a team in team set "{{teamset_name}}" to access this team assignment.
+            {% endblocktrans %}
+        </p>
     </div>
 </div>
 {% endspaceless %}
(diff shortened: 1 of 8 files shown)


jansenk:
before anyone else says it: I know, bump the version

openassessment/templates/openassessmentblock/message/oa_message_no_team.html
```diff
@@ -3,7 +3,13 @@
 <div id="openassessment__message__{{ xblock_id }}" class="message--warning openassessment__message message">
     <h4 class="message__title">{% trans "Team membership is required to view this assignment" %} </h4>
     <div class="message__content">
-        <p> {% trans "This is a team assignment. You are currently not on a team. You must be on a team to access this team assignment." %}</p>
+        <p> 
+            {% blocktrans %}
+            This is a team assignment for team set "{{ teamset_name }}".
+            You are currently not on a team in team set "{{teamset_name}}".
+            You must be on a team in team set "{{teamset_name}}" to access this team assignment.
```
iloveagent57: The tiniest possible nit: use consistent spacing between braces and variable names.

openassessment/xblock/team_mixin.py
```diff
@@ -22,6 +22,15 @@ def teams_service(self):
             return self.runtime.service(self, 'teams')
         except NoSuchServiceError:
             logger.error(u'{}: Teams service unavailable'.format(self.location))
+            raise
+
+    @cached_property
+    def teams_config_service(self):
```
iloveagent57: The second tiniest possible nit: should we name this (and other things below) using the full word "configuration" so we don't have to think too much about any possible differences between the service key on the LMS side and what things are named on the ORA side?

jansenk: that seems like a good idea.

PR #1357, opened 2020-02-11
[PROD-515] - ORA xBlock color contrast issues (WCAG 2.1)

saadyousafarbi:
#### Part 1:
The textarea for Your Response has a border with color ##b9c1c9.  The background color is #f5f6f7.  These colors have luminance contrast of 1.68:1.  WCAG 2.1 has a luminance contrast minimum of 3.0:1 for interactive elements, so the border needs to be darker. I suggest #7C8F9C here, which has contrast of 3.09:1 vs. the #f5f6f7 background.

#### Part 2:
Also, the radio buttons under assessments sections use #A6A6A6 for the border color for the radio buttons. That color has 2.43:1 vs. #ffffff background. To be >=3.0:1 I recommend #949494, which is 3.03:1 vs. #ffffff. (Not done in line with [PROD-513](https://openedx.atlassian.net/browse/PROD-513))

#### Text area border change
text_area_change

### Sandbox
https://pr23111.sandbox.edx.org/

### Reviewers
 - [ ] @awaisdar001 

### Post Review
 - [ ] Squash & Rebase commits


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "edx-ora2",
-  "version": "2.6.7",
+  "version": "2.6.8",
```
awaisdar001: Why this has changed?

awaisdar001: Pinging again ^^

saadyousafarbi:
@awaisdar001 yes, I have generated the files after the change!

openassessment/xblock/static/sass/oa/views/_oa-base.scss
```diff
@@ -547,6 +547,8 @@
 
   .submission__answer__part__text__value {
     margin: ($baseline-v/2) 0;
+    border-color: #7C8F9C;
```
awaisdar001: Can we put this color in a reusable variable? and use its name here instead?

openassessment/xblock/static/sass/xb/utilities/_variables.scss
```diff
@@ -39,6 +39,7 @@ $gray-d3: shade($gray,60%);
 $gray-d4: shade($gray,80%);
 $gray-d5: shade($gray,90%);
 $gray-d6: shade($gray,95%);
+$gray-border: rgb(124, 143, 156);
```
awaisdar001: should be just `gray`

saadyousafarbi: that is already declared within variables.

PR #1358, opened 2020-02-14
EDUCATOR-4914: The oa_response_submitted template should include team file URLs in its context

jansenk:
[EDUCATOR-4914: The oa_response_submitted template should include team file URLs in its context](https://openedx.atlassian.net/browse/EDUCATOR-4914)

The teammate files were already in the context, but we needed to add them to the template.

These are screenshots showing the change. They look the same, but one is the oa_response_submitted template and one is the oa_response_graded template. 

Screen Shot 2020-02-14 at 12 28 19 PM
Screen Shot 2020-02-14 at 12 31 59 PM


jansenk:
Ignore the random "1" in the one screenshot, that's not there anymore and was put there to make suer we actually were rendering different templates 😆

PR #1363, opened 2020-02-20
EDUCATOR-4911: Log exceptions for any error state in ORA file upload XBlock endpoints

jansenk:
[EDUCATOR-4911: Log exceptions for any error state in ORA file upload XBlock endpoints](https://openedx.atlassian.net/browse/EDUCATOR-4911)

add some more logs


jansenk:
The test failed because the return value of get_file_url is now being evaluated for truthiness, and since there's no explicit return_value defined, it adds another call to the mock to check truthiness

PR #1120, opened 2018-07-26
Fix backward compatibility for older blocks' required text response

kluo:
Older Ora2 blocks that required text responses but had optional file uploads would lose the required text response field because its checking `file_upload_response` instead of `file_upload_response_raw`

PR #1366, opened 2020-03-02
Older Ora2 blocks that required text responses but had optional file …

iloveagent57:
…uploads would lose the required text response field because its checking file_upload_response instead of file_upload_response_raw

Re-build of this PR, for testing's sake: https://github.com/edx/edx-ora2/pull/1120

**What this changes**
The `text_response` property of the ORA XBlock class.  It currently looks at `file_upload_response`, which is an evaluated property, rather than the direct attribute `file_upload_response_raw` that older ora2 blocks did not specify. The problem is that checking `file_upload_response` isn't strictly supporting backward compatibility, for the following reason: when an old ora2 block has required text responses and optional file uploads, checking `file_upload_response` returns `'optional'` which causes `text_response` to evaluate as `None` rather than `'required'`.

### DIFF
--- a/openassessment/xblock/openassessmentblock.py
+++ b/openassessment/xblock/openassessmentblock.py
@@ -285,7 +285,7 @@ class OpenAssessmentBlock(MessageMixin,
         Backward compatibility for existing blocks that were created without text_response
         or file_upload_response fields. These blocks will be treated as required text.
         """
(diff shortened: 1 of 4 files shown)


nsprenkle:
Is there a test that reflects this change?

PR #1377, opened 2020-03-27
PROD-1376 : File name None issue workaround and logs added

AhtishamShahid:
## [PROD-1376](https://openedx.atlassian.net/browse/PROD-1376)

ORA peer assessment and final grades are not loading for the learner. The issue seems to be caused by data inconsistency so added workaround and logs to fix the issue

### DIFF
--- a/openassessment/xblock/submission_mixin.py
+++ b/openassessment/xblock/submission_mixin.py
@@ -505,7 +505,11 @@ class SubmissionMixin(object):
                 file_download_url = self._get_url_by_file_key(key)
                 if file_download_url:
                     file_description = descriptions[idx].strip() if idx < len(descriptions) else ''
-                    file_name = file_names[idx].strip() if idx < len(file_names) else ''
+                    try:
+                        file_name = file_names[idx].strip() if idx < len(file_names) else ''
+                    except AttributeError:
+                        file_name = ''
+                        logger.error('descriptions[idx] is None in {}'.format(submission))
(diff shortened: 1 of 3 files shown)


openassessment/xblock/submission_mixin.py
```diff
@@ -505,7 +505,11 @@ def get_download_urls_from_submission(self, submission):
                 file_download_url = self._get_url_by_file_key(key)
                 if file_download_url:
                     file_description = descriptions[idx].strip() if idx < len(descriptions) else ''
-                    file_name = file_names[idx].strip() if idx < len(file_names) else ''
+                    try:
+                        file_name = file_names[idx].strip() if idx < len(file_names) else ''
+                    except AttributeError:
+                        file_name = ''
+                        logger.error('descriptions[idx] is None in {}'.format(submission))
```
awaisdar001: @AhtishamShahid are these temporary logs?

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "edx-ora2",
```
awaisdar001: I don't think we should merge this change.

openassessment/xblock/submission_mixin.py
AhtishamShahid: Yes, it is a temporary log that will be removed after the root cause of the issue is found.

package-lock.json
AhtishamShahid: Mean `package-lock.json` file ???

PR #1383, opened 2020-04-03
Create team assessment workflow API

nsprenkle:
**Scope:**

Create an API to extend assessment workflows to team assessments. 

See `openassessment/workflow/api.py` and `openassessment/workflow/models.py::AssessmentWorkflow` model for comparison.

**Changes:**
- Implement and test API-level functions for:
   - Creating, getting, and serializing team workflows
   - Getting status information for team workflows for a problem
- Stub model-level functions

**Out of scope:**
- Implement workflow cancellations. Since this work requires new models, it will be done in a different PR to reduce scope.
- Implement model-level functions. Out-of-scope, to be completed in [EDUCATOR-4986](https://openedx.atlassian.net/browse/EDUCATOR-4986)

**JIRA:** [EDUCATOR-4978](https://openedx.atlassian.net/browse/EDUCATOR-4978)
**FYI:** @edx/masters-devs-gta


openassessment/workflow/team_api.py
```diff
@@ -0,0 +1,139 @@
...
+        team_workflow.update_from_assessments()
+        logger.info((
+            u"Updated workflow for team submission UUID {uuid} "
+        ).format(uuid=team_submission_uuid))
+    except Exception as exc:
+        err_msg = (
+            u"Could not update team assessment workflow: {exc}"
+        ).format(exc=exc)
+        logger.exception(err_msg)
+        raise AssessmentWorkflowInternalError(err_msg)
```
Andytr1, line 64: Would it make sense to have TeamAssessmentWorfklowInternalError, so we can differentiate between vanilla assessment and team specific issues?

openassessment/workflow/team_api.py
```diff
@@ -0,0 +1,139 @@
...
+logger = logging.getLogger('openassessment.workflow.models')  # pylint: disable=invalid-name
+
+
+def create_workflow(team_submission_uuid):
+    """
+    A team submission should only be assessible by staff.  Therefore, unlike
+    the analogous `create_workflow()` method for individual submissions,
+    we don't accept `steps` or `on_init_params` as parameters to this function,
+    since those are only used to indicate which assessment steps (e.g. "peer", "self")
+    are to be included in the workflow.
```
Andytr1, line 25: Add `Raises` block to the doc string.

openassessment/workflow/team_api.py
```diff
@@ -0,0 +1,139 @@
...
+        logger.exception(err_msg)
+        raise AssessmentWorkflowInternalError(err_msg)
+
+
+def get_workflow_for_submission(team_submission_uuid):
+    """
+    Like `api.get_workflow_for_submission`, but for teams.  We don't need
+    an analogous `assessment_requirements` parameter, because team submissions
+    are only assessible by staff (where requirements like "must_grade" and
+    "must_be_graded_by" are not supported).
```
Andytr1, line 48: Same here - raises block in doc string.

openassessment/workflow/models.py
```diff
@@ -613,8 +613,37 @@ class TeamAssessmentWorkflow(AssessmentWorkflow):
     """
     Extends AssessmentWorkflow to support team based assessments.
     """
+    # Only staff assessments are supported for teams
+    STEPS = ['staff']
```
Andytr1, line 617: nit: since we can ever have one step - why not make a string and called it STAFF_STEP ?

nsprenkle: It was meant to follow the convention of `AssessmentWorkflow` and allow for extensibility in the future

openassessment/workflow/team_api.py
nsprenkle: I've intentionally left these kind of vague since I don't know what the model level will raise or if we want to rethink any error-handling. That said, I feel like creating a new error class is overkill.

Andytr1: Sounds good.

openassessment/workflow/team_api.py
```diff
@@ -0,0 +1,139 @@
...
+from openassessment.workflow.errors import (
+    AssessmentWorkflowInternalError,
+    AssessmentWorkflowRequestError,
+    AssessmentWorkflowNotFoundError
+)
+from openassessment.workflow.models import TeamAssessmentWorkflow
+from openassessment.workflow.serializers import TeamAssessmentWorkflowSerializer
+
+
+logger = logging.getLogger('openassessment.workflow.models')  # pylint: disable=invalid-name
```
iloveagent57, line 16: It's customary to do `logger = logging.getLogger(__name__)`, which helps avoid copypasta :)

openassessment/workflow/models.py
```diff
@@ -613,8 +613,37 @@ class TeamAssessmentWorkflow(AssessmentWorkflow):
...
+
+    @classmethod
+    def get_by_team_submission_uuid(cls, team_submission_uuid):
+        try:
+            return cls.objects.get(team_submission_uuid=team_submission_uuid)
+        except cls.DoesNotExist:
+            return None
+        except DatabaseError as exc:
+            message = (
+                u"Error finding workflow for team submission UUID {uuid} due to error: {exc}."
```
iloveagent57, line 637: Nit (which applies to many string literals below): you don't have to use that `u` prefix, because python 3 (although it's still valid syntax).

openassessment/workflow/team_api.py
```diff
@@ -0,0 +1,139 @@
...
+        statuses.remove('ai')
+
+    return [
+        {
+            "status": status,
+            "count": TeamAssessmentWorkflow.objects.filter(
+                status=status,
+                course_id=course_id,
+                item_id=item_id,
+            ).count()
```
iloveagent57, line 136: You should be able to do this with one query with a Django [annotation](https://docs.djangoproject.com/en/2.0/topics/db/aggregation/#topics-db-aggregation):
```python
from django.db.models import Count
queryset = TeamAssessmentWorkflow.objects.filter(
    status__in=statuses,
    course_id=course_id, 
    item_id=item_id,
).values('status').annotate(count=Count('status'))
return [row for row in queryset]
```

nsprenkle: This produces slightly different behavior since it won't have counts of zero. Suggestions?

Andytr1: In that case, using your original query would make sense. Is your issue analogous to [this](https://stackoverflow.com/questions/51858201/django-annotate-count-include-0-results)?

nsprenkle: @Andytr1, yes, exactly! We could modify this but, following the non-team example: we want to know the steps/statuses even if they have a count of 0
```
[
  {"staus": "staff", "count": 0}
  {"status": "waiting", "count": 1}
  ...
]
```

Andytr1: If I was to implement @iloveagent57's suggestion, one approach is to create a dictionary with all statuses and the count initialized to 0. Then get the results of the query and set the values in the existing dictionary to the actual count. Any statuses not found in db, would remain at 0. It would be a little more code, but only a single query against the db would be executed.

iloveagent57: Yeah, what Andrew suggested is what I would do, too, something like:
```python
queryset = TeamAssessmentWorkflow.objects.filter(
    status__in=statuses,
    course_id=course_id, 
    item_id=item_id,
).values('status').annotate(count=Count('status'))
counts_by_status = {status: 0 for status in statuses}
for row in queryset:
    counts_by_status[row['status']] = row['count']
return [
    {'status': status, 'count': count}
    for status, count
    in counts_by_status.items()
]
```

PR #1386, opened 2020-04-09
Add Team Workflow Deletions to Teams API

nsprenkle:
**Context:** Implement workflow deletion functions at team assignment API level.

**Changes:**
- Implement and test (API)
    - `cancel_workflow`
    - `get_assessment_workflow_cancellation`
    - `is_workflow_cancelled`
- Delete `TeamAssessmentWorkflowCancellation` model and serializer, inherit behavior from parent class
- Version bump to ~2.6.27~ ~2.6.28~ 2.6.29
- Add extra spacing in `setup.py` to remove pylint warning

JIRA: 1/2 of [EDUCATOR-4978](https://openedx.atlassian.net/browse/EDUCATOR-4978)
FYI: @edx/masters-devs-gta


openassessment/workflow/team_api.py
```diff
@@ -155,3 +163,62 @@ def get_status_counts(course_id, item_id):
         for status, count
         in counts_by_status.items()
     ]
+
+
+def cancel_workflow(team_submission_uuid, comments, cancelled_by_id):
+    """
+    Add an entry in AssessmentWorkflowCancellation table for a AssessmentWorkflow.
```
Andytr1, line 170: nit: Technically it's TeamAssessmentWorkflow that should be referenced, not AssessmentWorkflow.

openassessment/workflow/team_api.py
```diff
@@ -155,3 +163,62 @@ def get_status_counts(course_id, item_id):
...
+        submission_uuid = _get_workflow_model(team_submission_uuid).submission_uuid
+        TeamAssessmentWorkflow.cancel_workflow(
+            submission_uuid,
+            comments,
+            cancelled_by_id,
+            TeamAssessmentWorkflow.REQUIREMENTS
+        )
+    except Exception as exc:
+        err_msg = (
+            u"Could not cancel team assessment workflow with team_submission_uuid {uuid} due to error: {exc}"
```
Andytr1, line 188: nit: don't need 'u' prefix anymore as all python 3 strings are unicode.

setup.py
```diff
@@ -33,9 +33,10 @@ def load_requirements(*requirements_paths):
             )
     return list(requirements)
 
+
```
jansenk, line 36: nit: remove blank line

nsprenkle: That was actually intentional: Pylint expects an extra line before the `setup` function so this removes a warning message.

PR #1392, opened 2020-04-23
Adding python38 into travis and tox.

awais786:
https://openedx.atlassian.net/browse/BOM-1569

- Removing django prior 2.2
- Adding py3.8
- Added allowd failures for 3.8 ( build is failing on 3.8)
- Fixing pylint issue.
- Update the version in setup.py and will release after merge.

ora2 shows failures on py38 so i have created a followup [ticket](https://openedx.atlassian.net/browse/BOM-1582) for fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,19 +9,16 @@ sudo: false
 
 python:
   - 3.5
+  - 3.8
 
 env:
-  - TOX_ENV=django111
-  - TOX_ENV=django20
-  - TOX_ENV=django21
   - TOX_ENV=django22
+  - TOX_ENV=js
+  - TOX_ENV=quality
 
-matrix:
-  include:
-    - python: 3.5
-      env: TOX_ENV=js
-    - python: 3.5
-      env: TOX_ENV=quality
+jobs:
+  allow_failures:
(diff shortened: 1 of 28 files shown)


.travis.yml
```diff
@@ -23,6 +21,10 @@ matrix:
     - python: 3.5
       env: TOX_ENV=quality
 
+  allow_failures:
+  - python: 3.8
+    env: TOX_ENV=django22
```
morenol, line 26: I think that we can remove this line

awais786: > I think that we can remove this line

Please review now.

.travis.yml
```diff
@@ -9,19 +9,16 @@ sudo: false
...
+  - TOX_ENV=js
+  - TOX_ENV=quality
 
-matrix:
-  include:
-    - python: 3.5
-      env: TOX_ENV=js
-    - python: 3.5
-      env: TOX_ENV=quality
+jobs:
```
morenol, line 19: Just for curiosity, why it was changed to jobs?

awais786: Removed the matrix part and just using allow_failures only.

PR #1393, opened 2020-04-23
Team assessment API

nsprenkle:
## TL;DR
Add API for team assessments (staff grading of team assignments)

## Tell me more!
- API endpoints mirror logic in `openassessment.assessment.api.staff.py` but operate on a `team_submission_uuid` instead of `submission_uuid`
- Some functions (e.g. `_create_assessment`) are imported from Staff API instead of being duplicated.
- Tests deviate from the patterns in `test_staff.py` to be more succinct.
- Pinned some dependencies:
   - `freezegun` pinned at `0.3.14`, the pi release, to fix `datetime` issue
   - `libsass` updated to `0.10.1` since `0.10.0` not found
   - Constrained `Django`, `mock`, and `zipp` per @iloveagent57 and @Andytr1

## Refactoring?

Right now, the code is pretty copy/paste heavy which helps to explicitly lay out the desired behaviors, but it can be made more elegant by encapsulating shared logic in a base class. That work will arrive as a separate PR.

JIRA: [EDUCATOR-4987](https://openedx.atlassian.net/browse/EDUCATOR-4987)
FYI: @edx/masters-devs-gta


jansenk:
I know this still a draft, but make sure to make the matching change in `AssessmentWorkflowStep.api()` to expose this api

nsprenkle:
> I know this still a draft, but make sure to make the matching change in `AssessmentWorkflowStep.api()` to expose this api

What is the matching change?

openassessment/assessment/api/teams.py
```diff
@@ -0,0 +1,363 @@
...
+
+    """
+    assessment = get_latest_staff_assessment(team_submission_uuid)
+    if not assessment:
+        return None
+
+    return {
+        "points_earned": assessment["points_earned"],
+        "points_possible": assessment["points_possible"],
+        "contributing_assessments": [assessment["id"]],
```
Andytr1, line 149: for contributing_assessments - we appear to always return a single item. Is their a plan to return more?

nsprenkle: Good question! The data structure that the staff API produces has a `contributing_assessments` array with only one assessment, so I kept the structure the same here.

Why is it an array? Possibly to allow showing of assessments and override history in the future, but this behavior hasn't been implemented in the staff API.

mattcarter: likelihood is for model consistency with peer assessments where multiple assessments are expected.

openassessment/assessment/api/teams.py
```diff
@@ -0,0 +1,363 @@
...
+        team_submission_uuid,
+        scorer_id,
+        options_selected,
+        criterion_feedback,
+        overall_feedback,
+        rubric_dict,
+        scored_at=None
+):
+    """
+    Creates an assessment for each member of the submitting team.
```
Andytr1, line 315: Is it necessary to create an assessment for each team member? Wouldn't a single assessment be sufficient ?

openassessment/assessment/test/test_team.py
```diff
@@ -0,0 +1,327 @@
...
+
+        self.default_assessment = (
+            self.staff_user_id,  # scorer_id
+            OPTIONS_SELECTED_DICT["few"]["options"],  # options_selected
+            dict(),  # critereon_feedback
+            '',  # overall_feedback
+            RUBRIC  # rubric_dict
+        )
+
+    def test_submitter_is_finished(self):
```
Andytr1, line 49: For future, consider using `setUpTestData` method: https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.TestCase.setUpTestData . That way, any objects you create will be rolled back once this test class is finished.

nsprenkle: Ooh, good point. I think I'll move the user creation into there, but keep the assessments in the `setUp` function so that we have clean data for each run.

openassessment/assessment/api/teams.py
nsprenkle: Since assessments are tied to a submission (instead of a team submission) we need an assessment tied to each member's individual submission. This should be noted up in the design docs.

PR #1427, opened 2020-06-06
[BD-5] [TNL-7274] BB-2504: Add grade status summary page

ahmedbilal:
Add a message on the “Your grade” step explaining how the final grade is calculated/determined.

**JIRA tickets**: 
* https://openedx.atlassian.net/browse/TNL-7274
* https://openedx.atlassian.net/browse/OSPR-4663

**Dependencies**: None

**Screenshots**: 
**[Flow: Self Assessment Only]**
image

**[Flow: Peer Assessment Only]**
image

**[Flow: Staff Assessment Only]**
image

**[Flow: If staff assessment is one of the assessment steps and all the assessments are not complete yet]**
image

**[Flow: Peer with incomplete peer reviews]**
image

**[Flow: Self Assessment Only but when it is not done yet]**
image

**Merge deadline**: None

**Testing instructions**:

1. Create a sample course.
2. Create ORA (Open Response Assessment) one for each assessment type e.g **self**, **peer**, **staff**, **self** + **peer**, ....., etc.
3. Create student and enroll them into the created course.
4. Get them to attempt the ORA assessments and get them to grading page.
5. Check the output HTML and see whether the grade explanation message match the following criteria.
    ### If all the assessments are complete

    * If there is a **Staff Assessment** then it should say 
        >You have successfully completed this problem and received a {earned_points}/{total_points}. The grade for this problem is determined by your Staff Grade.
    
        no matter whether there is also a **Peer** or **Self Assessment** or not as Staff Assessments overrides every other type of assessment.
    * If there is only a single type of assessment be it **Staff, Self or Peer**, it should output the following message corresponding to each assessment type
        * **Staff**: You have successfully completed this problem and received a {earned_points}/{total_points}. The grade for this problem is determined by your Staff Grade.
        * **Peer**: You have successfully completed this problem and received a {earned_points}/{total_points}. The grade for this problem is determined by the median score of your Peer Assessments.
        * **Self**: You have successfully completed this problem and received a {earned_points}/{total_points}. The grade for this problem is determined by your Self Assessment.
    * If an assignment have both **Self** and Peer **Assessment** than the output message would depend upon the `Grade Assessment Priority`. By default it is as follows
        1. Staff Assessment
        2. Peer Assessment
        3. Self Assessment

        So, in case of default assessment priority the message would be 
        >You have successfully completed this problem and received a {earned_points}/{total_points}. The grade for this problem is determined by the median score of you Peer Assessments.
        
        but if the **Self Assessment** have higher priority than **Peer Assessment** than the message would be
        >You have successfully completed this problem and received a {earned_points}/{total_points}. The grade for this problem is determined by your Self Assessment.

    ### If some assessment is incomplete

    * No message should contain the sentence 
        >You have successfully completed this problem and received a {earned_points}/{total_points}.
    * The second sentence of messages would follow the above rules.
    * First sentence of all flows would be empty except for one special case and that is **Peer Assessment** only. It should have the following first sentence
        >You have not yet received all necessary peer reviews to determine your final grade.

**Reviewers**
- [ ] @giovannicimolin
- [ ] edX reviewer[s] TBD


edx-webhook:
Thanks for the pull request, @ahmedbilal! I've created [OSPR-4663](https://openedx.atlassian.net/browse/OSPR-4663) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

openassessment/workflow/models.py
```diff
@@ -87,10 +87,7 @@ class AssessmentWorkflow(TimeStampedModel, StatusModel):
     # This Django setting is a list of assessment steps (defined in `settings.ORA2_ASSESSMENTS`)
     # in descending priority order.
     DEFAULT_ASSESSMENT_SCORE_PRIORITY = ['peer', 'self']  # pylint: disable=invalid-name
-    ASSESSMENT_SCORE_PRIORITY = getattr(
```
giovannicimolin, line 90: @ahmedbilal Why were these changes introduced? Is this an ORA bug?

openassessment/workflow/models.py
```diff
@@ -111,6 +108,12 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super(AssessmentWorkflow, self).__init__(*args, **kwargs)
+
```
giovannicimolin, line 111: Same as the comment above.

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +585,80 @@ def _assessment_grade_context(self, assessment):
...
+        based on assessment_steps.
+        Args:
+            assessment_steps (list): List of assessment_steps for some ORA.
+            score (dict): Dict containing points_possible, points_earned and its related attributes
+        Returns:
+            str
+        """
+
+        first_sentence = _(
+            "You have successfully completed this problem and received a {earned_points}/{total_points}."
```
giovannicimolin, line 601: The sentence is not always this one. If the user didn't finish all the steps, we should display this status box too.

giovannicimolin: The message box needs to be added alongside this: 
https://github.com/edx/edx-ora2/blob/685846f34743685636d9d03b94919c54ff2b62da/openassessment/templates/openassessmentblock/grade/oa_grade_incomplete.html#L27

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +585,80 @@ def _assessment_grade_context(self, assessment):
...
+        ).format(earned_points=score["points_earned"], total_points=score["points_possible"])
+
+        second_sentences_options = {
+            "self": _("The grade for this problem is determined by your Self Assessment."),
```
giovannicimolin, line 605: You'll need to update the translations using the available make files.
The error can be reproduced locally running:
```
make check_translations_up_to_date
```
And to fix the issue:
```
make extract
```

openassessment/workflow/models.py
ahmedbilal: The Assessment Score Priority was initialized at the class level by obtaining value of `ORA2_ASSESSMENT_SCORE_PRIORITY` variable from settings module . This is the right approach if we know that the Assessment Score Priority wouldn't be updated/changed during runtime but during tests we need to check the code with settings patched/updated at the runtime so that would only be possible if the Assessment Score Priority is initialized inside the constructor i.e initialized at the runtime.

openassessment/workflow/models.py
ahmedbilal: Please see the below comment.

openassessment/xblock/grade_mixin.py
ahmedbilal: @giovannicimolin I think the document is confusing `incomplete` as `waiting`. 

`incomplete` mean that student haven't complete some assessment that he/she must to get to the next step e.g `peer assessment` or `self assessment`.

`waiting` mean that student have complete all assessment on his/her part but some assessment is left either on **peer** side or **staff** side.

I think that the message box should be added to the `waiting` html template instead of `incomplete` html template.

However, that is my understanding. I will add the message box to `incomplete` and if you think I am right, i will put it to `waiting` instead.

Thanks

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +587,94 @@ def _assessment_grade_context(self, assessment):
...
+        Returns:
+            str
+        """
+
+        is_self_assessment = "self-assessment" in assessment_steps
+        is_peer_assessment = "peer-assessment" in assessment_steps
+        is_staff_assessment = "staff-assessment" in assessment_steps
+
+        # Assessment Score priorities in descending order
+        assessment_score_priority = AssessmentWorkflow().ASSESSMENT_SCORE_PRIORITY[::-1]
```
giovannicimolin, line 608: We can't directly import the `AssessmentWorkflow` from models to inside the XBlock since Django components are not parts of the XBlock API.
I'm not sure how to approach this currently, but the current solution is not acceptable.
The correct way here would be to modify the `AssessmentWorkflowSerializer` and add a property to the model reflecting where the grade came from, but that's out of the question given the task's budget.

giovannicimolin: For the current implementation, consider this priority:
1. Staff grade
2. Peer grade
3. Self grade

This will simplify a lot your logic down below.
Make sure to add a note in the PR description raising this concern (we're not using the `ASSESSMENT_SCORE_PRIORITY` because there's no API for it yet).
I'll create a follow up task to deal with this.

openassessment/xblock/test/test_grade_explanation.py
```diff
@@ -0,0 +1,263 @@
+# -*- coding: utf-8 -*-
+"""
+Tests for grade explanation in Open Response Assessment XBlock.
+"""
+from __future__ import absolute_import
+
+import json
+
+from openassessment.workflow.models import AssessmentWorkflow
```
giovannicimolin, line 9: Don't import models when running tests on the XBlock side.

openassessment/workflow/models.py
giovannicimolin: @ahmedbilal Can you revert these changes? 
I took a deeper look at your code and noticed that you were trying to import a Django model inside the XBlock, and it can't be done that way.

Django methods/models/etc can't be accessed inside an XBlock, and that's why we have multiple APIs that are in turn used by the XBlock.

For this ticket, just consider the platform's default priority when grading, but add a note to the PR explicitly stating it.

Can you revert these changes to the `models.py` given what I've said above?

openassessment/xblock/grade_mixin.py
giovannicimolin: If we were to add this information from the Django side to the XBlock side, I would probably modify `self.get_workflow_info()` (and this would go all the way down to the serializers) to get the actual grading priority for this.

ahmedbilal:
@giovannicimolin 
* Version bumped in both files
* **edx:edx-ora2:master** merged into **open-craft:edx-ora2:ahmed/bb-2504-add-grade-explanation-message**

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +590,101 @@ def _assessment_grade_context(self, assessment):
...
+        # making decisions upon it
+        _assessment_type = None
+        score = workflow['score']
+        complete = score is not None
+
+        is_self_assessment = "self-assessment" in self.assessment_steps
+        is_peer_assessment = "peer-assessment" in self.assessment_steps
+        is_staff_assessment = "staff-assessment" in self.assessment_steps
+
+        if is_staff_assessment:
```
giovannicimolin, line 605: Does staff assessment always has the highest priority?

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +590,101 @@ def _assessment_grade_context(self, assessment):
...
+            _assessment_type = "staff_defined"
```
giovannicimolin, line 606:
```suggestion
            return "staff_defined"
```

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +590,101 @@ def _assessment_grade_context(self, assessment):
...
+        else:
```
giovannicimolin, line 607: Simplify the code below by returning early and avoiding more if-else statements.

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +590,101 @@ def _assessment_grade_context(self, assessment):
...
+                    if not is_self_assessment:
+                        raise ValueError()
+                    self_index = workflow["assessment_score_priority"].index("self")
+                except ValueError:
+                    self_index = float("inf")
+
+                if min(self_index, peer_index) == peer_index:
+                    _assessment_type = "peer_defined"
+                else:
+                    _assessment_type = "self_defined"
```
giovannicimolin, line 640: Taking into account my other suggestions, and changing from `peer_defined` to just `peer` or `self` (without the `_defined`), the code can be boiled down to:
```python
for score_priority in workflow["assessment_score_priority"]:
    # Skip staff
    if score_priority == "staff":
        continue

    # Return first assessment type (the one with the higest priority is the first).                 
    return score_priority
```

The entire function can be globally optimized to get a smaller footprint like this:
```python
    def _get_assessment_type(self, workflow):
        """
        Some docstring here :)
        """
        score = workflow['score']
        complete = score is not None

        if "staff-assessment" in self.assessment_steps:
            return "staff"

        # Edge case: staff overrides the grade.
        # If a score is overriden by staff, it'll always have an
        # attached annotation type with the `staff_defined` value,
        # so we look for that in this problem's annotation and
        # return staff if it's found.
        grade_annotation_types = [annotation['annotation_type'] for annotation in score["annotations"]]
        if complete and "staff_defined" in grade_annotation_types:
            return "staff"

        # For other cases, we just need to figure out the
        # priority of each (either peer or self).
        # Just loop over the values and return the first one
        # after staff.
        for _assessment_type in workflow["assessment_score_priority"]:
            # assessment_step would always have staff in it, so skip it 
            # while checking the priority here.
            if _assessment_type == "staff":
                continue

            return _assessment_type
```
It's simpler than the current approach and is more human readable.
Take care to adjust `_get_score_explanation` to remove the `_defined` suffix.

openassessment/xblock/grade_mixin.py
ahmedbilal: Yes, The first thing we check is that if there is a staff assessment then the grade would be decided using staff assessment.

openassessment/workflow/models.py
```diff
@@ -110,6 +107,10 @@ class Meta:
         app_label = "workflow"
 
     def __init__(self, *args, **kwargs):
+        AssessmentWorkflow.ASSESSMENT_SCORE_PRIORITY = getattr(
```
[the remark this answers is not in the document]
ahmedbilal: Unfortunately, this is necessary as our tests need to check for different score assessment priorities and if we don't do the above thing the score assessment priorities would remain fixed as they are defined at the class level which would not satisfy what we need to test.

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +590,101 @@ def _assessment_grade_context(self, assessment):
...
+        is_peer_only_assessment = is_peer_assessment and not (is_self_assessment or is_staff_assessment)
+
+        sentences = {
+            "staff": _("The grade for this problem is determined by your Staff Grade."),
+            "peer": _(
+                "The grade for this problem is determined by the median score of "
+                "your Peer Assessments."
+            ),
+            "self": _("The grade for this problem is determined by your Self Assessment.")
+        }
```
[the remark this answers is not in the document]
ahmedbilal: If you see current version, you would understand why this should remain there. I think this would also be beneficial in future as the grade explanation function become more complex and then it would be very convenient to refer to this dict.

openassessment/xblock/grade_mixin.py
```diff
@@ -584,3 +590,79 @@ def _assessment_grade_context(self, assessment):
...
+            ).format(earned_points=score["points_earned"], total_points=score["points_possible"])
+        else:
+            first_sentence = ""
+            # Special Case i.e If the submission only have peer assessment
+            if "peer-assessment" in self.assessment_steps and "self-assessment" not in self.assessment_steps and \
+               "staff-assessment" not in self.assessment_steps:
+                first_sentence = _(
+                    "You have not yet received all necessary peer reviews to determine your final grade."
+                )
+                second_sentence = ""
```
giovannicimolin, line 666: You are already getting the assessment type above (where the grade is coming from).

With that you can just replace this chunk with:
```suggestion
            # Special Case i.e If the submission only have peer assessment
            if assessment_type == "peer":
                first_sentence = _(
                    "You have not yet received all necessary peer reviews to determine your final grade."
                )
```

giovannicimolin: @ahmedbilal Also, the second sentence should not be empty:
Peer | Peer Step Incomplete | You have not yet received all necessary peer reviews to determine your final grade. | The grade for this problem is determined by the median score of your Peer Assessments.
-- | -- | -- | --

(from the requirements list)

openassessment/xblock/grade_mixin.py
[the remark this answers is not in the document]
giovannicimolin: Makes sense, lets keep it as is then

openassessment/workflow/models.py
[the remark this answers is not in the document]
giovannicimolin: @ahmedbilal I still think this isn't necessary, did you try using the `override_settings` decorator from `from django.test.utils import override_settings`? 

It serves the same purpose here, and we don't need to change stable code.

openassessment/xblock/test/test_grade_explanation.py
```diff
@@ -0,0 +1,263 @@
...
+            xblock, self.SUBMISSION, [], [], SELF_ASSESSMENT,
+            waiting_for_peer=True
+        )
+        resp = self.request(xblock, 'render_grade', json.dumps(dict()))
+
+        self.assertIn(self.second_sentences_options["self"], resp.decode('utf-8'))
+
+    @scenario('data/grade_scenario_staff_only.xml', user_id='Bernard')
+    @data(*assessment_score_priority)
+    def test_render_explanation_grade_staff_only(self, xblock, assessment_score_priority):
```
giovannicimolin, line 47: Why test with 2 different priorities here if the final outcome doesn't matter for staff graded problems? 
Can you remove `@data(*assessment_score_priority)` and `assessment_score_priority` for staff related tests?

openassessment/xblock/test/test_grade_explanation.py
```diff
@@ -0,0 +1,263 @@
...
+        resp = self.request(xblock, 'render_grade', json.dumps(dict()))
+
+        if assessment_score_priority.index('self') < assessment_score_priority.index('peer'):
+            self.assertIn(self.second_sentences_options["self"], resp.decode('utf-8'))
+        else:
+            self.assertIn(self.second_sentences_options["peer"], resp.decode('utf-8'))
+
+    @scenario('data/grade_scenario_self_staff.xml', user_id='Bernard')
+    @data(*assessment_score_priority)
+    def test_render_grade_explanation_self_and_staff(self, xblock, assessment_score_priority):
```
giovannicimolin, line 84: Same as the previous comment here and in all other tests that include a `staff` grade, change the assessment priority if the response is always the same? `staff` will always prevail, no matter which other priorities exist, so these extra tests are not really usefull.

openassessment/workflow/models.py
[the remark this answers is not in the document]
giovannicimolin: @ahmedbilal I've spent some time investigating and testing things out in my devstack.
Without using this change (keeping this file as it was) I've managed to replace the value with `unittest.mock` using the code below:
```python
 # At the top of the file:
from mock import patch

# Then in the tests:
    @scenario('data/grade_scenario.xml', user_id='Bernard')
    @data(*assessment_score_priority)
    def test_render_grade_explanation_self_and_peer(self, xblock, assessment_score_priority):
        # See how I've structured the patch here:
        with patch(
            'openassessment.workflow.models.AssessmentWorkflow.ASSESSMENT_SCORE_PRIORITY',
            assessment_score_priority
        ):
            self.create_submission_and_assessments(
                xblock, self.SUBMISSION, self.PEERS, PEER_ASSESSMENTS, SELF_ASSESSMENT,
            )
            resp = self.request(xblock, 'render_grade', json.dumps(dict()))

            if assessment_score_priority.index('self') < assessment_score_priority.index('peer'):
                self.assertIn(self.second_sentences_options["self"], resp.decode('utf-8'))
            else:
                self.assertIn(self.second_sentences_options["peer"], resp.decode('utf-8'))
```

Can you change the implementation to revert these changes and use the `patch` method?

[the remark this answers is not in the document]
ahmedbilal: Done. Thanks

openassessment/xblock/grade_mixin.py
ahmedbilal: **re:** first comment: Yes, it could not work if there is a learning assessment step (use to train student how to correctly assess peer assessments)

Yes, thanks for pointing out. Actually, the upstream ticket is updated many times and then I forget to follow it.

openassessment/xblock/test/test_grade_explanation.py
ahmedbilal: Yes, that what we wanted to test that **"No matter what the priority is the outcome would be based on staff grade"**
If you still want me to remove it, I would.

openassessment/xblock/test/test_grade_explanation.py
ahmedbilal: See me previous comment.

ahmedbilal:
@davestgermain Ping. This is ready for edX review.

openassessment/workflow/serializers.py
```diff
@@ -14,6 +14,10 @@ class AssessmentWorkflowSerializer(serializers.ModelSerializer):
     Serialize a AssessmentWorkflow' model.
     """
     score = serializers.ReadOnlyField(required=False)
+    assessment_score_priority = serializers.SerializerMethodField()
```
davestgermain, line 17: Another way to do this is:
`assessment_score_priority = serializers.ReadOnlyField(source='ASSESSMENT_SCORE_PRIORITY')`

giovannicimolin:
@davestgermain Can you tag the `2.8.7` release? We need to create a version bump PR for the main platform.

PR #1432, opened 2020-06-11
change clear_team_state to not use get_team_for_anonymous_user

jansenk:
**TL;DR -** fix my implemention of `clear_team_state` to not use `get_team_for_anonymous_user`

**What changed?**

`get_team_for_anonymous_user` relies on the xmodule runtime, and we don't have access to that when `clear_user_state` is called from the LMS.

I now do the same individual user submission lookup as in `clear_user_state` and pass the submission in, and then just grab the `team_submission_uuid` off of the submission.

*Developer Checklist**
- [x] Translations up to date

JIRA: [EDUCATOR-5031](https://openedx.atlassian.net/browse/EDUCATOR-5031)

FIY: @edx/masters-devs-gta

### DIFF
--- a/openassessment/xblock/staff_area_mixin.py
+++ b/openassessment/xblock/staff_area_mixin.py
@@ -13,7 +13,7 @@ from openassessment.assessment.errors import PeerAssessmentInternalError
 from openassessment.workflow.errors import AssessmentWorkflowError, AssessmentWorkflowInternalError
 from openassessment.xblock.data_conversion import create_submission_dict, list_to_conversational_format
 from openassessment.xblock.resolve_dates import DISTANT_FUTURE, DISTANT_PAST
-from submissions.errors import SubmissionNotFoundError, TeamSubmissionNotFoundError
+from submissions.errors import SubmissionNotFoundError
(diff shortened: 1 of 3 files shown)


openassessment/xblock/staff_area_mixin.py
```diff
@@ -467,60 +462,57 @@ def clear_student_state(self, user_id, course_id, item_id, requesting_user_id):
...
-            logger.warning(warning_msg)
-            return
+        student_item_string = "course {} item {} user {}".format(course_id, item_id, user_id)
 
-        if user_team is None:
-            warning_msg = error_msg_base + 'but they are not on a team for course {} item {}.'.format(
-                course_id, item_id
-            )
-            logger.warning(warning_msg)
+        if not submissions:
```
Andytr1, line 492: Trivial nit: we are mixing string.format and %s for string formatting.

jansenk: Yeah, but at the top i'm building a string and then using that built string as a %s format parameter itself, idk i just wanted to try to save myself some lines

PR #1433, opened 2020-06-12
Change xblock-sdk installation

mzulqarnain1:
https://openedx.atlassian.net/browse/BOM-1699


morenol:
xblock-sdk is currently not installed in edx-platform, and this would change that, Is that Ok?

mzulqarnain1:
> xblock-sdk currently is not installed in edx-platform, and this would change that, Is that Ok?

@morenol there shouldn't be any problem with that imo.

PR #1437, opened 2020-06-16
Move xblock-sdk from base.in to test.in

mzulqarnain1:
See https://github.com/edx/edx-platform/pull/24227
`xblock-sdk` being in `base.in` is installed with ora2 which causes server error on platform. It should only be installed in testing environment.


mzulqarnain1:
@morenol Please review.

PR #1453, opened 2020-07-20
[BD-05] [TNL-7308] Add method to collect attachments data to OraAggregateData

0x29a:
This PR adds `collect_ora2_attachments` method, which collects all submission attachments and represents them in a way useful to create attachments zip archive.

**JIRA tickets**:
- [TNL-7308](https://openedx.atlassian.net/browse/TNL-7308)
- [BLENDED-473](https://openedx.atlassian.net/browse/BLENDED-473)

**Dependencies**:
This PR is a dependency for https://github.com/edx/edx-platform/pull/24541.

**Merge deadline**: None

**Testing instructions**:

Since there are just one new method and it's covered with test, I think CI pass is enough to consider this tested. If not, this method can be tested in dependent pull request.

**Reviewers**
- [ ] @xitij2000 
- [ ] edX reviewer[s] TBD

### DIFF
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -2,10 +2,15 @@
 Aggregate data for openassessment.
 """
 
-
 from collections import defaultdict
+from io import StringIO
+from urllib.parse import urljoin
+from zipfile import ZipFile
 import csv
 import json
+import os
+
+import requests
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -13,6 +18,7 @@ from django.db.models import F
 
 from submissions import api as sub_api
 from openassessment.assessment.models import Assessment, AssessmentFeedback, AssessmentPart
+from openassessment.fileupload.api import get_download_url
 from openassessment.workflow.models import AssessmentWorkflow, TeamAssessmentWorkflow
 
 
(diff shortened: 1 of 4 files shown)


edx-webhook:
Thanks for the pull request, @0x29a! I've created [BLENDED-473](https://openedx.atlassian.net/browse/BLENDED-473) to keep track of it in Jira.  More details are on the [BD-05](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1529970851/BD-05+ORA+v3+Usability+Moderation) project page.

When this pull request is ready, tag your edX technical lead.

openassessment/data.py
```diff
@@ -670,3 +672,52 @@ def collect_ora2_responses(cls, course_id, desired_statuses=None):
...
+                    texts: [...text dicts...]
+                },
+                ...
+            },
+            ...
+        }
+        ```
+        """
+
+        all_submission_information = list(sub_api.get_all_course_submission_information(course_id, 'openassessment'))
```
xitij2000, line 695: Why convert this to a list when it's going to be iterated over normally?

```suggestion
        all_submission_information = sub_api.get_all_course_submission_information(course_id, 'openassessment')
```

openassessment/data.py
```diff
@@ -670,3 +672,52 @@ def collect_ora2_responses(cls, course_id, desired_statuses=None):
...
+
+        result = {}
+
+        for student_info, submission, _ in all_submission_information:
+            files = result.setdefault(
+                student_info['item_id'], {}
+            ).setdefault(
+                student_info['student_id'], {'attachments': [], 'texts': []}
+            )
```
xitij2000, line 704: This is a bit hard to follow, can you break it down into multiple lines?

openassessment/data.py
0x29a: Ah, that's a copy-paste from another function. No point in turning this into list here.

openassessment/data.py
```diff
@@ -670,3 +676,178 @@ def collect_ora2_responses(cls, course_id, desired_statuses=None):
...
+                    file_data['type'],
+                    file_data['description'],
+                    file_data['size'],
+                    file_data['file_path'],
+                )
+            )
+
+        output_buffer = StringIO()
+        csvwriter = csv.writer(output_buffer)
+        csvwriter.writerows([header] + rows)
```
xitij2000, line 739: Note: 

If you use DictWriter you can simplify this code a lot. This whole bit a lot. All this can become:
```suggestion
        output_buffer = StringIO()
        csvwriter = csv.DictWriter(output_buffer, raw_fields)
        csvwriter.writerow(header)
        csvwriter.writerows(file_data)
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "edx-ora2",
-  "version": "2.8.10",
+  "version": "2.8.11",
```
xitij2000, line 3: I think this justifies a bump to 2.9.0 since it add a new functionality.

0x29a:
@davestgermain, this one is ready for review, as well as https://github.com/edx/edx-platform/pull/24541.

openassessment/data.py
davestgermain: I agree that [DictWriter](https://docs.python.org/3/library/csv.html#csv.DictWriter) would be cleaner. Since you're defining the column names in this class (`collect_ora2_submission_files`), you don't have to repeat them here.

openassessment/data.py
```diff
@@ -670,3 +676,178 @@ def collect_ora2_responses(cls, course_id, desired_statuses=None):
...
+                    'size': len(content),
+                    'file_path': os.path.join(
+                        str(course_id),
+                        student['item_id'],
+                        student['student_id'],
+                        file_name,
+                    )
+                })
+
+        return files
```
davestgermain, line 853: An optimization would be to make this method an iterator instead of returning a list of files. Just `yield` each file dict.

openassessment/data.py
```diff
@@ -670,3 +676,178 @@ def collect_ora2_responses(cls, course_id, desired_statuses=None):
                 result[item_id][status] += 1
 
         return result
+
+
+class OraDownloadData:
```
davestgermain, line 681: I don't really like that this is a class that only has classmethods. That seems to indicate that these are better off as simple functions in `data.py`. But I see that you're copying the pattern that was here already, in `OraAggregateData`. So I suppose it's ok for now.

openassessment/data.py
```diff
@@ -670,3 +676,178 @@ def collect_ora2_responses(cls, course_id, desired_statuses=None):
...
+            for file_data in submission_files_data:
+                file_content = (
+                    cls._download_file_by_key(file_data['key'])
+                    if file_data['type'] == cls.ATTACHMENT
+                    else file_data['content']
+                )
+
+                zip_file.writestr(file_data['file_path'], file_content)
+
+            downloads_csv = cls._create_downloads_csv(submission_files_data)
```
davestgermain, line 781: If the download csv is only ever used in this zip file, why not just create the csv as you iterate `submission_files_data`?

0x29a:
Thanks for suggesions, @davestgermain and @xitij2000.

`collect_ora2_submission_files` now yields results instead of constructing list. Also, I managed to get rid of separate `_create_downloads_csv`, now csv is generated along with writing files in zip.

openassessment/data.py
0x29a: >as simple functions in data.py

I agree. Probably, since there are already two such classes, we can split `data.py` on three different files, like `aggregate.py`, `download.py` and `utils.py` (for `CsvWriter`).

0x29a:
>Could you just squash the commits?

Done, @davestgermain.

PR #1458, opened 2020-07-27
Clear files when resetting student state

nsprenkle:
**TL;DR -** if we clear a student's/team's state and they have submitted, also remove the team's files. This creates a clean slate for the learner(s) to submit a new response.

**What changed?**
- Add new API endpoint for deleting team files
- Clearing state for a team ORA deletes shared file uploads for the team submission
- Clearing state for an individual ORA deletes file uploads associated with any submissions

**Note:** this does not extend to clearing files when state is cleared before a submission is created. There really isn't a use case when staff would want to do this and should probably be formally discouraged.

**Developer Checklist**

JIRA: [EDUCATOR-5108](https://openedx.atlassian.net/browse/EDUCATOR-5108)

FIY: @edx/masters-devs-gta


nsprenkle:
Open question: do we need access controls limiting this to staff?

nsprenkle:
Blocked pending arch review

openassessment/xblock/staff_area_mixin.py
```diff
@@ -510,10 +511,15 @@ def clear_team_state(self, user_id, course_id, item_id, requesting_user_id, subm
...
             requesting_user_id
         )
+
         # Tell the submissions API to orphan the submissions to prevent them from being accessed
         from submissions import team_api as team_submissions_api
         team_submissions_api.reset_scores(team_submission_uuid)
 
+        # Clean up shared files for the team
+        team_id = team_submissions_api.get_team_submission(team_submission_uuid).get('team_id', None)
+        delete_shared_files_for_team(course_id, item_id, team_id)
```
Andytr1, line 521: How does `delete_shared_files_for_team` behave when the `team_id` is `None` ? Does it make sense to add a test case for this?

nsprenkle: It can't have a `team_id` of None, since a `TeamSubmission` must have a `team_id`.

PR #1461, opened 2020-08-04
Fix bug where resetting state for a team does not allow resubmission

nsprenkle:
**TL;DR -** clear submissions from context when resetting state for teams.

**Notes:**
This commit adds just the correct clear submissions flag and won't be fully demonstrable until integrated with the [updated team cancellation workflow work](https://github.com/muselesscreator/edx-ora2/tree/teams_cancellation_workflow) and [reset student state work](https://github.com/edx/edx-platform/pull/24187/).

**Developer Checklist**

JIRA: [EDUCATOR-5183](https://openedx.atlassian.net/browse/EDUCATOR-5183)

FIY: @edx/masters-devs-gta

### DIFF
--- a/openassessment/xblock/staff_area_mixin.py
+++ b/openassessment/xblock/staff_area_mixin.py
@@ -512,7 +512,7 @@ class StaffAreaMixin:
         )
         # Tell the submissions API to orphan the submissions to prevent them from being accessed
         from submissions import team_api as team_submissions_api
-        team_submissions_api.reset_scores(team_submission_uuid)
+        team_submissions_api.reset_scores(team_submission_uuid, clear_state=True)
 
     @XBlock.json_handler
     @require_course_staff("STUDENT_INFO", with_json_handler=True)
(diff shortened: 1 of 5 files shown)


openassessment/xblock/staff_area_mixin.py
```diff
@@ -512,7 +512,7 @@ def clear_team_state(self, user_id, course_id, item_id, requesting_user_id, subm
         )
         # Tell the submissions API to orphan the submissions to prevent them from being accessed
         from submissions import team_api as team_submissions_api
-        team_submissions_api.reset_scores(team_submission_uuid)
+        team_submissions_api.reset_scores(team_submission_uuid, clear_state=True)
```
muselesscreator, line 515: Should there be a test updated to reflect this?

nsprenkle: The only non-integration way to test it that I can think of just asserts the correct arguments, but I don't know how useful it would be.

Andytr1: I would say the bulk of the testing will happen once the integration is complete with the other two prs linked above.

nsprenkle: I reverse that, @muselesscreator, I did find a way to test it. Good call 👍

PR #1467, opened 2020-08-17
Surface file info as a dict instead of a tuple

nsprenkle:
**TL;DR -** surface file info as a dict instead of a tuple.

**Benefit:** Treating file info as a dictionary allows us to remove or extend file data more easily when surfacing to Django templates.

**What changed?**
- Change file API to pass file info as a dict instead of a tuple: `file_descriptor_tuples` -> `files_metadata`
- Remove deprecated `FileDescriptor` and `TeamFileDescriptor` named tuples
- Add a Django template filter, `get_item`, to allow unpacking data from a dictionary
- Modify templates to unpack and use dictionary data for file info

**Developer Checklist**
- [x] Translations up to date
- [ ] Test suites passing on Jenkins
- [ ] Bumped version number in [setup.py](../setup.py) and [package.json](../package.json)

JIRA: [JIRA-XXXX](https://openedx.atlassian.net/browse/JIRA-XXXX)

FIY: @edx/masters-devs-gta

### DIFF
--- a/openassessment/templates/openassessmentblock/oa_team_uploaded_files.html
+++ b/openassessment/templates/openassessmentblock/oa_team_uploaded_files.html
@@ -1,5 +1,6 @@
 {% spaceless %}
 {% load i18n %}
+{% load oa_extras%}
 
 {% if file_upload_type %}
   <div class="{{ class_prefix }}__display__file {% if not team_file_urls %}is--hidden{% endif %} submission__{{ file_upload_type }}__upload" data-upload-type="{{ file_upload_type }}">
@@ -10,7 +11,9 @@
         {% endif %}
 
         <div class="submission__answer__files">
-        {% for file_url, file_description, file_name, owner_username in team_file_urls %}
+        {% for file in team_file_urls %}
+        {% with file_url=file|get_item:"download_url" file_description=file|get_item:"description" file_name=file|get_item:"name" owner_username=file|get_item:"uploaded_by" %}
+
             <div class="submission__answer__team__file__block submission__answer__team__file__block__{{ forloop.counter0 }}" {% if not file_url %} deleted {% endif %}>
             {% if file_url %}
                 {% if file_upload_type == "image" %}
@@ -31,6 +34,7 @@
                 [{% trans "Uploaded by" %} <strong class="emphasis">{{ owner_username }}</strong>]
             {% endif %}
             </div>
+        {% endwith %}
         {% endfor %}
         </div>
     </div>
--- a/openassessment/fileupload/api.py
+++ b/openassessment/fileupload/api.py
@@ -211,10 +211,6 @@ class FileUpload:
         return self._to_dict() == other._to_dict()  # pylint: disable=protected-access
 
 
-FileDescriptor = namedtuple('FileDescriptor', ['download_url', 'description', 'name', 'show_delete_button'])
-TeamFileDescriptor = namedtuple('TeamFileDescriptor', ['download_url', 'description', 'name', 'uploaded_by'])
-
-
 class FileUploadManager:
     """
     Manages the CRUD operations of file uploads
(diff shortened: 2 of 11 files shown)


jansenk:
I don't think we particularly want to avoid using the namedtuple in code, just in the template. I think you can call _asdict() on a namedtuple and it returns a dict representation.

openassessment/templates/openassessmentblock/oa_team_uploaded_files.html
```diff
@@ -10,7 +11,9 @@ <h5 class="submission__answer__part__text__title">
         {% endif %}
 
         <div class="submission__answer__files">
-        {% for file_url, file_description, file_name, owner_username in team_file_urls %}
+        {% for file in team_file_urls %}
+        {% with file_url=file|get_item:"download_url" file_description=file|get_item:"description" file_name=file|get_item:"name" owner_username=file|get_item:"uploaded_by" %}
```
muselesscreator, line 15: This is hard to read... is there any way to break it up or put this query into js?

jansenk, line 15: I don't think we need this?
Does `file.download_url`, `file. description` etc not work?

jansenk: actually .... i think that same exact syntax would work for the namedtuples as well

nsprenkle: Not that I'm aware of... I can break it up more so its not all on one line and just use the filters where the variable is used, but this makes the change-set easier to grok. It'd basically be the same thing but spread out.

nsprenkle: I stand corrected. @jansenk's syntax works and removes the need for a filter.

nsprenkle: Yup, this works. Will update.

nsprenkle:
> I don't think we particularly want to avoid using the namedtuple in code, just in the template. I think you can call _asdict() on a namedtuple and it returns a dict representation.

Is there a reson to prefer a `namedtuple` over a `dict` or is that just to reduce the change-set throughout the code?

openassessment/templates/openassessmentblock/oa_team_uploaded_files.html
```diff
@@ -10,25 +10,25 @@ <h5 class="submission__answer__part__text__title">
...
 
         <div class="submission__answer__files">
-        {% for file_url, file_description, file_name, owner_username in team_file_urls %}
-            <div class="submission__answer__team__file__block submission__answer__team__file__block__{{ forloop.counter0 }}" {% if not file_url %} deleted {% endif %}>
-            {% if file_url %}
+        {% for file in team_file_urls %}
+            <div class="submission__answer__team__file__block submission__answer__team__file__block__{{ forloop.counter0 }}" {% if not file.download_url %} deleted {% endif %}>
+            {% if file.download_url %}
                 {% if file_upload_type == "image" %}
                     {% if file_description %}
```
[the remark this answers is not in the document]
nsprenkle: Oof, I'm sort of concerned that the test suite didn't catch this...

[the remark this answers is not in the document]
jansenk: do we have a team render_submission test? Might b worthwhile to toss one in there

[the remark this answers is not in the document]
muselesscreator: I am adding one for my stuff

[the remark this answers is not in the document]
nsprenkle: @muselesscreator, worth me making one or would it be duplication of work?

PR #1469, opened 2020-08-19
EDUCATOR-5211 - display warning with existing submission

Andytr1:
If a learner was reassigned from a team that had a submission to a new team, display a warning that the learner's original submission with the first team will be used for grading. 

**What changed?**
Display warning that a learner has a submission from a team other than their current team.

**Developer Checklist**
- [x] Translations up to date

JIRA: [JIRA-5211](https://openedx.atlassian.net/browse/EDUCATOR-5211)

FIY: @edx/masters-devs-gta


openassessment/xblock/team_mixin.py
```diff
@@ -109,11 +112,23 @@ def get_team_info(self):
         if self.in_studio_preview:
             return self.STAFF_OR_PREVIEW_INFO
         elif self.has_team():
+            student_item_dict = self.get_student_item_dict()
```
jansenk, line 115: This should be moved to submission_mixin, to maybe be refactored into a get_team_context method.

Andytr1: I would like to hear more on your thoughts as to why. Do you mean the entire new block should be in the submission_mixin?

PR #1470, opened 2020-08-19
Regisb/boto2to3

regisb:
**What changed?**

Boto is the old library that was used to interact with the AWS APIs, and
in particular the S3 API. It was replaced by boto3. We need to upgrade
boto to boto3 because boto makes it near impossible to customize the
connection settings.

It goes without saying that this should be tested with edX.org's own S3 settings to check whether the new configuration is compatible. In particular, it is possible that the default calling format has changed from "subdomain" to "ordinary".

**Developer Checklist**
- [x] Reviewed the [release process](./release_process.md)
- [x] Translations up to date
- [ ] ~~JS minified, SASS compiled~~
- [x] Test suites passing on Jenkins
- [x] Bumped version number in [setup.py](../setup.py) and [package.json](../package.json)

This is a follow up of a PR that has become stale: https://github.com/edx/edx-ora2/pull/1414 @ormsbee Do you think you could review this please? This is ready for review.

PR #1504, opened 2020-09-21
Jkantor/regisb/boto2to3

jansenk:
Merge conflict resolution for https://github.com/edx/edx-ora2/pull/1470

I'm going to spool up a sandbox and run some tests 

Text of original PR: 

```
What changed?

Boto is the old library that was used to interact with the AWS APIs, and
in particular the S3 API. It was replaced by boto3. We need to upgrade
boto to boto3 because boto makes it near impossible to customize the
connection settings.

It goes without saying that this should be tested with edX.org's own S3 settings to check whether the new configuration is compatible. In particular, it is possible that the default calling format has changed from "subdomain" to "ordinary".

```

@edx/masters-devs-gta


jnlapierre (approved):

nsprenkle (approved):
Haven't run, but visually LGTM.

jansenk:
I'm gonna wait to merge this on Monday

PR #1509, opened 2020-09-25
Explicitly Set File Upload Extensions

nsprenkle:
**TL;DR -** to enable better file upload experiences, we need to better establish and communicate what files learners are allowed to upload for responses. We don't have a list of allowed files types for our existing presets (image, and pdf and image).

**What changed?**
This PR establishes an "allowed extension" list (sourced from the [MDN MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) page) for the 2 presets and uses it to apply default `white_listed_file_types` values where otherwise unset for the 2 touch points:
1. Loading the `white_listed_file_types` in submission context (student and instructor preview views).
  Behavior is modified to *always* pass the allowed file types list, instead of only for the "custom" file upload configuration.
2. Loading the `white_listed_file_types` string when editing ORA settings (ORA settings view)

**Why?**
This enables:
- Previewing the allowed extensions when configuring an ORA ([EDUCATOR-5344](https://openedx.atlassian.net/browse/EDUCATOR-5344))
- Previewing the allowed extensions when submitting a file response ([EDUCATOR-5345](https://openedx.atlassian.net/browse/EDUCATOR-5345))

Also added/updated some tests to reflect new functionality.

**How should I test this?**
Check the passed context at the following points for ORAs w/ file uploads enabled:
- LMS: `openassessment.xblock.submission_mixin.py::submission_path_and_context()`
- Studio: `openassessment.xblock.studio_mixin.py::editor_context()`

Verify context:
- ORA w/ `image` or `pdf-and-image` file upload type have extensions for `white_listed_file_types`
- ORA w/ `custom` file upload type has configured extensions for `white_listed_file_types`

Alternatively, surface the `white_listed_file_types` variable in Submission/Studio templates.

**Developer Checklist**
- [x] Translations up to date

JIRA: [EDUCATOR-5346](https://openedx.atlassian.net/browse/EDUCATOR-5346)

FIY: @edx/masters-devs-gta


openassessment/xblock/test/test_submission.py
```diff
@@ -964,7 +963,7 @@ def test_open_saved_response_misaligned_file_data(self, xblock):
         self._assert_path_and_context(
             xblock, 'openassessmentblock/response/oa_response.html',
             {
-                'text_response': 'required',
+                'text_response': None,
```
nsprenkle, line 966: This changed because I switched the scenario, needed to run the `parse_xml` import code.

openassessment/xblock/submission_mixin.py
```diff
@@ -806,7 +820,6 @@ def submission_path_and_context(self):
             context['team_file_urls'] = self.file_manager.team_file_descriptors(
                 team_id=team_id_for_current_submission
             )
-        if self.file_upload_type == 'custom':
```
Andytr1, line 809: Why are we removing this check?

openassessment/xblock/openassessmentblock.py
```diff
@@ -336,11 +336,22 @@ def file_upload_type(self, value):
         """
         self.file_upload_type_raw = value
 
+    def get_allowed_file_types_or_preset(self):
```
Andytr1, line 339: I am trying to wrap my head around terminology. Preset is refers to something we define / hard code and "allowed file type" is something the user defines if they choose "custom" ?

nsprenkle: Yes. Presets are allowing learners to upload `image` or `pdf-and-image` files, vs `custom` which allows course staff to specify what types of files learners can upload.

openassessment/xblock/submission_mixin.py
nsprenkle: We want to always surface the `white_listed_file_types` list when file uploads are enabled, not just for `custom` config but for presets as well.

openassessment/xblock/openassessmentblock.py
```diff
@@ -336,11 +336,22 @@ def file_upload_type(self, value):
...
+            return self.ALLOWED_FILE_EXTENSIONS
+        return None
+
     @property
     def white_listed_file_types_string(self):
         """
-        Join the white listed file types into comma delimited string
+        Join the allowed file types into comma delimited string
         """
+        self.white_listed_file_types = self.get_allowed_file_types_or_preset()
```
Andytr1, line 353: I am not sure if this line is OK, since we have an explicit setter for `white_listed_file_types_string` below.

nsprenkle: I was debating this piece. By default, it's a no-op (will adopt existing settings for `None` or `custom`), but ratchets forward by loading the preset on first fetch. Maybe the getter is the wrong place to do it but was the way I could think of that introduced the fewest changes.

Andytr1: What about creating a constructor and initializing that property there?

nsprenkle: The XBlock instructions say not to override the constructor function and I can't locate any other common entry point. 

Because of this, we'd end up having to use this preset logic in:
- Course import (loading from XML)
- Student view
- Staff view

We can hook each of these independently, but at the cost of a larger set of changes.

Andytr1 (approved):
Looks good. Would like a second opinion on getter with side effects - looks like this choice makes the most sense given the code base.

openassessment/xblock/openassessmentblock.py
jansenk, line 353: So, if I follow, this is doing the following:
 - If the block has a defined whitelist, use it
 - If the block /doesn't/ have a defined whitelist, grab the defaults and save them on the block so it now does have a whitelist.

I think it would be clearer if we put this logic somewhere line in `StudioMixin.update_editor_context()`. So in the future the whitelist is populated on save.
We can keep this to "backpopulate", but let's add some comments explicitly stating that that's what it's doing.

jansenk (changes requested):
I think we should make some changes in studiomixin for explicitness' sake

openassessment/xblock/openassessmentblock.py
nsprenkle: Agreed.

nsprenkle: ^ Updating w/ `update_editor_context` will happen as a part of [EDUCATOR-5344](https://openedx.atlassian.net/browse/EDUCATOR-5344), where we'll fill the allow list when a preset is selected, updating the XBlock on save. My goal here is to have an in-situ way of loading presets for non-updated ORAs.

jansenk: I'm gonna push back, I think this is the ticket for this change, but I'm down to have a discussion in standup tomorrow.

openassessment/xblock/studio_mixin.py
```diff
@@ -152,7 +156,7 @@ def editor_context(self):
             'file_upload_response': self.file_upload_response if self.file_upload_response else '',
             'necessity_options': self.NECESSITY_OPTIONS,
             'file_upload_type': self.file_upload_type,
-            'white_listed_file_types': self.white_listed_file_types_string,
```
jansenk, line 155: question. Here we were referencing `self.white_listed_file_types_string` and  in `get_submission_path_and_context()` we were referencing `self.white_listed_file_types` ... do we just store the same value on the block twice? is one a derived @property? Do we update them together? In `update_editor_context()` it only looks like we're saving `self.white_listed_file_types_string`?

jansenk (commented):
I still think we should be making a change in `update_editor_context` but I missed the meeting 🤷  as long as it eventually gets changed

nsprenkle:
> I still think we should be making a change in `update_editor_context` but I missed the meeting 🤷 as long as it eventually gets changed

I don't think we'll need the change in `update_editor_context`.  We're populating data in `editor_context`, effectively pre-filling the `edit` form. So when the UI change in [EDUCATOR-5344](https://openedx.atlassian.net/browse/EDUCATOR-5344) is complete, saving a change to ORA config will include the extension list automatically.

openassessment/xblock/studio_mixin.py
nsprenkle: Annoyingly, we store both the `white_listed_file_types` (an array of file extensions) and `white_listed_file_types_string` (a comma delimited list of extensions). The array is used for extension validation in `upload_url` and the string is used in the `editor_context` and serialization/deserialization from XML.

nsprenkle: The string is the actual persisted value, but the string setter also parses the string and sets the array variable for use around the ORA.

jansenk:
> I don't think we'll need the change in `update_editor_context`. We're populating data in `editor_context`, effectively pre-filling the `edit` form. So when the UI change in [EDUCATOR-5344](https://openedx.atlassian.net/browse/EDUCATOR-5344) is complete, saving a change to ORA config will include the extension list automatically.

Hmm, yeah, that'll work I guess. I don't think I have any actual Issues with this, i just think it'll be clearer what's happening if we set the value where we're setting the rest of the values.

nsprenkle:
> Hmm, yeah, that'll work. I don't think I have any actual Issues with this, i just think it'll be clearer what's happening if we set the value where we're setting the rest of the values.

That was an approach I considered, which is admittedly clearer about persisting the new values in code and more aggressive about pushing forward the new behavior. The reason I ended on the approach above is that it shows staff the changes before they set them which 1) enables one of the stories and 2) helps communicate the change to end-users.

Andytr1 (approved):
gtg

jnlapierre (approved):

PR #1513, opened 2020-10-02
Show file extensions for preset upload types in ORA settings

nsprenkle:
**TL;DR -** to make it clear to staff what file types will be allowed when configuring a file upload, show the allowed file extensions for images/pdfs in the ORA settings view.

**What changed?**

- Always show "allowed file types" box when file upload is enabled
- Show (but disable modifying) file types for "image" and "pdf-and-image" presets
- Add banner to explain how to modify allowed file extensions
- Modify file upload help text

Screen Shot 2020-10-02 at 17 01 47

**Developer Checklist**
- [x] Translations up to date

JIRA: [EDUCATOR-5344](https://openedx.atlassian.net/browse/EDUCATOR-5344)

FIY: @edx/masters-devs-gta


openassessment/templates/openassessmentblock/edit/oa_edit_basic_settings_list.html
```diff
@@ -42,21 +42,25 @@
...
                            value="{{ white_listed_file_types }}"
+                           {% if file_upload_type != "custom" %}disabled{% endif %}
                             />
                 </div>
                 <p class="setting-help">
                     {% trans "Enter the file extensions, separated by commas, that you want learners to be able to upload. For example: pdf,doc,docx." %}
                 </p>&nbsp;
                 <p class="setting-help message-status error"></p>
+                <p class="setting-help warning-banner extension-warning {% if file_upload_type == "custom" %}is--hidden{% endif %}">
+                    {% trans "To add more file extensions, select Custom File Types and enter the full list of acceptable file extensions you'd like to include." %}
```
Andytr1, line 62: nit: I am not a fan of "you'd like" in help text. I prefer more directional comments:
`and enter the full list of acceptable file extensions to be included.

Andytr1 (approved):
Code looks good.

openassessment/templates/openassessmentblock/edit/oa_edit_basic_settings_list.html
nsprenkle: Thanks, will run wording by Sapana 👍

jansenk:
Code looks good, doing some manual testing

jansenk (changes requested):
Code looks good. Two things I'd like to ask about though, one change request and one behavioral question.

1) When I enable File Uploads (choose required or optional) for the first time on a new ORA, the type dropdown is set to Images, but the box doesn't have file types in it. ~I feel like `oa_edit_settings::219` should be taking care of that? What is that line doing? I think you should change this before merging.~ Okay, I debugged and understand what that does now. I think you could add something around line 22ish that triggers the text input's `change` so the input types show up. 

2) When we select "Custom File Types" the box just becomes editable with the last-selected values in it.  Is this the desired behavior? I'd think we'd want to clear the box, but I'm not sure.

openassessment/templates/openassessmentblock/edit/oa_edit_basic_settings_list.html
nsprenkle: Updated 👍

nsprenkle:
> Code looks good. Two things I'd like to ask about though, one change request and one behavioral question.
> 
> 1. When I enable File Uploads (choose required or optional) for the first time on a new ORA, the type dropdown is set to Images, but the box doesn't have file types in it. ~I feel like `oa_edit_settings::219` should be taking care of that? What is that line doing? I think you should change this before merging.~ Okay, I debugged and understand what that does now. I think you could add something around line 22ish that triggers the text input's `change` so the input types show up.
> 2. When we select "Custom File Types" the box just becomes editable with the last-selected values in it.  Is this the desired behavior? I'd think we'd want to clear the box, but I'm not sure.

RE 1: added a manual trigger to refresh that val. Good catch 👍 
RE 2: I think that's the behavior we want so staff could start with a preset and extend it.

jansenk (approved):
Looks good!

PR #1514, opened 2020-10-05
Show allowed file extensions for file uploads

nsprenkle:
**TL;DR -** to make it clearer for students what file types they can upload, show allowed extensions next to the file select button.

**What changed?**

- Show allowed file extensions in upload UI
- Add `allow` values on file `<input/>` to soft-block selecting incompatible file types in file selector
- Updated wording on error banners. Since we already show the allowded file types, any bad uploads now trigger the error: "File upload failed: unsupported file type. Only the supported file types can be uploaded. If you have questions, please reach out to the course team."
- Minor refactoring

**Developer Checklist**
- [x] Translations up to date

JIRA: [EDUCATOR-5345](https://openedx.atlassian.net/browse/EDUCATOR-5345)

**Screenshots**
Shows new supported file types and error banner

Screen Shot 2020-10-05 at 17 16 41

FIY: @edx/masters-devs-gta


openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -591,14 +591,27 @@ OpenAssessment.ResponseView.prototype = {
...
 
-        for (var i = 0; i < files.length; i++) {
-            ext = files[i].name.split('.').pop().toLowerCase();
-            fileType = files[i].type;
+        var isUploadSupported = function(file, uploadType) {
+            var ext = file.name.split('.').pop().toLowerCase();
+            var fileType = file.type;
 
+            // Check upload type/extension matches allowed types
+            if (uploadType === 'image' && this.data.ALLOWED_IMAGE_MIME_TYPES.indexOf(fileType) === -1) {
```
jansenk, line 601: Could there ever be a case where a file is of the correct mime type but of a different file type and we allow it although it's not allowed by the file type list?

openassessment/xblock/submission_mixin.py
```diff
@@ -410,34 +410,59 @@ def upload_url(self, data, suffix=''):  # pylint: disable=unused-argument
...
-            return {'success': False, 'msg': self._(u"There was an error uploading your file.")}
+            raise KeyError()
+
         content_type = data['contentType']
         file_name = data['filename']
         file_name_parts = file_name.split('.')
-
-        file_num = int(data.get('filenum', 0))
-
         file_ext = file_name_parts[-1] if len(file_name_parts) > 1 else None
```
jansenk, line 451: I know that this is the existing behavior, but is there a more robust API for dealing with file types than just doing string split on periods?

openassessment/xblock/submission_mixin.py
```diff
@@ -410,34 +410,59 @@ def upload_url(self, data, suffix=''):  # pylint: disable=unused-argument
...
+                return {'success': False, 'msg': self._(
+                    "File upload failed: unsupported file type."
+                    "Only the supported file types can be uploaded."
+                    "If you have questions, please reach out to the course team."
+                )}
+        except KeyError:
+            return {'success': False, 'msg': self._("There was an error uploading your file.")}
+
+        # Attempt to upload
+        file_num = int(data.get('filenum', 0))
```
jansenk, line 425: nit: we could pass these into is_supported_upload_type rather than repeating those two lines

jansenk (commented):
Code looks good! I'll manually test in the AM

jansenk (approved):
Manual testing looks good! I'm still not sure about the mimetype/extension thing... but functionally for now this looks correct and good.

openassessment/xblock/static/js/src/lms/oa_response.js
jansenk: I guess this would mostly be covered by the input not allowing you to select things of non-supported extensions

openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -23,7 +23,10 @@ describe("OpenAssessment.ResponseView", function() {
     var FILE_TYPE_WHITE_LIST = ['pdf', 'doc', 'docx', 'html'];
     var FILE_EXT_BLACK_LIST = ['exe', 'msi', 'app', 'dmg'];
     var MAXIMUM_FILE_UPLOAD_COUNT = 20;
-    var COURSE_ID = 'course_id'
+    var COURSE_ID = 'course_id';
+
+    var BAD_FILETYPE_MESSAGE = 'File upload failed: unsupported file type. ' +
+        'Only the supported file types can be uploaded. If you have questions, please reach out to the course team.';
```
Andytr1, line 29: Nit: Perhaps remove `If you have questions, please reach out to the course team.` ?

Andytr1 (approved):
Looks good. Only optional suggestion: Error message update.

openassessment/xblock/static/js/src/lms/oa_response.js
nsprenkle: **TL;DR -** yes this could happen but I wouldn't expect it to cause any issues.

Right now, presets both in the frontend and backend are validated based on mime type and not file type. Only in the case of custom extensions do we block based on a bad file type. For presets, the shown extension list is more of a visual aide and hint to the file select menu than an explicit blocker. The file select menu in the browsers will automatically restrict you to the allowed filetypes but you could coerce it to a bizarre file type that still matches the mime type and I would expect it to go through w/out a problem.

openassessment/xblock/submission_mixin.py
nsprenkle: Which would we pass in?

openassessment/xblock/submission_mixin.py
nsprenkle: Yes. Have updated using `os.splitext()`

openassessment/xblock/submission_mixin.py
nsprenkle: Good call. Refactored.

openassessment/xblock/static/js/spec/lms/oa_response.js
nsprenkle: Planning to leave it, since it was provided by UX/Product.

PR #1518, opened 2020-10-15
EDUCATOR-5365: backwards compatibility for old ora submissions

jansenk:
**TL;DR -**  `OraDownloadData.collect_ora2_submission_files()` tries to access `answer['files_names']` and `answer['files_sizes'` which don't exist for ORA answers older than Jan 2020ish. This fixes this and fixes the linked bug.

**Testing instructions**
This is tricky. Create the default test ORA with uploads allowed on local or stage, and submit an answer with some number of files. ~Then , go into django admin and manually modify the submission.~  Actually, you can't edit submissions from django admin, you'll have to manually update it from mysql eww
 
1. Change the key `files_names` to `files_name`
2. Delete the `files_sizes` key and value.

Now, go to the instructor dashboard for the course and  click 'Generate Submission Files Archive'.
If you can download the file, the bug has been fixed.

**Developer Checklist**
- [x] Translations up to date

JIRA: [EDUCATOR-5365](https://openedx.atlassian.net/browse/EDUCATOR-5365)

FIY: @edx/masters-devs-gta

### DIFF
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -774,11 +774,17 @@ class OraDownloadData:
         all_submission_information = sub_api.get_all_course_submission_information(course_id, 'openassessment')
 
         for student, submission, _ in all_submission_information:
-            answer = submission['answer']
+            answer = submission.get('answer', dict())
 
             # collecting submission attachments with metadata
             for index, file_key in enumerate(answer.get('file_keys', [])):
-                file_name = answer['files_names'][index]
+                file_names = answer.get('files_names', answer.get('files_name', []))
+                file_name = file_names[index]
+
+                file_size = 0
+                file_sizes = answer.get('files_sizes')
+                if file_sizes:
+                    file_size = file_sizes[index]
 
                 yield {
                     'type': cls.ATTACHMENT,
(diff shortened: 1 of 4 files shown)


openassessment/data.py
```diff
@@ -774,11 +774,17 @@ def collect_ora2_submission_files(cls, course_id):
         all_submission_information = sub_api.get_all_course_submission_information(course_id, 'openassessment')
 
         for student, submission, _ in all_submission_information:
-            answer = submission['answer']
+            answer = submission.get('answer', dict())
 
             # collecting submission attachments with metadata
             for index, file_key in enumerate(answer.get('file_keys', [])):
-                file_name = answer['files_names'][index]
+                file_names = answer.get('files_names', answer.get('files_name', []))
```
Andytr1, line 781: I was confused by this on the first glance and then noticed files_name**s** vs files_name :(

jansenk:
Hard to do a before/after on something where the before is that nothing happens, but here's a screenshot of me successfully generating the report with my changes in place 
Screen Shot 2020-10-16 at 10 31 53 AM

nsprenkle (approved):
Would prefer more explicit comments in test data and `data.py` explaining how the formats differ and why the search for multiple keys. We understand this now, but it would be helpful to codify it so future devs get what's going on.

Otherwise, code looks good and have verified locally 👍

Andytr1 (approved):
GTG

jansenk:
I've created https://openedx.atlassian.net/browse/EDUCATOR-5382 to handle the more root cause of this issue.

PR #1528, opened 2020-10-28
fix manifest to include new webpack-compiled output files

jansenk:
**TL;DR -** we were deploying to pypi in a way that didn't include the new webpack output and it broke everything

**What changed?**

- we used to compile js to `openassessment-lms.min.js` and `openassessment-studio.min.js`
- as of 2.11.0 we compile them to `openassessment-lms.js` and `openassessment-studio.js`
- the manifest wasn't updated so we weren't picking up our comipled files anymore when we deployed to pypi

**Developer Checklist**
- [x] Translations up to date

JIRA: [EDUCATOR-5399](https://openedx.atlassian.net/browse/EDUCATOR-5399)

FIY: @edx/masters-devs-gta

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,7 +4,7 @@ include README.rst
 include setup.cfg
 include openassessment/xblock/static/css/*.css
 include openassessment/xblock/static/css/lib/backgrid/*.css
-include openassessment/xblock/static/js/openassessment*.min.js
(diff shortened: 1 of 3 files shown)


muselesscreator (approved):
...wow... that was it?!?!
Thanks for finding this!

jansenk:
> ...wow... that was it?!?!
> Thanks for finding this!

I _THINK_ that's it

PR #1575, opened 2021-01-21
Clean up attempt # 2

mzulqarnain1:
In this PR we are
- Dropping Python 3.5 support
- Upgrading code to Python 3.8
- Upgrading dependencies 
- Removing Python 3.5 tests
- Adding Django 3 in the test matrix

The first commit of this PR was generated by the cleanup-python-code Jenkins job, which ran
```
modernize_travis;modernize_tox;make upgrade;find . -type f -name '*.py' | while read fname; do pyupgrade --exit-zero-even-if-changed --keep-percent-format --py3-plus --py36-plus --py38-plus "$fname"; done
```


openassessment/assessment/api/peer.py
```diff
@@ -103,8 +103,8 @@ def submitter_is_finished(submission_uuid, peer_requirements):
         return False
     except PeerWorkflow.DoesNotExist:
         return False
-    except KeyError:
-        raise PeerAssessmentRequestError(u'Requirements dict must contain "must_grade" key')
+    except KeyError as ex:
```
jansenk, line 106: I'm not familiar with this, what does this change do?

mzulqarnain1: This is something recent versions of `pylint` have started enforcing, that we use Python 3’s exception chaining, in short, not using `raise from` makes the traceback inaccurate, because the message implies there is a bug in the exception-handling code itself, which is a separate situation than wrapping an exception. When we use `raise from`, the `__cause__` attribute is set, stating the original exception in traceback.

awais786 (approved):

awais786:
https://github.com/edx/edx-ora2/blob/master/setup.py#L54 remove this

mzulqarnain1:
> https://github.com/edx/edx-ora2/blob/master/setup.py#L54 remove this

Already Done.

PR #1586, opened 2021-02-03
[BD-05] [EDUCATOR-5478] [BB-3613] Use webpack to compile sass assets as well

xitij2000:
This PR gets rid of the existing python-based system for compiling SCSS in favour of a webpack based mechanism for js and scss. 

JIRA: [EDUCATOR-5478](https://openedx.atlassian.net/browse/EDUCATOR-5478)

**What changed?**

- This PR changes the webpack config to also compile the sass files
- It also updates the webpack config to output to a new directory instead of placing output files in the same dir. 
- It updates the python code to find the compiled assets in the new place
- It removes the old python-based scripts for compiling sass

**Developer Checklist**

- [ ] Reviewed the [release process](./release_process.md)

**Testing Instructions**

- Install the PR version of the block
- Make sure that the block still works without issues. 

**Reviewer Checklist**

FYI: @edx/masters-devs-gta


edx-webhook:
Thanks for the pull request, @xitij2000! I've created [OSPR-5601](https://openedx.atlassian.net/browse/OSPR-5601) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

giovannicimolin:
FYI @jnlapierre @muselesscreator ^

shimulch (approved):
LGTM :+1:

- [x] I tested this. Pulled in local devstack, checked if ora2 works correctly. 
- [x] I read through the code
- [N/A] I checked for accessibility issues
- [x] Includes documentation
- [N/A] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

xitij2000:
Note: I went for compiled since 'dist' and 'build' are ignored directories. If it's preferable to use one of those names, I can update the ignore patterns instead.

muselesscreator:
@xitij2000 I would definitely lean towards un-ignoring the `dist` folder and going with that name.

**Nit Question:**
Also, would we like to move towards the model of importing the css into the JS to make entry-specific css, or are we comfortable keeping them separated?  I know @giovannicimolin was talking about wanting to be able to import css in the JS, which I tend to think is a reasonable paradigm to follow when we don't need the css importable separately

muselesscreator (changes requested):
LGTM, once the compiled folder is switched to a more standard (dist) naming.

giovannicimolin:
This seems to break the studio styles, can anyone else confirm?

xitij2000:
> Also, would we like to move towards the model of importing the css into the JS to make entry-specific css, or are we comfortable keeping them separated?

I think that's a bit complicated in this case since the rtl and ltr stylesheets are separate and which css file is attached depends on the python code. This will need more changes to move to the javascript realm.

muselesscreator (commented):
Dir updates look fine to me as long as it is tested across all views before merge.

muselesscreator (approved):

giovannicimolin:
@xitij2000 I'm still seeing issues in the Studio styling:
image

Can you investigate and fix the issue?

muselesscreator (changes requested):
Please address Giovanni's concerns re: studio styles before merging.

shimulch:
@giovannicimolin,

The output of the new webpack build and the current [libsass script](https://github.com/edx/edx-ora2/blob/master/scripts/compile_sass.py) build is not the same. It seems the webpack build adds ``[dir]`` selectors on the output CSS which were not present in previous builds. Due to this a lot of the styles were being ignored as the root element of the ORA edit popup doesn't contain the ``dir`` attribute.

I've tried to fix the issue by removing the ``[dir]`` selector from the output, but I couldn't figure out any way to do so. 

Then I've added ``dir`` attribute on the edit modal itself via Javascript. This solution seems to be working. Could you review this and see if this is an acceptable approach?

giovannicimolin:
@shimulch @xitij2000 I've had some time to kill and found out why the issue was happening:
The `fedx-build` scripts come with the `PostCssRtlPlugin` in the CSS post processing stages, which added those pesky `[dir]` selectors. Removing it fixed the issues.

I've tried to do it in a reasonably clean way, let me know if I can improve the approach.

@muselesscreator Can you do a review pass too?

webpack.prod.config.js
```diff
@@ -2,40 +2,97 @@ process.env.BABEL_ENV = 'production';
...
     ],
+  },
+};
+
+// Merge back to configuration
+config = mergeWithRules({
+  module: {
+    rules: {
+      test: 'match',
+      use: 'replace',
```
shimulch, line 64: Is it possible to replace only the `postcss-loader`? Like -

```
use: {
    loader: 'match',
    option: 'replace'
}
```
[Reference](https://github.com/survivejs/webpack-merge#mergewithrules).

If that works then we might not need to define other rules.

shimulch:
@giovannicimolin, Thanks for finding the actual issue. But I think we have to add this rule in `webpack.dev.config.js` too, right?

webpack.prod.config.js
giovannicimolin: @shimulch It doesn't work, not sure why. 
`MiniCssExtractPlugin.loader` and `resolve-url-loader` break if I do it.

giovannicimolin:
@shimulch I don't think that the dev config even works, but I'll update. :stuck_out_tongue:

giovannicimolin:
@shimulch I've rebased the PR, re-tested all changes and made some code formatting adjustments.
TinyMCE and the custom styles are still working as expected.

shimulch (approved):
Thanks for fixing the issue, @giovannicimolin.  LGTM :+1:

- [x] I tested this in my local devstack. Checked if the sass compilation provides correct output. Studio and LMS views are working fine.
- [x] I read through the code
- [n/a] I checked for accessibility issues
- [n/a] Includes documentation

giovannicimolin:
@muselesscreator @jnlapierre This is ready for review.

muselesscreator (approved):
Looks fine to me if Giovanni's concerns have been met.

giovannicimolin:
@muselesscreator I actually added the last few commits to fix the issue, so I don't feel ok approving my own work :stuck_out_tongue_closed_eyes: 

@shimulch Reviewed and approved the changes, so it's good for your final checks. Once that is ok I'll rebase and prepare to merge.

giovannicimolin (changes requested):
While doing a full testing round, I've found one issue with the prompt editor in Studio.

Steps to reproduce: 
1. Add a new ORA problem on Studio.
2. Enable the WYSIWYG editor.
3. Check that the prompt is not visible anymore, and that TinyMCE doesn't load to edit prompts anymore.
image

I have not checked if this bug is present in the `master` branch, so far it seems like the new static compilation process is mixing up the two TinyMCE versions (response editor and prompt editor).

xitij2000:
After rebasing on the latest master (v3.3.1) the issue seems to be fixed.

giovannicimolin (approved):
@xitij2000 
:+1: 

- [x] I tested this:
I've done spot checks on the following ORA pages (with and without WYSIWYG editor):
- Student view
- Studio view
- Studio settings
- Instructor dashboard
- [x] I read through the code
- [x] I checked for accessibility issues
- [ ] ~~Includes documentation~~ NA

giovannicimolin:
@muselesscreator @jnlapierre This is ready for review, testing and merge.
The issue I reported about styles is now fixed here and the editor issue was fixed on https://github.com/edx/edx-ora2/pull/1601

jnlapierre (approved):
LGTM

PR #1590, opened 2021-02-09
EDUCATOR-4827: Toggle rubric visibility for learners

jansenk:
**TL;DR -** Add an ORA setting to allow users to see a collapsable grading rubric while they work on their response

**What changed?**

- Added a field to ORA, `show_rubric_during_response`, that can be set to true or false in the ORA settings in studio
- Added functionality to oa_rubric.html to be rendered in "read-only" mode, which looks the same except that all input elements are hidden. 
- When `show_rubric_during_response` is set to true, render the rubric in read-only mode in a collapsable container at the top of the response section

**Testing Instructions**

- Create an ORA. In the settings page, scroll down to the bottom and set `Show Rubric During Response` to be true.
- Publish the ORA, and then open it as a learner
- The `What will this assignment be graded on?` section should appear in the Response section, and clicking it should display the rubric with no inputs (radio buttons, text inputs, etc)

Screen Shot 2021-02-19 at 11 46 52 AM

Screen Shot 2021-02-19 at 11 49 19 AM

Screen Shot 2021-02-19 at 11 49 27 AM

**Developer Checklist**
- [ ] Reviewed the [release process](./release_process.md)
- [ ] Translations up to date
- [ ] JS minified, SASS compiled
- [ ] Test suites passing on Jenkins
- [ ] Bumped version number in [setup.py](../setup.py) and [package.json](../package.json)
- [ ] UI/UX review

JIRA: [EDUCATOR-4827](https://openedx.atlassian.net/browse/EDUCATOR-4827)

FIY: @edx/masters-devs-gta


openassessment/templates/openassessmentblock/oa_rubric.html
```diff
@@ -7,29 +7,35 @@
...
-                                       class="answer__value rubric_{{ criterion.order_num }}_{{ option.order_num }}"
-                                       id="{{ rubric_type }}__assessment__rubric__question--{{ criterion.order_num }}__{{ option.order_num }}__{{ xblock_id }}"
-                                       value="{{ option.name }}"
-                                       name="{{ rubric_type }}__assessment__rubric__question--{{ criterion.order_num }}"
-                                       aria-describedby="{{ rubric_type }}__assessment__rubric__meta--{{ criterion.order_num }}__{{ option.order_num }}__{{ xblock_id }}"/>
-                                <label for="{{ rubric_type }}__assessment__rubric__question--{{ criterion.order_num }}__{{ option.order_num }}__{{ xblock_id }}"
-                                       class="answer__label"
-                                       >{{ option.label }}</label>
+                                {% if rubric_type == 'read_only' %}
+                                    <label for="{{ rubric_type }}__assessment__rubric__meta--{{ criterion.order_num }}__{{ option.order_num }}__{{ xblock_id }}"
```
jansenk, line 24: easy to miss, but the reason I don't just gate off the `<input>` and leave the `<label>` is because here I also point this label's `for` at the proceeding `<div>` rather than the `<input>`

(i say easy to miss because when I came back to this i was totally confused why i did this)

openassessment/xblock/schema.py
```diff
@@ -169,5 +169,6 @@ def datetime_validator(value):
                 })
             ]
         })
-    ]
+    ],
+    Required('show_rubric_during_response'): bool,
```
nsprenkle, line 173: Should this have a default value? Or will older XBlocks automatically populate this value on change?
```suggestion
    Required('show_rubric_during_response', default=False): bool,
```

openassessment/xblock/static/sass/oa/utilities/_developer.scss
```diff
@@ -1127,6 +1127,51 @@
...
+        // fields
+        .assessment__fields {
+          margin-bottom: $baseline-v;
+        }
+    
+        // rubric question
+        .assessment__rubric__question {
+          @extend %ui-rubric-question;
+
+
```
nsprenkle, line 1165: Nit: clean up newlines throughout this addition.

nsprenkle (commented):
Still reviewing but I wanted to surface some initial questions in case I get diverted.

nsprenkle:
I also think you'll probably need to update `openassessment/xblock/xml.py` to allow for XML import/export of the new setting.

openassessment/xblock/schema.py
jansenk: I believe it will be automatically populated, but there's no harm in being more explicit

openassessment/templates/openassessmentblock/oa_rubric.html
```diff
@@ -7,29 +7,35 @@
             class="field field--radio is--required assessment__rubric__question is--showing ui-slidable__container {% if criterion.options %}has--options{% endif %} {{ rubric_type }}__assessment__rubric__question--{{ criterion.order_num }}"
         >
             <div class="question__title ui-slidable__control">
-                <button class="ui-slidable" aria-expanded="true" id="oa_rubric__{{ rubric_type }}__{{ submission.uuid }}__{{ criterion.order_num }}" aria-controls="oa_rubric__{{ rubric_type }}__{{ submission.uuid }}__{{ criterion.order_num }}__content">
+                <button class="ui-slidable" aria-expanded="true" id="oa_rubric__{{ rubric_type }}__{{ submission.uuid|default:'dummy_uuid' }}__{{ criterion.order_num }}" aria-controls="oa_rubric__{{ rubric_type }}__{{ submission.uuid|default:'dummy_uuid' }}__{{ criterion.order_num }}__content">
```
nsprenkle, line 10: Is `dummy_uuid` needed for testing or is this to fix some breakage somewhere else?

nsprenkle (commented):
Visual code review looks good 👍 
I'll return after verifying funcitonality locally

openassessment/templates/openassessmentblock/oa_rubric.html
jansenk: It's just a placeholder value for when we render the template in the submission screen, since there's no submission to get a uuid from

nsprenkle:
UX nit: In terms of workflow, I'd guess a learner would want to know what the prompt is before they know the rubric. Therefore, I'd argue that this makes sense to put the rubric expand between the prompt and response (first screenshot).

Stylistic nit: even ignoring the above, I think the padding needs touching up (right now there's a chasm between the bottom of the "What will this assignment be graded on?" copy and the next section, where all the other sections have pretty consistent spacing (second and third screenshots or before/after spacing).

rubric-placement

original-spacing

more-consistent-spacing

nsprenkle (approved):
Tested locally, works as described 👍 
Some fit-and-finish suggestions but they could be postponed to a v2.

jansenk:
> UX nit: In terms of workflow, I'd imagine a learner would want to know what the prompt is before they know the rubric. Therefore, I'd argue that this makes sense to put the rubric expand between the prompt and response (first screenshot).

I don't disagree, but the ticket specifies the div should be before any prompts. @sapanathomas523 thoughts?

jnlapierre (approved):
LGTM (Code reviewed)

jansenk:
Looks a bit better now 
Screen Shot 2021-02-24 at 3 59 07 PM

nsprenkle:
> UX nit: In terms of workflow, I'd imagine a learner would want to know what the prompt is before they know the rubric. Therefore, I'd argue that this makes sense to put the rubric expand between the prompt and response (first screenshot).
> 
> Stylistic nit: even ignoring the above, I think the padding needs touching up (right now there's a chasm between the bottom of the "What will this assignment be graded on?" copy and the next section, where all the other sections have pretty consistent spacing (second and third screenshots or before/after spacing).
> 
> rubric-placement
> 
> original-spacing
> 
> more-consistent-spacing

These have all been resolved 👍

PR #1592, opened 2021-02-10
Disallow combined upload and submit step

nsprenkle:
**TL;DR -** remove the shortcut for uploading and submitting in a single step, which can cause broken submission states, plus a bunch of adjacent JS cleanup.

JIRA: [EDUCATOR-5494](https://openedx.atlassian.net/browse/EDUCATOR-5494)

**What changed?**
- When descriptions change, there's a check to see if we can upload and/or submit. Remove that so state of staging files blocks submitting.
- Remove redundant logic in submit button that prompts to upload files if files are staged.
- Significantly restructured submit logic to remove deprecated jQuery calls, unnecessary variable binding, and increase readability.
- For `window.confirm()`s, pass the result directly instead of wrapping in `$.Deffered()`s

**Developer Checklist**

**Testing Instructions**

1. Go to an ORA with optional/no text response but file uploads required
2. Select some files to upload
3. Provide file descriptions, giving all files descriptions should enable the `Upload` button
4. Verify that the `Submit` button is still disabled - previously it would have been enabled
5. Upload files, verify the`Submit` button enables after the files have successfully uploaded

Bonus points:
1. Repeat steps 1-4 above.
2. Cheat. Manually enable the `Submit` button by removing the `disabled` prop from the button.
3. Click `Submit`, verify that you are not presented with the option to upload files and that the submission does not progress.

**Reviewer Checklist**

FYI: @edx/masters-devs-gta


openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -211,7 +211,8 @@ export class ResponseView {
             && !textFieldsIsNotBlank && !filesFiledIsNotBlank) {
         readyToSubmit = false;
       }
-      if (this.hasPendingUploadFiles() && !this.collectFilesDescriptions()) {
+      if (this.hasPendingUploadFiles()) {
+        this.collectFilesDescriptions();
```
nsprenkle, line 215: Preemptive note: we still need to call `collectFilesDescriptions()` each time because it's coupled with the `Upload` button enable logic. This just removes it's use for short-circuiting submit-ability.

jnlapierre (approved):
LGTM (Code review)

nsprenkle:
Substantial change invalidates @jnlapierre's earlier approval. Re-requesting review.

openassessment/xblock/static/js/spec/lms/oa_response.js
```diff
@@ -892,10 +892,12 @@ describe("OpenAssessment.ResponseView", function() {
...
         $(firstDescriptionField2).val('test2');
 
-        // user finishes adding descriptions which enables submit button
+        // user finishes adding descriptions which enables the upload button
         view.checkSubmissionAbility(true);
-        expect(view.submitEnabled()).toBe(true);
         expect(view.uploadEnabled()).toBe(true);
+
+        // the submit button remains disabled until files are uploaded
+        expect(view.submitEnabled()).toBe(false);
```
muselesscreator, line 900: Should there be a test for when files are uploaded?

nsprenkle: A "re-enable submit button after upload" step?

nsprenkle: I'd argue that is already covered by existing tests. See "tests that new file uploads are appended". These changes are only concerned with Submit/Upload button states while staging.

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -487,47 +488,35 @@ export class ResponseView {
...
-
-      // If the submission was submitted successfully, move to the next step
+      view.confirmSubmission()
+        // Proceed with submission
+        .done(() => {
+          const submission = view.response();
+          baseView.toggleActionError('response', null);
+
+          // Send the submission to the server, async
+          submitDefer = view.server.submit(submission);
```
jansenk, line 505: I've been looking at this and line 512 and I can't prove that there's something wrong with this, but I can't shake the feeling.

jansenk: so submitDefer is now the promise returned by server.submit(), but hasn't line 212 already executed by this point? I might just be confusing myself. Any chance you'd put that `.done().fail()` just directly on the `server.submit()` call?  we don't really need the defer variable anymore, right?

nsprenkle: I wrote it this way because I find the "flatter" layout more readable but yes, we could nest the `done` and `fail` calls. 

IMHO, it would be even better if we changed the `confirmSubmission` to not use a `$.Deferred()` since it's just wrapping a `window.confirm` and doesn't really _need_ to be async. That way we could have just the one layer of promises on the `server.submit()` call.

openassessment/xblock/static/js/src/lms/oa_response.js
```diff
@@ -801,25 +772,23 @@ export class ResponseView {
...
-     return view.confirmRemoveUploadedFile(filenum).done(() => view.server.removeUploadedFile(filenum).done(() => {
-       const sel = $('.step--response', view.element);
-       const block = sel.find(`.submission__answer__file__block__${filenum}`);
-       block.html('');
-       block.prop('deleted', true);
-       view.checkSubmissionAbility();
-     }).fail((errMsg) => {
-       view.baseView.toggleActionError('delete', errMsg);
-     }));
+     if (this.confirmRemoveUploadedFile(filenum)) {
```
nsprenkle, line 775: This formerly relied on another `$.Deferred()`-wrapped `window.confirm()` and having both paradigms in the tests was gross, so I updated this case as well.

muselesscreator (approved):
niiiice

jansenk (approved):
lgtm

jnlapierre (approved):
LGTM, had to remember "deferred" but the changes looked clean.

PR #1596, opened 2021-02-19
Update filename encoding to handle more characters

nsprenkle:
**TL;DR -** Having a comma in a filename was causing download issues in Chrome. Update file name encoding to handle a broader range of special characters.

JIRA: [EDUCATOR-5592](https://openedx.atlassian.net/browse/EDUCATOR-5592)

**What changed?**

- Switch `encodeURI` to `encodeURIComponent` to encode additional characters, like the offending `,`
- Add additional encoding logic from this [MDN-suggested solution](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#encoding_for_content-disposition_and_link_headers)
- Add JS test for filename encoding

**Note:** I've tested this with a wide set of special characters, none of which I've tested break download/upload, but not all survive the encoding process. Specifically, I've noticed `|`, `?`, `~`, and `*` appear on the file link but not in the downloaded file. We may add this to any file info text on the page, something like "avoid using special characters in filenames, as they may not show up properly in downloads."

e.g. a file called `tëšt|,png*.png` will download as `tëšt_,png_.png`.

**Developer Checklist**

**Testing Instructions**

1. Configure testing environment to use S3 or an S3 stand-in (see [devstack/nsprenkle/localstack](https://github.com/edx/devstack/tree/nsprenkle/localstack))
2. In Chrome, upload files with different special characters and non-ASCII characters, for example `tëšt|^``#,;?@&=+$-_.!~*'().png`
3. Verify that upload/downloads work with (mostly) preserved filenames
4. Repeat on different browsers

**Reviewer Checklist**

FYI: @edx/masters-devs-gta


openassessment/xblock/static/js/spec/lms/oa_file_upload.js
```diff
@@ -46,7 +49,31 @@ describe("OpenAssessment.FileUploader", function() {
...
+
+        // Verify that a PUT request was sent with the right parameters
+        expect($.ajax).toHaveBeenCalledWith({
+            url: TEST_URL,
+            type: 'PUT',
+            data: specialCharacterFile,
+            async: false,
+            processData: false,
+            contentType: 'image/jpeg',
+            headers: { 'Content-Disposition': `attachment; filename*=UTF-8''${(FILE_NAME_SPECIAL_ENCODED)}` },
```
jansenk, line 75: we don't really care that these specific characters have been sent , we really just care that the name has been encoded.

on the other hand, if we made the line `headers: { 'Content-Disposition': `attachment; filename*=UTF-8''${fileUploader.encodeRFC5987ValueChars(FILE_NAME_SPECIAL_CHARS)}` },`, we're pretty much just duplicating the functionality in a test, which we also don't want to do. 

ultimately, don't really matter. go with your heart .....unless someone else feels strongly about this

nsprenkle: Yeah, I'm torn about how to test this, but at least this will break if we make any changes to the encoding and force us to go revisit the test.

jnlapierre: I like tests that break when something is wrong.

jnlapierre (approved):
LGTM (Code review)

jansenk:
I'm having a hell of a time debugging, but I have a file called `tëšt|,png*.png` and I am not able to upload, I am getting an error event of `TypeError: Failed to execute 'setRequestHeader' on 'XMLHttpRequest': String contains non ISO-8859-1 code point.`

nsprenkle:
> I'm having a hell of a time debugging, but I have a file called `tëšt|,png*.png` and I am not able to upload, I am getting an error event of `TypeError: Failed to execute 'setRequestHeader' on 'XMLHttpRequest': String contains non ISO-8859-1 code point.`
> 
> It looks like this is blowing up before we make the request to AWS because I don't see the actual request being made.

That suggests to me that something like the ë isn't getting encoded correctly... Maybe you have old JS loaded somehow, or I built some of it incorrectly?

Can you try `make javascript` and re-installing/restarting locally to double check it's not just stale JS?

jansenk:
> Can you try `make javascript` and re-installing/restarting locally to double check it's not just stale JS?

Yep, I'm in the process of doing that now, fingers crosed

jansenk (approved):
After reinstalling locally (which, gosh, i swear I did already) everything looks good! Tried a whole buncha different characters

PR #1622, opened 2021-04-29
test modify js rubric stuff

jansenk:
test for automatically setting the rubric js ui to an arbitary value

click the "do something!!!!" div to replace the contents of a rubric tab with some values defined in code


jansenk:
Two questions:

- [ ] Do we need to modify stuff in the "Peer Review" step?
- [ ] We should probably figure out a way to silence the "notifier" alerts. Or, at least, hide them?

PR #1625, opened 2021-05-06
fix: handling of ghost files in submission archive

leangseu-edx:
**TL;DR -** 
Ghost files can appear in submission archive for students who experience upload failures. Update submission archive to not include (and warn) about missing files.

JIRA: [EDUCATOR-5581](https://openedx.atlassian.net/browse/EDUCATOR-5581)

**What changed?**

- Add valid column to the csv as an indication of error files
- Log warning on archiving invalid files

**Developer Checklist**

**Testing Instructions**

- `pytest openassessment/tests/test_data.py`
- The additional test function `test_create_zip_with_failed_attachments`

**Reviewer Checklist**

FYI: @edx/masters-devs-gta

### DIFF
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -10,7 +10,7 @@ from itertools import chain
 import csv
 import json
 import os
-
+import logging
 import requests
 
 from django.conf import settings
@@ -24,6 +24,7 @@ from openassessment.assessment.models import Assessment, AssessmentFeedback, Ass
 from openassessment.fileupload.api import get_download_url
 from openassessment.workflow.models import AssessmentWorkflow, TeamAssessmentWorkflow
 
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 class CsvWriter:
     """
@@ -945,17 +946,20 @@ class OraDownloadData:
         'description',
         'size',
         'file_path',
+        'valid',
     )
 
     @classmethod
     def _download_file_by_key(cls, key):
+        url = get_download_url(key)
+        if not bool(url):
+            raise FileMissingException
         download_url = urljoin(
-            settings.LMS_ROOT_URL, get_download_url(key)
+            settings.LMS_ROOT_URL, url
         )
 
         response = requests.get(download_url)
         response.raise_for_status()
-
         return response.content
 
     @classmethod
@@ -974,7 +978,7 @@ class OraDownloadData:
             │   │   │   ├── SomeFile1
             │   │   │   └── SomeFile2
             │   │   ├── part_0.txt
-            │   │   └── part_1.txt
+            │   │  n └── part_1.txt
             │   └── StudentId2
             │       ├── part_0.txt
             │       └── part_1.txt
@@ -993,15 +997,26 @@ class OraDownloadData:
 
         with ZipFile(file, 'w') as zip_file:
             for file_data in submission_files_data:
-                file_content = (
-                    cls._download_file_by_key(file_data['key'])
-                    if file_data['type'] == cls.ATTACHMENT
-                    else file_data['content']
-                )
+                key = file_data['key']
+                file_name = file_data['file_path']
 
-                zip_file.writestr(file_data['file_path'], file_content)
+                try:
+                    file_content = (
+                        cls._download_file_by_key(file_data['key'])
+                        if file_data['type'] == cls.ATTACHMENT
+                        else file_data['content']
+                    )
+                except FileMissingException:
+                    file_data['valid'] = False
+                    # added a header to csv file to indicate that the file in valid or not.
+                    # not sure if I should create a {file_name}.error.txt to indicate the file error more clearly.
+                    logging.warning('File: {file_name} with key: {key} was missing.'
+                                    .format(file_name=file_name, key=key))
+                else:
+                    file_data['valid'] = True
+                    zip_file.writestr(file_name, file_content)
+
                 csvwriter.writerow(file_data)
-
             downloads_csv_path = os.path.join(str(course_id), 'downloads.csv')
 
             zip_file.writestr(
@@ -1206,6 +1221,8 @@ ZIPPED_LIST_SUBMISSION_VERSIONS = [
 class VersionNotFoundException(Exception):
     """ Raised when we are unable to resolve a given submission to a submission version """
 
+class FileMissingException(Exception):
+    """ Raise when file is not found on generated CSV"""
 
 class ZippedListSubmissionAnswer(OraSubmissionAnswer):
     """
--- a/openassessment/tests/test_data.py
+++ b/openassessment/tests/test_data.py
@@ -22,7 +22,7 @@ import openassessment.assessment.api.peer as peer_api
 from openassessment.data import (
     CsvWriter, OraAggregateData, OraDownloadData, SubmissionFileUpload, OraSubmissionAnswerFactory,
     VersionNotFoundException, ZippedListSubmissionAnswer, OraSubmissionAnswer, ZIPPED_LIST_SUBMISSION_VERSIONS,
-    TextOnlySubmissionAnswer
+    TextOnlySubmissionAnswer, FileMissingException
 )
 from openassessment.test_utils import TransactionCacheResetTest
 from openassessment.tests.factories import *  # pylint: disable=wildcard-import
@@ -1315,6 +1315,58 @@ class TestOraDownloadDataIntegration(TransactionCacheResetTest):
         self.assertEqual(
             zip_file.read(self.submission_files_data[7]['file_path']),
             self.answer_text.encode('utf-8')
+        )
+
+        self.assertTrue(zip_file.read(os.path.join(COURSE_ID, 'downloads.csv')))
+
+    def test_create_zip_with_failed_attachments(self):
+        file = BytesIO()
+
+        with patch(
+            'openassessment.data.OraDownloadData._download_file_by_key'
+        ) as download_mock:
+            download_mock.side_effect = FileMissingException
+            OraDownloadData.create_zip_with_attachments(file, COURSE_ID, self.submission_files_data)
+
+            download_mock.assert_has_calls([
+                call(self.file_key_5),
+                call(self.file_key_4),
+                call(self.file_key_1),
+                call(self.file_key_2),
+                call(self.file_key_3),
+            ])
+
+        zip_file = zipfile.ZipFile(file)
+
+        # archive should contain only three parts text file and one csv because all of the attachments are invalid
+        self.assertEqual(len(zip_file.infolist()), 4)
+
+        # check for pre_file_name_user's text file
+        self.assertEqual(
+            zip_file.read(self.submission_files_data[1]['file_path']),
+            self.answer_text.encode('utf-8')
+        )
+        self.assertEqual(
+            zip_file.read(self.submission_files_data[3]['file_path']),
+            self.answer_text.encode('utf-8')
+        )
+
+        # main user's text response
+        self.assertEqual(
+            zip_file.read(self.submission_files_data[7]['file_path']),
+            self.answer_text.encode('utf-8')
+        )
+
+        # expect the submission file to be invalid
+        self.assertEqual(
+            self.submission_files_data[6]['valid'],
+            False
+        )
+
+        # expect file not found in the zip file
+        self.assertRaises(
+            KeyError,
+            lambda: zip_file.read(self.submission_files_data[6]['file_path']),
         )
 
         self.assertTrue(zip_file.read(os.path.join(COURSE_ID, 'downloads.csv')))


openassessment/data.py
```diff
@@ -974,7 +978,7 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
             │   │   │   ├── SomeFile1
             │   │   │   └── SomeFile2
             │   │   ├── part_0.txt
-            │   │   └── part_1.txt
+            │   │  n └── part_1.txt
```
nsprenkle, line 981: Is this intentional?

leangseu-edx: must have been debugging `next` :sweat_smile:

openassessment/data.py
```diff
@@ -945,17 +946,20 @@ class OraDownloadData:
         'description',
         'size',
         'file_path',
+        'valid',
     )
 
     @classmethod
     def _download_file_by_key(cls, key):
+        url = get_download_url(key)
+        if not bool(url):
```
jansenk, line 955: you generally don't need to do this, `if not url:` is sufficient

openassessment/data.py
```diff
@@ -993,15 +997,26 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
...
-                    cls._download_file_by_key(file_data['key'])
-                    if file_data['type'] == cls.ATTACHMENT
-                    else file_data['content']
-                )
+                key = file_data['key']
+                file_name = file_data['file_path']
+
+                try:
+                    file_content = (
+                        cls._download_file_by_key(file_data['key'])
```
jansenk, line 1005: may as well make this `key` now

openassessment/data.py
```diff
@@ -993,15 +997,26 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
...
+                        if file_data['type'] == cls.ATTACHMENT
+                        else file_data['content']
+                    )
+                except FileMissingException:
+                    file_data['valid'] = False
```
jansenk, line 1010: I think that `file_found` (or something like that) might be a better header than `valid`, since `valid` kinda implies there's something wrong with the file. There's nothing necessarily wrong with the file itself, we just can't find it.

openassessment/data.py
```diff
@@ -993,15 +997,26 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
...
+                    # added a header to csv file to indicate that the file in valid or not.
+                    # not sure if I should create a {file_name}.error.txt to indicate the file error more clearly.
+                    logging.warning('File: {file_name} with key: {key} was missing.'
```
jansenk, line 1013: I'd also include the user, course, and the block location so that this error message can be more easily searched. Unless that's already included by default in the prefix? I don't think it is.

openassessment/data.py
```diff
@@ -993,15 +997,26 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
 
         with ZipFile(file, 'w') as zip_file:
             for file_data in submission_files_data:
-                file_content = (
-                    cls._download_file_by_key(file_data['key'])
-                    if file_data['type'] == cls.ATTACHMENT
-                    else file_data['content']
-                )
+                key = file_data['key']
+                file_name = file_data['file_path']
```
jansenk, line 1001: nit: `file_name` -> `file_path` just for consistency's sake, plus, technically I think it is actually a file path

openassessment/tests/test_data.py
```diff
@@ -1319,6 +1319,58 @@ def test_create_zip_with_attachments(self):
...
+        )
+
+        # main user's text response
+        self.assertEqual(
+            zip_file.read(self.submission_files_data[7]['file_path']),
+            self.answer_text.encode('utf-8')
+        )
+
+        # expect the submission file to be invalid
+        self.assertEqual(
```
jansenk, line 1361: hmm, I didn't catch this till now, but I don't think we should be modifying the passed-in data dictionary. Do we have any tests that test the contents of the generated CSV file?

openassessment/tests/test_data.py
```diff
@@ -1319,6 +1319,58 @@ def test_create_zip_with_attachments(self):
...
+            self.submission_files_data[6]['valid'],
+            False
+        )
+
+        # expect file not found in the zip file
+        self.assertRaises(
+            KeyError,
```
jansenk, line 1368: There's only a few submission files, I'd say you might as well just make sure they're all missing. Also, this is a total stylistic nit, but I think the context manager form of this looks nicer:
```
with self.assertRaises(KeyError): 
   zip_file.read()
```

jansenk: Also, if you'd like, it looks like you can also use Path (https://docs.python.org/3/library/zipfile.html#path-objects) to do something like 
```
self.assertFalse(Path(zip_file, <file_path>). exists()
```

jansenk (changes requested):
My one main request is changing the name of the CSV header from `valid` to something like `file_found`. I'd also request that you don't modify the passed in data dictionary, and instead make a copy, add the field, and then write to CSV.

openassessment/data.py
jansenk: course_id, block_id, and student_id should all be available to you in file_data

openassessment/tests/test_data.py
jansenk: this may be a harder ask than I originally intended, since there's not really any prior art on testing this CSV (the stuff at the top of the file is unrelated). If I pointed you at some libraries we use and some code we've written elsewhere, would you be down to try adding some tests for the content of the CSV? Other option: you could make another test where you call the CSV function with just like, one user and one or two uploaded files, and then just assert that the text of the CSV is equal to another static string you define in the test

leangseu-edx: I will make a copy of data dictionary instead. Also, You can pointed the libraries and the code elsewhere. I do not have preference on this.

openassessment/data.py
```diff
@@ -993,15 +999,37 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
 
         with ZipFile(file, 'w') as zip_file:
```
nsprenkle, line 1000: Update docstring to note that missing files will not be included in the archive.

openassessment/data.py
```diff
@@ -993,15 +999,37 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
...
             for file_data in submission_files_data:
```
nsprenkle, line 1001: The `csvwriter` needs to be updated with the new header field. Right now it won't write `file_found` header so that information will be stripped from the CSV.

openassessment/data.py
```diff
@@ -993,15 +999,37 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
...
+                        cls._download_file_by_key(key)
+                        if clone_file_data['type'] == cls.ATTACHMENT
+                        else clone_file_data['content']
+                    )
+                except FileMissingException:
+                    clone_file_data['file_found'] = False
+                    # added a header to csv file to indicate that the file was found or not.
+                    # not sure if I should create a {file_path}.error.txt to indicate the file error more clearly.
+                    logger.warning(
+                        'File: \'{file_path}\' with key: \'{key}\' was missing.\n'
```
nsprenkle, line 1016: Make log more explicit. What does "missing" mean? To make searching easier, propose making the first part a static string and logging out info afterwards. Proposed:

```suggestion
                        'File for submission could not be downloaded for ORA submission archive.\n'
```

Also include updated log (so we can see what those fields look like) in the PR description for verification. You may also find that some info is redundant and can be removed from the log.

openassessment/data.py
```diff
@@ -993,15 +999,37 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
...
+                        '\n\tType: {type}'
+                        '\n\tDescription: {description}'
+                        '\n\tFile Path: {file_path}'
+                        '\n\tFile Size: {size}'.format(**clone_file_data)
+                    )
+                else:
+                    clone_file_data['file_found'] = True
+                    zip_file.writestr(file_path, file_content)
 
+                csvwriter.writerow(clone_file_data)
```
nsprenkle, line 1032: I think there's a cleaner way to do this without duplicating or writing to the source object. Since you're only adding `file_found` key:

```suggestion
                csvwriter.writerow(**file_data, 'file_found': fileFound)
```

openassessment/tests/test_data.py
```diff
@@ -1319,6 +1319,62 @@ def test_create_zip_with_attachments(self):
 
         self.assertTrue(zip_file.read(os.path.join(COURSE_ID, 'downloads.csv')))
 
+    def test_create_zip_with_failed_attachments(self):
```
nsprenkle, line 1322: This can probably be decomposed into smaller tests: e.g.
- test a missing file raises an exception
- test a missing file does not appear in the output zip
- test an existing file does appear in the output zip
- test files appear with correct `file_found` tag in downloads.csv

nsprenkle (changes requested):
A handful of suggested improvements. Blocking issue is that the `csvwriter` needs updated headers. 

FWIW, this is why I'm requesting example output CSV in the PR description, because it shows modified behavior and would have caught the missing requirement.

openassessment/data.py
leangseu-edx: I updated the `cls.DOWNLOADS_CSV_HEADER` variables in earlier commit.

jansenk (commented):
Looks good! I finally found the code that does some CSV testing. https://github.com/edx/edx-platform/blob/3152c14192dcccac79b7b7a76eba91b380eba78a/lms/djangoapps/teams/tests/test_csv.py#L67 try taking a look at that

openassessment/data.py
jansenk: Will this work? I'm not super familiar with this function call, but it looks like you're mixing syntax

jansenk: specifically, wouldn't you need to wrap this in {}? or no?

leangseu-edx: Yes, the correct syntax is to wrap it in {}. so it would be `csvwriter.writerow({**file_data, 'file_found': False})`

openassessment/tests/test_data.py
leangseu-edx: changed to use spread instead of copy object from the original

openassessment/tests/test_data.py
leangseu-edx: I was just following the earlier example on `test_create_zip_with_attachments`. 

I know look closer to E2E but the test when focus entirely on one function which is `create_zip_with_attachments`

openassessment/data.py
nsprenkle: It's not showing in the diff, so I don't think the changes were committed correctly.

nsprenkle: LOL, JK. I'm still half asleep.

openassessment/data.py
```diff
@@ -964,6 +970,9 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
         Opens given stream as a zip file and writes into it all submission
         attachments and csv with list of all downloads.
 
+        Missing file will no be include in the zip. It will show as file_found=False
```
jansenk, line 973:
```suggestion
        Files that cannot be found in the backend will not be included in the zip. It will be listed as file_found=False
```

jnlapierre (approved):
LGTM

openassessment/data.py
```diff
@@ -993,14 +1002,33 @@ def create_zip_with_attachments(cls, file, course_id, submission_files_data):
...
+                try:
+                    file_content = (
+                        cls._download_file_by_key(key)
+                        if file_data['type'] == cls.ATTACHMENT
+                        else file_data['content']
+                    )
+                except FileMissingException:
+                    file_found = False
+                    # added a header to csv file to indicate that the file was found or not.
+                    # not sure if I should create a {file_path}.error.txt to indicate the file error more clearly.
```
jansenk, line 1016:
```suggestion
                    # TODO: (EDUCATOR-5777) should we create a {file_path}.error.txt to indicate the file error more clearly?
```

openassessment/tests/test_data.py
```diff
@@ -1317,7 +1317,110 @@ def test_create_zip_with_attachments(self):
             self.answer_text.encode('utf-8')
         )
 
-        self.assertTrue(zip_file.read(os.path.join(COURSE_ID, 'downloads.csv')))
```
jansenk, line 1320: oops, i don't think we want to delete this line from this other test, I think it's still worthwhile to assert that the file exists

openassessment/tests/test_data.py
```diff
@@ -1317,7 +1317,110 @@ def test_create_zip_with_attachments(self):
...
+        ) as download_mock:
+            OraDownloadData.create_zip_with_attachments(file, COURSE_ID, self.submission_files_data)
+
+        zip_file = zipfile.ZipFile(file)
+        # zip file contain csv file
+        csv_path = os.path.join(COURSE_ID, 'downloads.csv')
+        self.assertTrue(zip_file.read(csv_path))
+
+        with zip_file.open(csv_path) as csv_file:
+            # breakpoint()
```
jansenk, line 1336: remove

openassessment/tests/test_data.py
```diff
@@ -1317,7 +1317,110 @@ def test_create_zip_with_attachments(self):
...
+            csv_reader = csv.DictReader(TextIOWrapper(csv_file, 'utf-8'))
+            # csv_reader = csv.reader(csv_file)
```
jansenk, line 1338: remove

openassessment/tests/test_data.py
```diff
@@ -1317,7 +1317,110 @@ def test_create_zip_with_attachments(self):
             self.answer_text.encode('utf-8')
         )
 
-        self.assertTrue(zip_file.read(os.path.join(COURSE_ID, 'downloads.csv')))
+    def test_csv_file_for_create_zip_with_attachments(self):
```
jansenk, line 1320: looks great!

jansenk (changes requested):
The test changes looks great! just some small wording tweaks, and some comment cleanup in the tests

jansenk (approved):

PR #1576, opened 2021-01-22
[BD-05] [TNL-7310] [BB-3388] Pluggable ORA Submission Editor - TinyMCE

shimulch:
Depends on https://github.com/edx/edx-ora2/pull/1574

This PR adds TinyMCE v5 as the WYSIWYG editor for ORA Response.

**JIRA tickets**:
- https://openedx.atlassian.net/browse/TNL-7310
- https://tasks.opencraft.com/browse/BB-3388

**Discussions**: None

**Dependencies**: https://github.com/edx/edx-ora2/pull/1574

**Screenshots**: 
Screenshot from 2021-01-22 10-34-42

**Sandbox URL**: https://bb3388.sandbox.opencraft.hosting/

**Merge deadline**: None

**Testing instructions**:

1. Clone this PR on your local devstack. Install using make install-local-ora.
2. Create an ORA block. On the settings tab select ``WYSIWYG Editor`` from Response Editor Dropdown.
3. Publish ORA Block
4. Check that TinyMCE editor loaded in both LMS & Studio. 
5. Write on the editor, check ``Save Your Progress`` becomes enabled.
6. Click on ``Save Your Progress`` and reload the page. You should come back where you left off.
7. Check if the submission process working.

**Author notes and concerns**:

1. This PR includes TinyMCE v5 and all of the assets related to the editor.

**Reviewers**
- [ ] @viadanna
- [ ] @xitij2000
- [ ] @giovannicimolin
- [ ] edX reviewer[s] TBD

PR #1640, opened 2021-05-28
fix: textarea render multiple times

leangseu-edx:
Putting html within textarea made xblock re-render escape html on another div.

JIRA: [EDUCATOR-5806](https://openedx.atlassian.net/browse/EDUCATOR-5806)

**What changed?**

- Added if condition for rendering different response type
- The code change is mostly a revert of `pre-tinymce` feature [BD-05](https://github.com/edx/edx-ora2/pull/1576/files?file-filters%5B%5D=.html#diff-14dfd474b604ecd99ce7fbf79b1ab1ac003307e4a99972670b632b7bee911c87)
- Additional bug fix: WYSIWYG editor couldn't render image/link properly

**Developer Checklist**

**Testing Instructions**

- Normal TextArea
- Open submitted ORA

**NOTE**: To reproduce, you need to grade the response. Try refresh a couple times if you do not see the bugs.
Before:
Screen Shot 2021-06-01 at 1 34 19 PM

After:
Screen Shot 2021-06-01 at 1 35 45 PM

- WYSIWYG editor
  - Open submitted ORA
  - The response is properly formatted
Before:
Screen Shot 2021-06-01 at 1 30 07 PM 

After:
Screen Shot 2021-06-01 at 1 30 20 PM

[**Use this if needed**](https://studio.stage.edx.org/container/block-v1:edx+TS101+2020_T1+type@vertical+block@336977df509b4580bab132749a04a8a5)
NOTE: I'm not sure if I can write unit test for this.
 
**Reviewer Checklist**

FYI: @edx/masters-devs-gta


jansenk:
I'm trying to test this change but I'm unable to reproduce the original bug. What should my text response be?

nsprenkle (changes requested):
This appears to be introducing a separate issue: The response in **Assess Your Response** section is editable, but should not be.

Reproduce by creating an ORA with WYSIWYG editor and Self Assessment step

jansenk:
I've been poking at this all day, and I don't know if this is a different bug, but the way that I reproduced turned out to be something that triggered when the `Final Grade` section loaded??? Somehow that section loading inserts another section? I think there's a much nastier bug here than we're seeing.

leangseu-edx:
> This appears to be introducing a separate issue: The response in **Assess Your Response** section is editable, but should not be.
> 
> Reproduce by creating an ORA with WYSIWYG editor and Self Assessment step

Thought it was a separate problem because I could not reproduce it on stage. After carefully tested on released ORA2, it is appear to be the side effect.

leangseu-edx:
> I've been poking at this all day, and I don't know if this is a different bug, but the way that I reproduced turned out to be something that triggered when the `Final Grade` section loaded??? Somehow that section loading inserts another section? I think there's a much nastier bug here than we're seeing.

I also observe the behavior. It started with the change to support [tiny-mce](https://github.com/edx/edx-ora2/pull/1576/files?file-filters%5B%5D=.html#diff-14dfd474b604ecd99ce7fbf79b1ab1ac003307e4a99972670b632b7bee911c87).

From my understanding, it is bad to put`html` tag inside `textarea`. It doesn't render properly. I tried to fix it from the other angle where I think `tiny-mce` can do better. However, I cannot find the specific code that render the text. In the end, I approach it with this solution instead and I think it is equally good.

nsprenkle:
> I've been poking at this all day, and I don't know if this is a different bug, but the way that I reproduced turned out to be something that triggered when the `Final Grade` section loaded??? Somehow that section loading inserts another section? I think there's a much nastier bug here than we're seeing.

@jansenk, that was the originally reported bug and this ticket should aim to fix it.

nsprenkle:
> > I've been poking at this all day, and I don't know if this is a different bug, but the way that I reproduced turned out to be something that triggered when the `Final Grade` section loaded??? Somehow that section loading inserts another section? I think there's a much nastier bug here than we're seeing.
> 
> I also observe the behavior. It started with the change to support [tiny-mce](https://github.com/edx/edx-ora2/pull/1576/files?file-filters%5B%5D=.html#diff-14dfd474b604ecd99ce7fbf79b1ab1ac003307e4a99972670b632b7bee911c87).
> 
> From my understanding, it is bad to put`html` tag inside `textarea`. It doesn't render properly. I tried to fix it from the other angle where I think `tiny-mce` can do better. However, I cannot find the specific code that render the text. In the end, I approach it with this solution instead and I think it is equally good.

@leangseu-edx, I would look back into the `tiny-mce` setup. I think the `html` inside the `textarea` is actually part of setting up the `tiny-mce` editor (in that `tiny-mce` is actual an html renderer but it needs a home for the source markup before the rendering step occurs).

leangseu-edx:
> > > I've been poking at this all day, and I don't know if this is a different bug, but the way that I reproduced turned out to be something that triggered when the `Final Grade` section loaded??? Somehow that section loading inserts another section? I think there's a much nastier bug here than we're seeing.
> > 
> > 
> > I also observe the behavior. It started with the change to support [tiny-mce](https://github.com/edx/edx-ora2/pull/1576/files?file-filters%5B%5D=.html#diff-14dfd474b604ecd99ce7fbf79b1ab1ac003307e4a99972670b632b7bee911c87).
> > From my understanding, it is bad to put`html` tag inside `textarea`. It doesn't render properly. I tried to fix it from the other angle where I think `tiny-mce` can do better. However, I cannot find the specific code that render the text. In the end, I approach it with this solution instead and I think it is equally good.
> 
> @leangseu-edx, I would look back into the `tiny-mce` setup. I think the `html` inside the `textarea` is actually part of setting up the `tiny-mce` editor (in that `tiny-mce` is actual an html renderer but it needs a home for the source markup before the rendering step occurs).

After looking through it, the current fix is probably more efficient. `openassessment/templates/openassessmentblock/oa_submission_answer.html` has to aware of the text editor it is using. Because text escaping have to behave differently for between WYSIWYG and normal text area. The bug has always been there if you try to submit an image tag using WYSIWYG editor.

I also look into the multiple time rendering issue. Apparently, all of the `mixin` doesn't aware of its own rendering block (not sure if this is the correct term). It aware of the xblock as a whole. The reuse of `openassessment/templates/openassessmentblock/oa_submission_answer.html` confuse the dom selector and cause multiple rendering within xblock.

openassessment/xblock/openassessmentblock.py
```diff
@@ -1008,6 +1008,8 @@ def render_assessment(self, path, context_dict=None):
         if not context_dict:
             context_dict = {}
 
+        context_dict['text_response_editor'] = self.text_response_editor
```
leangseu-edx, line 1011: I think this is pretty high level on the change but it would have been the same if I were to go through all of the assessment and pass this as context.

openassessment/xblock/static/dist/95ec738c0b7faac5b5c9126794446bbd.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="36" height="36" viewBox="0 0 36 36" fill="none" xmlns="http://www.w3.org/2000/svg">
```
nsprenkle, line 1: What is this file?

nsprenkle (approved):
Looks good 👍 Good work!

PR #1655, opened 2021-06-21
Jkantor/pylint

jansenk:
**TL;DR -** Pylint was silently failing

JIRA: [AU-26](https://openedx.atlassian.net/browse/AU-26)

**What changed?**

- this PR (https://github.com/edx/edx-ora2/pull/1594) added the --django-settings-module flag to the pylint call, but pylint fails because `edx-ora2.settings` isn't a python module. pylint would explode and write nothing to the expected output file, and the post-pylint output file processing script we had interpreted a blank file as a success with 0 failures.
- removing the `edx-ora2` from the module name makes it point to the settings folder, which is a valid python module
- also added code to `run-pylint.sh` to check if pylint exploded, and if so, exit with an error code so travis knows something went wrong.
- I also added the acceptance tests to pylint ignore, because they've been nonfunctional, untouched, and unused for a while now. Doing so resulted in big changes to `pylintrc`, but it's mostly edx-lint reorganizing things alphabetically.

- **Note**: The majority of the lint errors are something i believe is a newly enforced rule in edx-lint: `logging-format-interpolation`. This rule wants you to use old-style string formatting (`'format var = %s' % param`) and pass the params as args to the logging function rather than string.format(). I'm not entirely sure why, but the logging documentation insists that this is best practice. Something about lazy evaluation I think. 

**Developer Checklist**

**Testing Instructions**

Look at the quality test for this PR, and make sure the call to pylint doesn't explode.
You can also checkout this branch and run `make quality` in your virtualenv and make sure pylint doesn't fail to run

**Reviewer Checklist**

FYI: @edx/masters-devs-gta


scripts/run-pylint.sh
```diff
@@ -6,5 +6,11 @@ mkdir -p test/logs
 PYLINT_VIOLATIONS=test/logs/pylint.txt
 touch $PYLINT_VIOLATIONS
 
-pylint --django-settings-module=edx-ora2.settings.test --rcfile=pylintrc openassessment test --msg-template='"{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}"'> $PYLINT_VIOLATIONS
+export DJANGO_SETTINGS_MODULE=settings.test
+pylint --rcfile=pylintrc openassessment test --msg-template='"{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}"'> $PYLINT_VIOLATIONS
```
jansenk, line 10: changing the `--django-settings-module=` arg worked outside of the container, but not inside the container? I don't know. don't care. this works.

openassessment/assessment/api/student_training.py
```diff
@@ -389,20 +389,15 @@ def get_training_example(submission_uuid, rubric, examples):
...
         raise StudentTrainingRequestError(msg) from ex
     except DatabaseError as ex:
-        msg = (
-            u"Could not retrieve a training example for the learner with submission UUID {}"
-        ).format(submission_uuid)
-        logger.exception(msg)
-        raise StudentTrainingInternalError(msg) from ex
+        msg = "Could not retrieve a training example for the learner with submission UUID %s"
+        logger.exception(msg, submission_uuid)
+        raise StudentTrainingInternalError(msg % submission_uuid) from ex
```
nsprenkle, line 400: TIL what the % does here 🆒

scripts/run-pylint.sh
nsprenkle: Does this change anything about how/where we run the script? Does it need doc/README changes?

nsprenkle (approved):
Visual review looks good. I don't know if there's really a smoke test to run... I say :shipit:

scripts/run-pylint.sh
jansenk: Nope. This is just called straight, no args, from the makefile.

openassessment/assessment/api/student_training.py
jansenk: Yeah, it's how older python did string formatting before fstrings or string.format. It's kinda lame that logging still wants you to do it this way tbh

jansenk (commented):
i didnt want to leave a review i just wanted to comment :(

nsprenkle (approved):
Approved again!

PR #1661, opened 2021-07-08
chore(add logging around ORA submission ZIP)

jansenk:
**TL;DR -** Added lots of temporary logging around the ORA submission ZIP files to resolve "stuck" processes 

JIRA: [EDUCATOR-5878](https://openedx.atlassian.net/browse/EDUCATOR-5878)

Some will be removed in https://openedx.atlassian.net/browse/EDUCATOR-5886

FYI: @edx/masters-devs-gta

### DIFF
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -28,12 +28,6 @@ from openassessment.workflow.models import AssessmentWorkflow, TeamAssessmentWor
 logger = logging.getLogger(__name__)
 
 
-def _log_5862(msg, *args, **kwargs):
-    """ Helper method for debugging EDUCATOR-5862"""
-    msg = '[EDUCATOR-5862]: ' + msg
-    logger.info(msg, *args, **kwargs)
-
-
 def _usernames_enabled():
     """
     Checks if toggle for deanonymized usernames in report enabled.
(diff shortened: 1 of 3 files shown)


mattcarter (approved):
LGTM

openassessment/data.py
```diff
@@ -1168,8 +1169,26 @@ def create_zip_with_attachments(cls, file, submission_files_data):
         csvwriter = csv.DictWriter(csv_output_buffer, cls.SUBMISSIONS_CSV_HEADER, extrasaction='ignore')
         csvwriter.writeheader()
 
+        logger.info("Beginning writing to ZIP file")
```
nsprenkle, line 1172: Does the logger give contextual information like [this is coming from ORA attempting to create the submission archive]? I could see this message being a bit vague without it.

jansenk: Yeah, I agree that this isn't particularly useful out of context, but in the specific use case of  "hey, does this method get called? Where the heck does execution end?" it has a purpose

jansenk: This task is called infrequently enough that there's very very little chance of overlap

jansenk: I've got https://openedx.atlassian.net/browse/EDUCATOR-5886 to address post-diagnosis cleanup

PR #1673, opened 2021-08-03
fix(rubric reuseplaceholder message)

jansenk:
**TL;DR -** I changed a variable name in one place and forgot to change it in another

JIRA: [AU-22](https://openedx.atlassian.net/browse/AU-22)

**What changed?**
-  whoops

**Testing Instructions**

- Install ORA, enable flag (openresponseassessment.rubric_reuse) go to a course with other ORAs and add another ORA.
- Before the change, the rubric reuse will work correctly, but the placeholder on the input says that there aren't any other ORAs in the course. You also won't be able to click on the input (but the button will still show the dropdown)
- After the change, it shouldn't show that message, and you should be able to type
- But if you go to a blank course and create an ORA, it should say that there are no ORAs in the course
**Reviewer Checklist**

FYI: @edx/masters-devs-gta


jansenk:
Another issue, closing until i figure it out

jansenk:
This issue was that I originally was passing a list, so I could check {% if not ... but it became a string, so that check was no longer working. This allows me to pass a list again

nsprenkle (approved):
Testing looks good

PR #1677, opened 2021-08-10
Hide step actions for response if in studio

muselesscreator:
**TL;DR -** 
Added a has_real_user prop based on whether the runtime is bound to a real user.

This is now linked such that nothing below the response shows up if it is true.

JIRA: [AU-58](https://openedx.atlassian.net/browse/AU-58)

**Developer Checklist**

**Testing Instructions**

View an ORA in studio and verify the Save, file upload, and actions (submit button) don't show up
Verify that they do for the LMS view.

**Reviewer Checklist**

FYI: @edx/masters-devs-gta

### DIFF
--- a/openassessment/xblock/openassessmentblock.py
+++ b/openassessment/xblock/openassessmentblock.py
@@ -751,6 +751,18 @@ class OpenAssessmentBlock(MessageMixin,
         # student ID, so we can't rely on that)
         return self.scope_ids.user_id is None
 
+    @property
+    def in_studio(self):
+        """
+        Checks whether we are in Studio view
+
+        Returns:
+            bool
+        """
+        if hasattr(self, 'xmodule_runtime'):
+            return self.xmodule_runtime.get_real_user is None  # pylint: disable=no-member
+        return False
+
     def _create_ui_models(self):
         """Combine UI attributes and XBlock configuration into a UI model.
 
--- a/openassessment/xblock/openassessmentblock.py
+++ b/openassessment/xblock/openassessmentblock.py
@@ -752,15 +752,15 @@ class OpenAssessmentBlock(MessageMixin,
         return self.scope_ids.user_id is None
 
     @property
-    def in_studio(self):
+    def has_real_user(self):
         """
-        Checks whether we are in Studio view
+        Checks whether the runtime is tied to a real user
 
         Returns:
             bool
         """
         if hasattr(self, 'xmodule_runtime'):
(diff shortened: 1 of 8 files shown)


openassessment/xblock/openassessmentblock.py
```diff
@@ -746,10 +746,9 @@ def in_studio_preview(self):
             bool
 
         """
-        # When we're running in Studio Preview mode, the XBlock won't provide us with a user ID.
-        # (Note that `self.xmodule_runtime` will still provide an anonymous
-        # student ID, so we can't rely on that)
-        return self.scope_ids.user_id is None
+        # When we're running in Studio Preview mode, the XBlock won't
+        # be tied to a real user
+        return self.xmodule_runtime.get_real_user is None
```
nsprenkle, line 751: Follow the pattern from other instances of calling the `xmodule_runtime`
```suggestion
        if hasattr(self, 'xmodule_runtime'):
            return self.xmodule_runtime.get_real_user is None  # pylint: disable=no-member
        return False
```

nsprenkle (commented):
Looks good except for failing quality test. Have added suggested fix following patterns throughout the repo. Will test locally.

nsprenkle (approved):
I've never been so excited to _not_ see something on the page.

openassessment/xblock/openassessmentblock.py
```diff
@@ -751,6 +751,18 @@ def in_studio_preview(self):
         # student ID, so we can't rely on that)
         return self.scope_ids.user_id is None
 
+    @property
+    def in_studio(self):
```
jansenk, line 755: I feel uneasy about this for some reason, and I think we should instead be explicit and just call this "has_no_real_user" or something like that. It's a blurry, kinda backwards way of checking if we're in studio. It works, but I think we'd be safer just calling this what it is, and plus, it makes sense down the line too. Why do we care about not submitting in studio? Because there's no real user. Cut out the middle step and hide the button if there's no user. Does that make sense? I'm kinda rambling.

jansenk:
Does this need an associated template change? Was that already done?

nsprenkle (dismissed):
RE @jansenk, looks like template updates got lost in your recent changes. Need to fix that.

jansenk (approved):
LGTM

leangseu-edx (approved):
LGTM 👍

mattcarter (approved):
LGTM 👍

PR #1688, opened 2021-08-26
wip validate on tab click

jansenk:
**TL;DR -** Re-validate validation errors on tab change

JIRA: [AU-272](https://openedx.atlassian.net/browse/AU-272)

**What changed?**

- Refactored the validation flow into a helper method which is then called from save and from the tab change callback 
- added actuveTab getter and setter for better test readability
- Changed some validation tests to call the validation function rather than the save function
- added a test for tab change validation

**Developer Checklist**

**Testing Instructions**

- Make an ORA. Open the edit modal
- make a field invalid (ie change a date/time to a nonsense string of letters) 
- attempt to save. You should see the validation error
- change the value to something valid, and you should be able to save.
- Change it back to something invalid, and then change tabs. The tab should be marked and you should see the alert banner.
- go back, fix it, and change tabs again. the validation should be removed.

**Reviewer Checklist**

FYI: @edx/masters-devs-gta


openassessment/xblock/static/js/src/studio/oa_edit.js
```diff
@@ -38,7 +38,7 @@ export class StudioView {
     this.fixModalHeight();
 
     // Initializes the tabbing functionality and activates the last used.
-    this.initializeTabs();
+    this.tabsWidget = this.initializeTabs();
```
leangseu-edx, line 41: This doesn't return anything. Beside it is probably better just to assign attribute to some value within that function.

PR #1691, opened 2021-09-03
Submission lock API

nsprenkle:
**TL;DR -** Add an API for CRUD actions on submission locks.

JIRA: [AU-54](https://openedx.atlassian.net/browse/AU-54)
Linked PR: https://github.com/edx/edx-platform/pull/28816

**What changed?**

- Added new `SubmissionGradingLock ` model to support locking a submission
- Added `StaffGraderMixin` to add XBlock JSON handler methods:
    - `check_submission_lock`
    - `claim_submission_lock`
    - `delete_submission_lock`

**API Documentation**
All endpoints follow the structure (due to them being `XBlock.json_handler`s)

```
POST http://{{lms_url}}/courses/{{course_id}}/xblock/{{block-locator}}/handler/{{name}}

BODY {
    "submission_id": <submission_uuid>
}

RESPONSE {
    "submission_uuid": <submisison_uuid>,
    "owner_id": <lock_owner_anon_id>,
    "created_at": <lock_creation_timestamp>
}

ERRORS
- 400 - missing submission_id
- 403 - lock contested OR insufficient permissions
- 405 - not logged in
```

**Developer Checklist**

## Testing Instructions

**Test Setup**

The best way to test is using a tool like Postman to exercise the APIs. First, since we're plugged in to LMS CSRF and Auth framework:

1. Capture login session (using Postman/similar):
    1. Use cookie synchronization to copy over cookies from an active login session.
    2. Copy `csrftoken` value from cookie, add it to a new `X-CSRFToken` header.
2. Find a submission ID to work with for your course (for example, by looking at [Django Admin > Submissions](http://localhost:18000/admin/submissions/submission/)

Verify the following cases:

**Access tests** - These are built-in by `XBlock.json_handler` but might be good to verify anyways.
1. A non-logged in user gets a `405` (Unauthenticated) for each request.
2. A non-staff user for the given course receives a `403` (Forbidden) for each request.

**Method tests**
As a user with staff permissions:
1. `get_submission_lock` returns submission lock info or Empty `{}`
2. `claim_submission_lock` for a submission without a lock creates a new lock and returns lock info.
3. `clear_submission_lock` for a submission with (or without) a lock clears the lock and returns `{}`

Next, we want to verify only the owner of current locks can renew/clear them. As (probably a different) user with staff permissions:
1. `get_submission_lock` returns submission lock info on the other user's lock.
2. `claim_submission_lock` for a lock owned by another user returns a `403` with response `{ "Error": "Submission already locked"}`
2. `clear_submission_lock` for a lock owned by another user returns a `403` with response `{ "Error": "Unable to clear submission lock"}`

**Reviewer Checklist**

FYI: @edx/masters-devs-gta


openassessment/staff_grader/api.py
```diff
@@ -0,0 +1,117 @@
...
+            ).values(
+                "anonymoususerid__anonymous_user_id"
+            )
+        ).get()
+        return user_anon_id["anonymoususerid__anonymous_user_id"]
+    except (ObjectDoesNotExist, MultipleObjectsReturned):
+        return None
+
+
+def has_access(user, course_id, access_level="staff"):
```
jansenk, line 91: This should be a decorator! I know we do similar things elsewhere. Also, what is this logic based on? Do we not have something that already does this?

openassessment/staff_grader/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import url
+
+from openassessment.staff_grader.api import locks_view
+
+urlpatterns = [
+    # Enhanced Staff Grader (ESG) locking
+    url(r'submission/(?P<submission_uuid>.+)/lock$', locks_view, name='openassessment-submission-lock')
```
jansenk, line 7: I gotta say I strongly oppose this name. We have a service named submissions that is already a bit too coupled (imo) to ORA and this would be very confusing. Also, should the course also be a part of the URL?

jansenk (commented):
Soome thoughts

openassessment/assessment/models/staff.py
```diff
@@ -54,6 +70,64 @@ def is_cancelled(self):
         """
         return bool(self.cancelled_at)
 
+    @property
+    def is_locked(self):
```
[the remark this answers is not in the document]
nsprenkle: Lock is the term we've been using when talking about the feature, but I'm okay changing it if we have a better term.

openassessment/assessment/models/staff.py
```diff
@@ -54,6 +70,64 @@ def is_cancelled(self):
...
+            self.scorer_id = scorer_id
+            self.grading_started_at = now()
+            self.save()
+        except DatabaseError as ex:
+            error_message = (f'An internal error occurred trying to lock submission: {self.submission_uuid}')
+            logger.exception(error_message)
+            raise StaffAssessmentInternalError(error_message) from ex
+        return True
+
+    def clear_lock(self, scorer_id):
```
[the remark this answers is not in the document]
nsprenkle: This is an unclarified part of the feature: Can an active lock be cleared by someone who doesn't own the lock? My assumption was only the owner of the lock should be able to clear it until after it expires. Will ask @sapanathomas523 to clarify.

openassessment/staff_grader/api.py
nsprenkle: ^ This is built from similar decorators in LMS which we don't have access to outside of XBlock context and which (I don't think) we have the ability to import.

openassessment/assessment/models/staff.py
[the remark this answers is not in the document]
jansenk: I agree that from a user perspective maybe we want that, but I think we should be able to programmatically clear this without needing to know the current user's id

openassessment/staff_grader/urls.py
nsprenkle: LMS URL routing includes the course so the whole URL path is:
```
{lms}/courses/{course}/staff_grader/submission/{submission_id}/lock
```

I went with this naming because we're trying to lock a submission (well, really a staff workflow, but the more user-accessible piece is the submission) and it's under the `staff_grader` section to disambiguate from regular submission API(s)

jansenk: I still think it should be changed to avoid confusion. Would you be opposed to `workflow`?

openassessment/assessment/models/staff.py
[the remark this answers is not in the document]
nsprenkle: @jansenk , okay, so maybe that happens from the API standpoint and not the model standpoint. Good point.

openassessment/staff_grader/urls.py
muselesscreator: fwiw, from the staff_grader perspective, this is working on the submission... though I understand that terminology is muddy.  But having `workflow` tied to the submission id in this call would seem more odd to me

openassessment/staffgrader/tests/test_submission_lock_model.py
```diff
@@ -0,0 +1,74 @@
...
+            created_at=datetime.now(tz=timezone.utc)
+        )
+
+    def test_lock_active(self):
+        # A lock created within the TIMEOUT is considered active
+        assert self.existing_submission_lock.is_active is True
+
+    def test_lock_inactive(self):
+        # A lock created and left for more than TIMEOUT is considered inactive
+        self.existing_submission_lock.created_at = self.existing_submission_lock.created_at - timedelta(hours=12)
```
leangseu-edx, line 35: why not use `TIMEOUT` variable?

leangseu-edx (approved):
LGTM ✅

openassessment/staffgrader/tests/test_submission_lock_model.py
nsprenkle: Valid, I just needed "time in the future past timeout" but I can make that clearer

openassessment/staffgrader/models/submission_lock.py
```diff
@@ -0,0 +1,83 @@
...
+        # Get the current lock, assuming it exists
+        current_lock = cls.get_submission_lock(submission_uuid)
+        if not current_lock:
+            return
+
+        # Only the owner can clear the lock
+        if current_lock.owner_id != user_id:
+            raise SubmissionLockContestedError(_("Unable to clear submission lock"))
+
+        current_lock.delete()
```
jansenk, line 83: mmm, I think we shouldn't delete these models. I thought the reason this is a new model was for traceablility? If we aren't keeping these around, there's really no point of adding this model when we already have the scorer_id and graded_started_at on the staff workflow

openassessment/staffgrader/models/submission_lock.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class SubmissionGradingLock(models.Model):
+    """
+    Internal model for locking a submission for exclusive grading
+    """
+    TIMEOUT = StaffWorkflow.TIME_LIMIT
+
+    # NOTE - submission_uuid can refer to either the team or individual submission
+    submission_uuid = models.CharField(max_length=128, db_index=True, unique=True)
```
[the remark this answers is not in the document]
nsprenkle: I'd like to delay this until a separate PR until there's demonstrated contract for `StaffWorkflow`

openassessment/staffgrader/models/submission_lock.py
jansenk: I don't really follow what you mean by that. Do we want to keep a history of who locked what or no? What benefit does this model give us over what we have already then?

openassessment/staffgrader/models/submission_lock.py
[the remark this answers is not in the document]
jansenk: The "SubmissionList" view is really a StaffWorkflow view (all the data that we're looking for is on StaffWorkflow).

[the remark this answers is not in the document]
jansenk: The query currently doesn't specifically /need/ this FK but it's probably worth the performance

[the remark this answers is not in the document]
jansenk: I actually take that back, it would be a huge benefit. Rather than having to do complicated subqueries, I can just prefetch_related and hand it to the serializer

openassessment/staffgrader/models/submission_lock.py
nsprenkle: We don't really need a history. What this buys us over the previous system is allowing us to add locks for regrading. Otherwise, we'd have to manage timestamps on completed staff workflows which would be gross. This also allows us future ability to expand how we do lock management, like tying a lock to a session instead of a user ID.

openassessment/staffgrader/models/submission_lock.py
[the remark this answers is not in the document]
nsprenkle: If we add StaffWorkflow foreign key, any other constraints/info on how it's used?

openassessment/staffgrader/models/submission_lock.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        # See if there's already a lock
+        current_lock = cls.get_submission_lock(submission_uuid)
+
+        # If there's already an active lock, raise an error
+        if current_lock and current_lock.is_active:
+            raise SubmissionLockContestedError(_("Submission already locked"))
+
+        # Otherwise, create a new lock
+        new_lock, created = cls.objects.update_or_create(  # pylint: disable=unused-variable
```
jansenk, line 60: We'll need to do a StaffWorkflow lookup here .... is it worth it? Everything else is already so disconnected. Maybe just drop the Foreign Key field, idk

jansenk:
We can either drop the ForeignKey or do a lookup on create, but we gotta do one or the other before merging.

nsprenkle:
> We can either drop the ForeignKey or do a lookup on create, but we gotta do one or the other before merging.

@jansenk, since I don't understand the use-case as well, can you propose one?

nsprenkle:
> I'd like if you could add a test for the case when there is an old, expired lock that exists in the DB and we request to lock the submission. I'm unsure how that will work with the unique constraint on submission_uuid? I am also willing to write the test if you're busy

Good point. Will add.

jansenk:
Doesn't have to be done here but should probably be done, I threw https://openedx.atlassian.net/browse/AU-345 into the backlog for tests of calling the handlers as handlers

openassessment/staffgrader/models/submission_lock.py
```diff
@@ -0,0 +1,87 @@
...
+            raise SubmissionLockContestedError(_("Submission already locked"))
+
+        # Otherwise, delete the lock. This is needed so we don't violate the unique submisison_uuid constraint
+        elif current_lock:
+            current_lock.delete()
+
+        # Create a new lock
+        new_lock = cls.objects.create(
+            owner_id=user_id,
+            created_at=now(),
```
jansenk, line 65: Rather than manually specifying, do we want to just set `default=datetime.now`on the field?

nsprenkle:
Merging into feature branch so I'm going to skip stuff like translations and version bumping.

PR #1713, opened 2021-10-21
fix: change how submission zip looks up block children

jansenk:
**TL;DR -** change how submission zip looks up block children

JIRA: [AU-374](https://openedx.atlassian.net/browse/AU-374)

**What changed?**

- There was a bug where the way that we look up child blocks in the ZIP download was not working
- It was a questionable approach anyways, so this is a change to the method that definitely works.
- I also added some additional logging. (My previous logging let me identify the issue way faster. Thanks, past me!)

**Testing Instructions**

1.  Install a local ORA. 
2. Go to a course with ORAs and submissions. Go to instructor dashboard -> data download and click "Generate submissions archive"
3.  Wait for the task to complete. You will see a link, but we cannot click it in devstack.
4. To check the file, run the following command: `docker cp edx.devstack.lms:/tmp/edx-s3/grades lms_tmp` This will copy your lms devstack """uploads""" directory to a local folder called lms_tmp. In there will be some folders. You might have to root around a bit but you'll find a zip file with the same name as your submission zip. Unzip it, and look inside.
5. There should be a csv and a folder called "removed from course"
6. Now, switch to this branch and allow the LMS to restart
7. Delete the lms_tmp folder and run steps 2-4 again.
8. This time, the folder should contain folders that somewhat map to course structure. To be honest I don't totally understand how it's laid out but it won't all be "removed from course"

**Reviewer Checklist**

FYI: @edx/masters-devs-gta

### DIFF
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -964,14 +964,15 @@ class OraDownloadData:
         where key is a string representation of ORA's usage key, and value is a dictionary with
         all information needed to build the submission file path.
         """
-
         blocks = _get_course_blocks(course_id)
+        logger.info("[%s] _get_course_blocks returned %d blocks", course_id, len(blocks))
 
         path_info = {}
 
         def children(usage_key, condition=None):
             # pylint: disable=filter-builtin-not-iterating
-            filtered = filter(condition, blocks.get_xblock_field(usage_key, 'children') or [])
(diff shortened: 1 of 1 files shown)


leangseu-edx (approved):
LGTM 👍

PR #1759, opened 2021-11-29
Jkantor/add file size

jansenk:
**TL;DR -** Add file size to the  ESG file info response 

JIRA: [AU-381](https://openedx.atlassian.net/browse/AU-381)

FYI: @edx/masters-devs-gta

### DIFF
--- a/openassessment/xblock/submission_mixin.py
+++ b/openassessment/xblock/submission_mixin.py
@@ -614,9 +614,7 @@ class SubmissionMixin:
                 with related content
 
         Returns:
-            List with URLs to related content. If there is no content related to this
-            key, or if there is no key for the submission, returns an empty
-            list.
+            List of FileDescriptor dicts for each file associated with the submission
 
         """
         urls = []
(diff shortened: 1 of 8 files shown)


openassessment/xblock/submission_mixin.py
```diff
@@ -671,6 +672,7 @@ def get_files_info_from_user_state(self, username):
                             download_url=download_url,
                             description=description,
                             name=file_name,
+                            size=None,
```
mattcarter, line 675: Should we update the docstring to match?

openassessment/xblock/submission_mixin.py
```diff
@@ -723,6 +725,7 @@ def get_all_upload_urls_for_user(self, username_or_email):
                         download_url=download_url,
                         description='',
                         name='',
+                        size=None,
```
mattcarter, line 728: Should we update the docstring to match?

openassessment/xblock/submission_mixin.py
jansenk: Yeah, this is actually super outdated

nsprenkle (approved):
Vaguely worried by how many places this touches. Looks like you've caught all them but it would be worth some manual smoke testing if it hasn't been done already.

leangseu-edx (approved):
LGTM 👍

PR #1770, opened 2021-12-16
refactor: return lock info with lock contest errors

nsprenkle:
**TL;DR -** change lock contest error shape to return lock info

Paired with PR TBD in `edx-platform` to ingest new error data.

JIRA: [AU-445](https://openedx.atlassian.net/browse/AU-445) 

**What changed?**

- Lock contestation error used to return static error code `403 { "error": "ERR_LOCK_CONTESTED" }`
- Now returns `403 { "error": { (lock_info) }` so that we can update frontend with lock status on error

Low risk in confusion since there is really only one error type we expect to see (outside of network/server issues) which is lock contestation, so the error code doesn't really disambiguate anything.

**Testing Instructions**

1. Create a lock with one user using `claim_submission_lock`
2. Attempt to `delete_submission_lock` or `claim_submission_lock` as another user
3. Verify error information matches the pattern below:

```
{ 
    "error": {
        "submission_uuid": <submission_uuid>,
        "owner_id": <owner_anon_user_id>,
        "created_at": <timestamp>,
        "lock_status": "locked",
    }
}
```

**Reviewer Checklist**

FYI: @edx/masters-devs-gta

### DIFF
--- a/openassessment/staffgrader/errors/__init__.py
+++ b/openassessment/staffgrader/errors/__init__.py
@@ -1,0 +1,1 @@
+""" Error classes, codes, etc. for Enhanced Staff Grader (ESG) """
--- a/openassessment/staffgrader/errors/submission_lock.py
+++ b/openassessment/staffgrader/errors/submission_lock.py
@@ -11,3 +11,6 @@ class SubmissionLockContestedError(Exception):
 
     def get_error_code(self):
         return self.error_code
+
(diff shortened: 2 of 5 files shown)


nsprenkle:
Updated approach to catch and re-query lock status from BFF instead of trying to include that context as part of the error response.

PR #1787, opened 2022-01-21
fix: input criteria name -> label, allow null/missing feedback

jansenk:
Management command fixes:
- change input key for "selected" criteria from name -> label, since that's what we're actually specifying
- change lookup of criteria to map criteria and option labels to name
- allow null/missing feedback for criteria
FYI: @edx/masters-devs-gta


mattcarter (approved):
LGTM

PR #1788, opened 2022-01-21
refactor: Django storage returns full URL

nsprenkle:
**TL;DR -** to make file uploads work for ESG in devstack, update the `django_storage` file backend to return fully-qualified file download URLs instead of the relative-to-LMS URLs previously returned.

- [x] TODO - test updates

**Testing Instructions**

Do it work tho?

**Reviewer Checklist**

FYI: @edx/masters-devs-gta

### DIFF
--- a/openassessment/fileupload/backends/django_storage.py
+++ b/openassessment/fileupload/backends/django_storage.py
@@ -33,7 +33,7 @@ class Backend(BaseBackend):
             storage_path = default_storage.url(path)
 
             # Return a fully-qualified URL
-            lms_url = getattr(settings, 'LMS_ROOT_URL')
+            lms_url = settings.LMS_ROOT_URL
             return urljoin(lms_url, storage_path)
         return None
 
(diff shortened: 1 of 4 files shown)


mattcarter (approved):
lgtm

jansenk (approved):
This looks like an reasonable change. I was worried at first but I don't think this will break anything. (once the tests are passing)

openassessment/fileupload/backends/django_storage.py
```diff
@@ -28,7 +30,11 @@ def get_download_url(self, key):
         """
         path = self._get_file_path(key)
         if default_storage.exists(path):
-            return default_storage.url(path)
+            storage_path = default_storage.url(path)
+
+            # Return a fully-qualified URL
+            lms_url = settings.LMS_ROOT_URL
```
jansenk, line 36: would it complain if you had `getattr(settings, 'LMS_ROOT_URL', '')`? (the added default) That feels like a better way and I'd argue we should just add a pylint ignore if that makes it complain.

nsprenkle: We use all 3 paradigms (`settings.LMS_ROOT_URL`, `getattr` with and w/out default) and it complains about the `getattr` implementations. Have pylint ignored.

PR #1791, opened 2022-01-24
refactor: make the list endpoint use a serializer

jansenk:
**TL;DR -** make the list endpoint use a serializer rather than manual serialization

JIRA: [AU-488](https://openedx.atlassian.net/browse/AU-488)

**What changed?**

- Pulled the lookups out of `staff_workflows_to_api_format` into new function,` _get_list_workflows_serializer_context`
- Then, rather than manually serializing, pass workflows to a serializer along with the context looked up in `_get_list_workflows_serializer_context`

**Reviewer Checklist**

FYI: @edx/masters-devs-gta

### DIFF
--- a/openassessment/staffgrader/serializers/submission_list.py
+++ b/openassessment/staffgrader/serializers/submission_list.py
@@ -1,0 +1,100 @@
+"""
+Serializers for the Submission List endpoint
+"""
+from rest_framework import serializers
+from openassessment.assessment.models.staff import StaffWorkflow
+
+
+class MissingContextException(Exception):
+    pass
+
+
+# pylint: disable=abstract-method
+class SubmissionListScoreSerializer(serializers.Serializer):
+    pointsEarned = serializers.IntegerField(source='points_earned')
+    pointsPossible = serializers.IntegerField(source='points_possible')
+
+
+class SubmissionListSerializer(serializers.ModelSerializer):
+    """
+    Serialized info about an item returned from the submission list endpoint
+    """
+    class Meta:
+        model = StaffWorkflow
+        fields = [
+            'submissionUuid',
+            'dateSubmitted',
+            'dateGraded',
+            'gradingStatus',
+            'lockStatus',
+            'gradedBy',
+            'username',
+            'score'
+        ]
+        read_only_fields = fields
+
+    requires_context = True
+
+    CONTEXT_SUB_TO_ANON_ID = 'submission_uuid_to_student_id'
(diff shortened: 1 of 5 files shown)


openassessment/staffgrader/serializers/submission_list.py
```diff
@@ -0,0 +1,100 @@
...
+    def _get_anonymous_id_from_context(self, submission_uuid):
+        try:
+            return self.context[self.CONTEXT_SUB_TO_ANON_ID][submission_uuid]
+        except KeyError as e:
+            raise MissingContextException(
+                f"No submitter anonymous user id found for submission uuid {submission_uuid}"
+            ) from e
+
+    def get_dateGraded(self, workflow):
+        return str(workflow.grading_completed_at)
```
nsprenkle, line 82: Any reason to prefer this to a `DateTimeField`?

openassessment/staffgrader/staff_grader_mixin.py
```diff
@@ -140,13 +144,22 @@ def list_staff_workflows(self, data, suffix=''):  # pylint: disable=unused-argum
...
+
+        # Serialize workflows with the context, and return the dict of submissions
+        result = {}
+        for staff_workflow in staff_workflows:
+            try:
+                serialized_workflow = SubmissionListSerializer(staff_workflow, context=serializer_context).data
+                result[staff_workflow.identifying_uuid] = serialized_workflow
+            except MissingContextException as e:
+                log.exception("Failed to serialize workflow %d: %s", staff_workflow.id, str(e), exc_info=True)
+        return result
```
nsprenkle, line 158: Obligatory, "should this live in a super serializer"?

nsprenkle (approved):

openassessment/staffgrader/serializers/submission_list.py
jansenk: This is gross. The right thing to do here is return None if there's no dateGraded, and the DateTimeField will do that, but the current behavior is to return 'None' which this will do. I don't want to mix this refactor with that fix; I have another ticket for going through and making all the final changes we need to the ORA -> BFF api

openassessment/staffgrader/staff_grader_mixin.py
jansenk: I don't think so because I don't know that there's a way to do variable keys through a serializer. If we wanted this to just be a list of submissions, yes, but like this, I don't think there's a way to do this that isn't messier than this.

PR #1794, opened 2022-01-26
fix: update static url for image and link

leangseu-edx:
**TL;DR -** correctly update static url for image and link

JIRA: [AU-490](https://openedx.atlassian.net/browse/AU-490)

**What changed?**

- Modify **static** url on insert/edit image and anchor link (update logic)
- Modify **static** url on save contents (getcontent in this case)
- Remove **Sanitize** url on init because it would be redundant with on save.

**Developer Checklist**

**Testing Instructions**

Follow the jira ticket instruction

**Reviewer Checklist**

FYI: @edx/masters-devs-gta


openassessment/xblock/static/js/src/studio/oa_tiny_mce.js
```diff
@@ -23,30 +23,46 @@ export const oaTinyMCE = (options) => {
 
   const getFonts = () => CUSTOM_FONTS + STANDARD_FONTS;
   const baseAssetUrl = options.base_asset_url;
+  const staticUrl = '/static';
+
+  const sanitizeContent = (content) => {
+    // rewrite anchor link
+    let newContent = rewriteStaticLinks(
```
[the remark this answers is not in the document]
leangseu-edx: `edx-platform` a global js function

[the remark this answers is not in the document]
matthugs: looks like here: https://github.com/openedx/edx-platform/blob/db32ff2cdf678fa8edd12c9da76a76eef0478614/common/static/js/src/utility.js#L21

openassessment/xblock/static/js/src/studio/oa_tiny_mce.js
```diff
@@ -23,30 +23,46 @@ export const oaTinyMCE = (options) => {
...
+      content,
+      `<a href="${staticUrl}`,
+      `<a href="${baseAssetUrl}`,
+    );
+
+    // rewrite image src
+    newContent = rewriteStaticLinks(
+      newContent,
+      `<img src="${staticUrl}`,
+      `<img src="${baseAssetUrl}`,
```
matthugs, line 40: This is a bit of a hack, which would fail on html like, say ``

Why not use this the same way we do elsewhere in this library? https://github.com/openedx/edx-ora2/blob/dd6adeaf0572ecff3491e487f1df8eb0e1595c6f/openassessment/xblock/static/js/src/studio/oa_tiny_mce.js#L42

openassessment/xblock/static/js/src/studio/oa_tiny_mce.js
```diff
@@ -23,30 +23,46 @@ export const oaTinyMCE = (options) => {
...
+    if (data[key]) {
+      data[key] = rewriteStaticLinks(data[key], ...urls);
     }
   };
 
-  const setupTinyMCE = (ed) => {
-    ed.on('SaveImage', dataHandler('src', baseAssetUrl, '/static'));
-    ed.on('EditImage', dataHandler('src', '/static', baseAssetUrl));
-    ed.on('SaveLink', dataHandler('html', baseAssetUrl, '/static'));
-    ed.on('EditLink', dataHandler('html', '/static', baseAssetUrl));
```
matthugs, line 38: It looks like the original code wanted to translate from one format in one direction and to that same format in the other direction. Did we go about doing that incorrectly, before?

Said another way: how does this code behave in the following user story?
As a course author, I expect that when I write certain text into the platform and save it, then close the tab and return to edit it later, I find the same text I entered is still there.

openassessment/xblock/static/js/src/studio/oa_tiny_mce.js
leangseu-edx: getContent is actually get everything in html. If the user put anything even text that has `/static/` will get replace. It can be quite a confusing experience. This is just an improvement to what it was.

For image tag, it always generate ``. I tested with a few iteration of input.

openassessment/xblock/static/js/src/studio/oa_tiny_mce.js
leangseu-edx: I was a bit confuse too. As far as I can tell, `SaveLink` and `EditLink` was added later. The `dataHandler` was written to work for image only. There is slight change I put in.
For how it ended up in reverse order for one of them, I am really not sure. I tested with `debugger` on, it doesn't seem to have a correct behavior.

Current behavior, if you input `/static/...` in the url, it will get transform immediately. I do not have opinion on when it should get transform. I was just trying to make work from the previous code.

openassessment/xblock/static/js/src/studio/oa_tiny_mce.js
matthugs: Ok, so it sounds like we only support the rewriting of static links if folks insert the image via the WYSIWYG widget for adding images, and we don't rewrite static links if the user types html manually. Is that right?

If I have that right, I think a comment explaining what you've explained here would be helpful. (I would also urge you to make the comment longer than you'd consider necessary.)

openassessment/xblock/static/js/src/studio/oa_tiny_mce.js
matthugs: @leangseu-edx could you answer directly my question about the user story?

leangseu-edx: When the user write text, it should return the same contents he entered.
If he input images or anchor, there will be modification if and only if the user are copying from `studio -> File & Upload -> Copy Urls for studio`. It has the specific format of `/static/....`.

matthugs (approved):

jansenk (approved):

openassessment/xblock/static/js/src/studio/oa_tiny_mce.js
```diff
@@ -23,30 +23,57 @@ export const oaTinyMCE = (options) => {
...
+   * Within Studio -> File & Uploads -> Copy Urls.
+   * If the user used copy studios, the link get will be formatted
+   * similiar to `/static/....`. This will be invalid for usage and we
+   * need to modify it to be a correct link.
+   * @param {string} content HTML String
+   * @returns {string} new content
+   */
+  const rewriteContentUrl = (content) => {
+    // These url are generated by WYSIWYG editor, it is assumed to be formatted
+    // <a href='.....' >...</a> for link.
```
matthugs, line 39: I think a future reader of this code would be able to read the lines below it and know that this code would only work for things formatted as in the example in this comment. So they can probably figure out what your comment tells them, even without you having commented it.

The part that is most non-obvious about this code is why the assumption that anchor tags (and image tags, below) can safely be assumed to be in that format. Why that is is probably the most important thing to include in a comment here.

PR #1802, opened 2022-02-04
chore: bump pip-tools to 6.5.0 to resolve pip bug

jansenk:
make upgrade failed because of https://github.com/jazzband/pip-tools/issues/1558

6.5.0 resolves the issue so after merge requirements upgrade should work again


mattcarter (approved):
approved

PR #1818, opened 2022-02-24
fix: restore original tinymce behavior, fix trailing slash

jansenk:
**TL;DR -** I'm just going back to the original tinyMCE behavior, plus fixing the small `/` bug

JIRA: [AU-490](https://openedx.atlassian.net/browse/AU-490)

**What changed?**

 - In https://github.com/openedx/edx-ora2/pull/1512 we introduced the bug by improperly combining the "dataHandler" methods into one method and only setting/checking `src` rather than also including `href`
 - In https://github.com/openedx/edx-ora2/pull/1794 we attempted to fix the issue by fixing the order of the to->from in the dataHandler. We also did a more specific replacement of `<a href=<url>` and `` This fixed the issue of inserting bad links and images, but the static url const used in the "load all content" transform was missing a slash resulting in incorrect "resolved" URLs. Thankfully according to the read replica, there are only 3 of those that currently exist.
 
In this PR I'm:
- Removing the handlers for EditImage and EditLink. Originally their purpose was so that you would only ever see the "static" link. When the image/link was saved, we'd convert it from the static link to the real link, and then when you would click "edit" we'd transition back from the real link to the "static" link. I'm getting rid of that. The URL presented in the edit dialog should actually be the URL and having it go back and forth hides the actual real value of the URL. Unnecessary and potentially confusing.

- Removing the specific `<a href=<url>` and `` filtering for the full content transform. It's nice, and it will potentially prevent issues where we have a `/static/` that we don't want transformed, but unfortunately having the global replace is the current only way we can support resolving static URLs in audio or video embeds. (see https://openedx.atlassian.net/browse/AU-544)

**Developer Checklist**

**Testing Instructions**

 - Check out ORA tag 3.8.1 and install-local-ora. Insert a static link, and save. The link should be in the broken unresolved state.
 - Check out this branch, force-restart studio, re-local-install, and navigate to the ORA. The Link should still be unresolved. Open the edit modal and immediately hit save. The link should now be resolved. 
 
**Reviewer Checklist**

FYI: @openedx/content-aurora


nsprenkle (approved):
Looks good and verified behavior and smoke tested locally :shipit:

PR #1819, opened 2022-02-25
feat: ORA Report data should clarify which assessments are "scored"

jansenk:
### TL;DR - This fixes a bug, and adds a new feature on top of it.  

**Bug**: Incorrectly marking peer assessments as "scored" (part of final grade) when they are not.

**Feature**: The ORA Data report now marks each assessment as "scored" or not.

### JIRA: [AU-520](https://openedx.atlassian.net/browse/AU-520)

**What changed?**

- Changed the logic for marking assessments as "scored" to take flexible peer grading into account
- Added an entry in the ORA data report to specify whether or not the assessment was used in the final grade.
- The ticket is phrased as if it is asking for unscored grades to not appear, but I thought it would be better to still include all assessments, but have additional info about whether or not they were "scored". 

**Developer Checklist**

**Testing Instructions**

- In a course with at least three learners, create an ORA with a peer step, flexible peer grading, must_grade = 3 and must_be_graded_by = 3.
- Submit with all three learners.
- Enter the local sql shell. 
- Get the three submission ids, `SELECT id FROM submissions_submission ORDER BY id DESC LIMIT 3;` 
- Update the submissions to be more than seven days ago: `UPDATE  submissions_submission SET submitted_at="2021-03-01 00:00:00.0000" where id in (<ids>)` This will now reduce the required peer assessments to 1.
- Do three peer assessments as each learner
- Navigate to the instructor dashboard -> student data -> Generate ORA Data Report
- Open the file 
- The file should include an entry for all ORA submissions in the course. Find the ORA we are interested in. The section that describes assessments should include a line specifying that exactly one assessment per submission is included as a part of the score.

**Reviewer Checklist**

FYI: @openedx/content-aurora


nsprenkle (approved):
Visually looks good but haven't gotten to verify since there's a lot of manual testbed setup.

Also requesting future PRs relating to output reports contain a before/after in the PR description to assert against.

jansenk:
Here's a before / after of the same report:
Screen Shot 2022-03-15 at 2 06 50 PM
Screen Shot 2022-03-15 at 2 06 46 PM

PR #1828, opened 2022-03-18
test: added test suite for create_oa_submissions_from_file management command

jansenk:
**TL;DR -** added test suite for create_oa_submissions_from_file management command

JIRA: [AU-514](https://openedx.atlassian.net/browse/AU-514)

**What changed?**

- Tests! 
- Some changes to the management command itself to make it more testable
- **Note**: I'm not bumping version number for this

**Developer Checklist**

**Reviewer Checklist**

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/management/commands/create_oa_submissions_from_file.py
+++ b/openassessment/management/commands/create_oa_submissions_from_file.py
@@ -85,6 +85,9 @@ class Command(BaseCommand):
     """
     Management command to create submissions, assessments, and locks to make testing easier.
     """
+
+    CONFIG_FILE_LOCATION_BASE = join('/' 'edx', 'src', 'edx-ora2')
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.username_to_anonymous_user_id = None
@@ -148,7 +151,7 @@ class Command(BaseCommand):
         Check that the input file exits, and attempt to open and json parse the file.
         Returns the json parsed input file.
         """
-        file_path = join('/' 'edx', 'src', 'edx-ora2', file_path)
+        file_path = join(self.CONFIG_FILE_LOCATION_BASE, file_path)
         if not exists(file_path):
             raise CommandError(f'File {file_path} not found.')
 
@@ -190,18 +193,22 @@ class Command(BaseCommand):
 
         return learners, course_staff
 
-    def load_anonymous_user_ids(self, course_id, usernames):
-        """
-        Look up all Users with the given usernames, look up their anonymous user ids for the specified course, and
-        store a mapping from username to anonaymous id in self.username_to_anonymous_user_id
-        """
+    def _import_anonymous_id_for_user(self):
         try:
             from common.djangoapps.student.models import anonymous_id_for_user  # pylint: disable=import-error
+            return anonymous_id_for_user
         except ModuleNotFoundError as e:
             raise CommandError((
                 "Cannot import common.djangoapps.student.models.anonymous_id_for_user. "
                 "This management command must be run from the LMS shell."
             )) from e
+
+    def load_anonymous_user_ids(self, course_id, usernames):
+        """
+        Look up all Users with the given usernames, look up their anonymous user ids for the specified course, and
+        store a mapping from username to anonaymous id in self.username_to_anonymous_user_id
+        """
+        anonymous_id_for_user = self._import_anonymous_id_for_user()
         # Also include a superuser because when we reset we need to include a "reset by" anonymous id
         usernames.add(SUPERUSER_USERNAME)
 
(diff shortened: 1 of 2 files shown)


openassessment/management/commands/create_oa_submissions_from_file.py
```diff
@@ -224,14 +231,13 @@ def get_display_names(self, submissions_config):
         for ora_config in submissions_config:
             display_name = ora_config['displayName']
             if display_name in display_names:
-                raise CommandError('Duplicate ORA display name found in configuration file:' + display_name)
+                raise CommandError('Duplicate ORA display name found in configuration file: ' + display_name)
             display_names.append(display_name)
         return display_names
 
-    def load_ora_blocks(self, course_id, display_names):
+    def _load_ora_blocks_from_modulestore(self, course_id):
```
jansenk, line 238: same here, smallest possible mockable untestable unit

openassessment/management/tests/test_create_oa_submissions_from_file.py
```diff
@@ -0,0 +1,357 @@
...
+        super().setUp()
+        for username in [USERNAME_1, USERNAME_2, STAFF_USER_1, STAFF_USER_2, SUPERUSER_USERNAME]:
+            UserFactory.create(username=username)
+
+        self.mock_block = Mock(**MOCK_BLOCK_KWARGS)
+
+        def mock_anonymous_id_for_user(user, course_id):  # pylint: disable=unused-argument
+            return anonymous_user_id(user.username)
+
+        read_config_patcher = patch.object(
```
jansenk, line 285: Did this here rather than inline to avoid having to do four long daisy chained context managers

openassessment/management/tests/test_create_oa_submissions_from_file.py
```diff
@@ -0,0 +1,357 @@
...
+        for reset_user in [USERNAME_1, USERNAME_2]:
+            self.mock_block.clear_student_state.assert_any_call(
+                anonymous_user_id(reset_user),
+                COURSE_ID,
+                str(self.mock_block.location),
+                anonymous_user_id(SUPERUSER_USERNAME)
+            )
+
+
+class CreateSubmissionsFromFileCallCommandTest(TestCase):
```
jansenk, line 273: Is this mostly just duplicating tests above? Yes. Is it still valuable to have a test that actually calls the command and does a simplified check that the side effects happened? I think so

openassessment/management/commands/create_oa_submissions_from_file.py
```diff
@@ -190,18 +193,22 @@ def get_usernames(self, submissions_config):
 
         return learners, course_staff
 
-    def load_anonymous_user_ids(self, course_id, usernames):
-        """
-        Look up all Users with the given usernames, look up their anonymous user ids for the specified course, and
-        store a mapping from username to anonaymous id in self.username_to_anonymous_user_id
-        """
+    def _import_anonymous_id_for_user(self):
```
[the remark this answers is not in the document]
nsprenkle: Interesting... I don't really agree with this refactor.

**Reasoning** - I sort of feel like it's *more* bizarre to mock private internals of a class than mocking the imports. The imports are *supposed* to be these separate, modular, things which can be mocked because they exist outside your class under test, whereas mocking an internal function of the class under test inherently ties your tests to the internal structure of the class which makes them brittle and leaks implementation into your test.

openassessment/management/tests/test_create_oa_submissions_from_file.py
```diff
@@ -0,0 +1,357 @@
...
+        """ Test for when we are unable to find an anonymous_user_id for a username """
+
+        def mock_anonymous_id_for_user(user, course_id):  # pylint: disable=unused-argument
+            return user.username + "_anonymous"
+
+        for username in [USERNAME_1, USERNAME_2, STAFF_USER_1, STAFF_USER_2, SUPERUSER_USERNAME]:
+            UserFactory.create(username=username)
+
+        expected_error = 'Unable to load anonymous id for user(s) user_not_found'
+        with patch.object(self.cmd, '_import_anonymous_id_for_user', return_value=mock_anonymous_id_for_user):
```
nsprenkle, line 213: Don't like patching class internals. I'm actually surprised that pylint isn't yelling at you for this, it usually does.

nsprenkle (commented):
The tests are good but I disagree with the refactor for "testability", I think setting up the mocks on private class under test functions is an antipattern we should avoid.

openassessment/management/commands/create_oa_submissions_from_file.py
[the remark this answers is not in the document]
jansenk: Generally yes - if this was a top level import, and I could do the "mock it where it's used" mock()-ing as intended and just do `mock(...create_oa_submissions_from_file.anonymous_id_for_user)` then yes, I agree.

However, since these are inline imports that won't be able to resolve when running tests isolated from platform, when I say "mocking imports" I don't mean mocking the imported value (sorry, definitely not clear) I mean literally mocking python's import logic. doing a patch over the python import internals. THAT's the hammer.

openassessment/management/tests/test_create_oa_submissions_from_file.py
jansenk: mm, I don't necessarily agree with that, but I can see where you're coming from

jansenk:
https://openedx.atlassian.net/browse/AU-574 opened to address discussions of mocking

nsprenkle (approved):
Further test refactors slated for https://openedx.atlassian.net/browse/AU-574

PR #1833, opened 2022-03-28
fix: submission too long has the wrong message

jansenk:
**TL;DR -** [ Fix the error message for when an ORA submission is too long ]

JIRA: [AU-559](https://openedx.atlassian.net/browse/AU-559)

**What changed?**

- Change the error message when an ORA is too long to be about that instead of multiple submissions
- Add a note about autocorrect

**Developer Checklist**

**Testing Instructions**

- In an ORA, submit something super long. Like ... the entire text of moby dick https://www.gutenberg.org/files/2701/2701-h/2701-h.htm and see the new error message

**Reviewer Checklist**

FYI: @openedx/content-aurora


openassessment/xblock/submission_mixin.py
```diff
@@ -171,6 +171,13 @@ def submit(self, data, suffix=''):  # pylint: disable=unused-argument
                 )
                 if answer_too_long:
                     status_tag = 'EANSWERLENGTH'
+                    max_size = f"({int(api.Submission.MAXSIZE / 1024)} KB)"
+                    base_error = self._("Response exceeds maximum allowed size.")
+                    extra_info = self._(
+                        "Note: if you have a spellcheck or grammar check browser extension, "
+                        "try disabling, reloading, and reentering your response before submitting."
+                    )
+                    status_text = f"{base_error} {max_size} {extra_info}"
```
nsprenkle, line 180: This code path doesn't log the error. Is that something we want to have logged for our tracking?

nsprenkle (commented):
Note about logging. Also, is there a reason that package-lock and one of the djangojs files are mostly empty?

openassessment/xblock/submission_mixin.py
jansenk: Yep, missed that

jansenk:
> is there a reason that package-lock and one of the djangojs files are mostly empty?

Good catch? I have no clue. I'll try redoing translations and see if it fixes it

openassessment/xblock/submission_mixin.py
```diff
@@ -170,7 +170,17 @@ def submit(self, data, suffix=''):  # pylint: disable=unused-argument
                     for answer_err in err.field_errors.get('answer', [])
                 )
                 if answer_too_long:
+                    logger.exception(
+                        f"Attempted to submit response for {student_item_dict} but the submission is too long"
+                    )
```
nsprenkle, line 175: Nit: putting the student item dict at the end of the log message would improve search-ability and proximity to the surfaced message would help as well.

```suggestion
                    logger.exception(
                        f"Response exceeds maximum allowed size: {student_item_dict}"
                    )
```

nsprenkle (approved):
Looks good, with a suggestion for improved log messaging

openassessment/xblock/submission_mixin.py
jansenk: I was about to get sarcastically annoyed because at firt I thought I'd have to re-do the translations nonsense but then I realized this is a log message and not translated! wahoo

PR #1840, opened 2022-04-06
fix: javascript template strings were not being translated

jansenk:
**TL;DR -** Remove backticked template strings from calls to `gettext` because they were being ignored by translations jobs

JIRA: [AU-606](https://openedx.atlassian.net/browse/AU-606)

**What changed?**

- A very common pain point I would have is when I tried to run translations from my "local" (in virtualenv) ora environment, I would consistently see the process remove the translations for `"ve detected errors on the following tabs: ${tabNames.join("`. Googling led me to find some ... kinda similar issues? Not the same, but similar enough to make me investigate more.  noticed that that was the only backticked text that was showing up in translations at all! There were three others that were just not included.
- I changed them to use normal quotes and concatenation rather than template strings and everything ended up working out nicely.

**Unanswered Questions**
- Why *exactly* was this happening? What caused the difference in behavior?
- Why was the "lms" environment picking up that one mangled instance but not the other (arguably nicer) ones?
- The "lms" xgettext was an older version than my "local" version. was this something purposefully removed?

**Developer Checklist**

**Reviewer Checklist**

FYI: @openedx/content-aurora


mattcarter (approved):
Approved. The change in verbiage seems to be innocuous, but let product know.

PR #1843, opened 2022-04-07
chore: ignore uncovered LMS import data function

jansenk:
We're right on the borderline of coverage percentages. This function is an "LMS import" function, where we're basically hiding the LMS imports that won't work in a testing environment, which, by its nature, won't be covered by test coverage.

This is a decently big block, and hopefully a small change that will save some grief, at least in the long term before this is refactored away on our "LMS Import Object" ticket

### DIFF
--- a/openassessment/data.py
+++ b/openassessment/data.py
@@ -54,7 +54,7 @@ def _use_read_replica(queryset):
     )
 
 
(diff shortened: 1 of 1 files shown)


mattcarter (approved):
approved

PR #1866, opened 2022-06-16
fix: catch exception

jansenk:
**TL;DR -** Catch no such service exception in ORA studio

JIRA: [AU-660](https://2u-internal.atlassian.net/browse/AU-660)

**What changed?**
-  Catch no such service exception in ORA studio
- Put another log on the pile of "reasons it would be easier if we just had an author_view rather than just trying to do the same thing as in LMS"

**Developer Checklist**

**Reviewer Checklist**

FYI: @openedx/content-aurora


mattcarter (approved):
LGTM

PR #1873, opened 2022-07-06
fix: remove blank template, filter out peer template so it doesn't display twice

jansenk:
**TL;DR -** Remove blank template and make peer assessment the default ORA template 

JIRA: [AU-75](https://2u-internal.atlassian.net/browse/AU-75)

**What changed?**

- Remove blank ORA template
- implement optional template function to filter out "duplicate" peer assessment (the list of templates includes peer-assessment, and we are inserting "peer-assessment" as the "default" template)

**Testing Instructions**

Check out this branch and https://github.com/openedx/edx-platform/pull/30705. Install local ORA.

FYI: @openedx/content-aurora


leangseu-edx (approved):
LGTM 👍

nsprenkle (commented):
Random thought, any reason we aren't keeping the blank template but filtering it out in the `filter_templates` function? Would be a smaller change (only 1 PR) and seems easier to understand ("filter out the blank template" instead of "use peer as default, but filter out a second copy of it since we don't have a different default").

nsprenkle (approved):

PR #1988, opened 2023-07-12
Jkantor/flexible override

jansenk:
**TL;DR -** Make ORA peer grading respect the course-level flexible override setting

JIRA: [AU-1288](https://2u-internal.atlassian.net/browse/AU-1288)

**What changed?**

- Created a method in the workflow mixin,  get_course_workflow_settings, which returns a dictionary, that currently only has a single key, force_on_flexible_peer_openassessments. It is intended to be a place where any future course-level fields that affect ORAs to be read from.
- Pass the value from that through .... too much stuff. We need to pass that data through several levels of API calls to get it to where it's actually needed. There is no way currently to make this less awkward.
- When we check if flex grading is turned on, we now check if the course flag is true, and then check the block.

**Developer Checklist**

**Testing Instructions**

- Turn off the flex-grading course-wide flag
- Make an ORA that requires you to grade 1 person and be graded by 2 people, and is not flexibly graded.
- Submit submissions as two learners. They should each grade each other. They should not get grades.
- Turn on the flex-grading course-wide flag
- Manually edit one of the submissions to have been submitted more than 7 days ago (submission.submitted_at)
- Log in as each user. The learner whose submission was thrust back into the past should now have a grade. The learner who did not, should still not have a grade.

**Reviewer Checklist**

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/assessment/api/peer.py
+++ b/openassessment/assessment/api/peer.py
@@ -27,7 +27,17 @@ FLEXIBLE_PEER_GRADING_REQUIRED_SUBMISSION_AGE_IN_DAYS = 7
 FLEXIBLE_PEER_GRADING_GRADED_BY_PERCENTAGE = 30
 
 
-def required_peer_grades(submission_uuid, peer_requirements):
+def flexible_peer_grading_enabled(peer_requirements, course_settings):
+    """
+    Is flexible peer grading turned on? Either at the course override
+    level or the block level?
+    """
+    if course_settings.get('force_on_flexible_peer_openassessments'):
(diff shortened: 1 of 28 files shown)


jansenk:
I still need to write tests for the new functions. All test changes are just adjusting method signatures. 
- [ ] Write new tests!

nsprenkle (approved):
Surprised this has to touch so many places, proposing to possibly change `course_settings` to a kwarg to minimize number of signature changes, but otherwise the behavior change looks good.

PR #2013, opened 2023-08-02
feat: create ORA "date_config_type" field

leangseu-edx:
**TL;DR -** add no_dates field to xblock and let ora use course start, end date.

JIRA: https://2u-internal.atlassian.net/browse/AU-1314

**What changed?**

- Create a new field on the ORA xblock `date_config_type`
- This field set to `manual`, `subsection` and `course_end`

NOTE: this only work with lms. `runtime` libs does not seem to work in studio. `is_course_staff` also not working on studio, so I assume this is acceptable.

**Developer Checklist**

**Testing Instructions**

**Reviewer Checklist**

FYI: @openedx/content-aurora


jansenk (changes requested):
This is entirely my fault - I forgot to hit "save" on this ticket when I was changing the description. I have gone in and updated the description to the changes we actually want. This is a perfectly valid implementation of the ticket as it was handed to you but isn't actually what we want to do. I take full responsibility for this mistake.

 https://2u-internal.atlassian.net/browse/AU-1314

openassessment/xblock/openassessmentblock.py
```diff
@@ -1127,6 +1133,13 @@ def is_closed(self, step=None, course_staff=None):
         if course_staff:
             return False, None, DISTANT_PAST, DISTANT_FUTURE
 
+        if self.date_config_type == 'course_end':
+            course = self.get_course()
+            if course.start and course.end:
```
jansenk, line 1138: is it possible for these to not be truthy? what is this checking for?

jansenk (changes requested):
Looks good but a few suggestions

openassessment/xblock/openassessmentblock.py
leangseu-edx: Locally, when the course just created. Start or end date doesn't get define. This is for the safety check.

jansenk (approved):

PR #2014, opened 2023-08-03
chore: fixing make upgrade

awais786:
Upgrading `xblock-sdk` It is using very old version
Bump the version.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -6,36 +6,37 @@
 #
 appdirs==1.4.4
     # via fs
-asgiref==3.6.0
+asgiref==3.7.2
     # via django
 bleach==6.0.0
     # via -r requirements/base.in
-boto3==1.26.94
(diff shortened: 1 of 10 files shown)


requirements/base.txt
```diff
@@ -119,17 +120,16 @@ pbr==5.11.1
...
+psutil==5.9.5
     # via edx-django-utils
 pycparser==2.21
     # via cffi
 pymongo==3.13.0
     # via edx-opaque-keys
 pynacl==1.5.0
     # via edx-django-utils
-python-dateutil==2.4.0
+python-dateutil==2.8.2
```
awais786, line 131: edx-platform is using 2.8.2
https://github.com/openedx/edx-platform/blob/master/requirements/edx/base.txt#L903

salman2013 (approved):

PR #2028, opened 2023-08-14
chore: assessment api methods

muselesscreator:
**TL;DR -** [ A short summary of what this PR does and why ]

Initial rip of api logic from non-peer assessment mixins.


openassessment/xblock/api/assessments/student_training.py
```diff
@@ -0,0 +1,104 @@
...
+            {'prompt': self.prompt, 'criteria': self.block.rubric_criteria_with_labels},
+            self.examples
+        )
+
+    @property
+    def example_context(self):
+        context, error_message = self._parse_example(self.example)
+        return {
+            'error_message': error_message,
+            'essay_context': context
```
nsprenkle, line 98: Is "essay" the original wording? Because it would avoid it if it is not, these are frequently (but not always) essays and using that language could cause confusion.

muselesscreator: was `training_essay_context`

openassessment/xblock/api/assessments/workflow.py
```diff
@@ -0,0 +1,42 @@
...
+    def is_cancelled(self):
+        return self.status == 'cancelled'
+
+    @property
+    def is_done(self);
+        return self.status == 'done'
+
+    @property
+    def is_waiting(self);
+        return self.status == 'waiting'
```
nsprenkle, line 33: Some sneaky semicolons
```suggestion
    @property
    def is_done(self):
        return self.status == 'done'

    @property
    def is_waiting(self):
        return self.status == 'waiting'
```

openassessment/xblock/api/assessments/workflow.py
```diff
@@ -0,0 +1,42 @@
...
+    @property
+    def has_status(self):
+        return bool(self.status)
+
+    @property
+    def is_self_complete(self):
+        return self.workflow
+            .get('status_details', {})
+            .get('self', {})
+            .get('complete', False)
```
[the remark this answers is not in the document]
muselesscreator: :-p

PR #2031, opened 2023-08-18
feat: API to perform ORA Workflows batch update

jszewczulak:
**TL;DR -** Added API to batch-update ORA Workflows for different scopes. Each endpoint is decorated as a Celery task to be executed asynchronously 

JIRA: [JIRA-1295](https://2u-internal.atlassian.net/browse/AU-1295)

**What changed?**

- added new module with API (Tasks

**Developer Checklist**

**Testing Instructions**
- `pytest openassessment/tests/test_workflow_batch_update_api.py::TestWorkflowBatchUpdateAPI` 
-  end-to-end testing will be performed in stage env when Celery worker and queue are in place (separate PR) and scheduler is configured (WIP)

**Reviewer Checklist**

FYI: @openedx/content-aurora


openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,269 @@
...
+
+from opaque_keys.edx.keys import UsageKey
+from openassessment.runtime_imports.functions import modulestore
+from openassessment.assessment.models import PeerWorkflow
+from openassessment.workflow import api
+
+logger = logging.getLogger(__name__)
+
+
+def update_workflows_for_ora_block(item_id):
```
jansenk, line 18: My understanding was that these would be a chain of functions that each put more tasks on the celery queue. My understanding of the high-level flow was:

```
- Query for which ORAs had learners that were waiting for a peer review for more than 7 days
- for each course, create a celery task, passing in the course id and list of potential
 ora blocks as params
- that celery task would load the course block and check if the course-wide flexible 
override was enabled, then for each ora item it was given, put a new celery task on 
the queue with the override value and the item id as parameters
- that celery task would load the specified ora block. if the course override and the
 block-level setting for flexible grading were both disabled, noop, otherwise, look 
 up all waiting 7 day old workflows for this block id and put individual celery tasks 
 on the queue with a workflow identifier and the block's assessment requirements
- that celery task would then run the individual workflow update.
```

Why did this architecture change? Here we have the multiple levels, but not really in a chaining async task way, but in the sense of "choose which context you want to do this in" and then we do all work within the one worker. updating multiple courses within a single celery task is too much for a single worker to do. 

I know we discussed potentially "squashing" one or two of those "layers" above into one layer, but I think the current approach goes a bit too far. The work needs to be split up more.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,269 @@
...
+        bool: True if given ORA is configured with flexible peer grading set "ON"
+    """
+    workflow_requirements = openassessmentblock.workflow_requirements()
+    if workflow_requirements.get('peer') and workflow_requirements['peer'].get('enable_flexible_grading'):
+        return True
+
+    return False
+
+
+def get_blocked_peer_workflows():
```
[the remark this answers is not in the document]
jansenk: This, I think, would also make more sense as a class method on PeerWorkflow

jansenk (changes requested):
I have some questions about the overall approach.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,269 @@
...
+    Args:
+    item_id (str): Identifier for the ORA Block
+        e.g. 'block-v1:edX+DemoX+Demo_Course+type@openassessment+block@1676f4b05f0642249ff724e7c07d869e'
+
+    Raises:
+        OraWorkflowBatchUpdateException: If batch process fails and cannot continue, e.g. number of errors
+                                        threshold was exceeded, etc.
+    """
+    try:
+        start = time.time()
```
nsprenkle, line 31: Proposed refactor: replace manual time calculation with a timing decorator, can remove those 7 lines of code from each function.

nsprenkle (commented):
This is a start but still feels unpolished. There are certainly more elegant ways and with less code duplication by focusing on more modular scope (single submission, group of submissions for a single ORA, and all candidate submissions for an ORA) instead of re-writing the function over and over with slight modification.

I think @jansenk has good points of how to make a smaller set of functions that can be called with the appropriate scope.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,269 @@
...
+        peer_workflows (list): list of `PeerWorkflow` objects
+
+    Returns:
+        assessment_requirements_dict (dict): Dictionary:  <submission_uuid>:<assessment_requirements>
+    """
+    # dictionary <submission_uuid>:<workflow_requirements>
+    workflow_requirements = {}
+    store = modulestore()
+
+    for peer_workflow in peer_workflows:
```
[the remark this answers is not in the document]
jszewczulak: Yes, caching makes sense I just didn't want to introduce it in the first iteration before running it and capturing execution times, etc. If the performance is acceptable there is a significant complexity reduction. 
Anyway  caching was added along with other changes.

openassessment/workflow_batch_update_api.py
jszewczulak: Makes sense for simple cases, however we log other information besides args and execution times, so keeping it in function code.

jszewczulak:
Based on our discussions I have introduced chained asynchronous task execution and data caching.
These are the scopes (async tasks) that will be executed:

```
>update_workflows_for_all_blocked_submissions_task
>>update_workflows_for_course_task
>>>update_workflows_for_ora_block_task
>>>>update_workflow_for_submission_task
```

I left out the `(update workflows for) student `scope because it breaks the scalability pattern thus requires further complexity in logic and local data cache handling. Out of scope for this effort. 

Additional thoughts on implemented changes comparing to earlier (flattened) execution sequence:

**Chained/nested Celery tasks** execution may prove to be problematic in maintenance: 

1. Celery provides means to control and monitor tasks but they are designed to cover all use cases thus may not be completely coverying our unique needs. E.g. Handling errors when updating individual submissions - before we could set threshold to ignore certain number of errors, then stop execution if the number of errors are greater than threshold. 
2. Asynchronous execution is always challenging to orchestrate, with proper error handling, etc. Stacking multiple async layers, delegating process control and having no single place to handle errors (the way we need) is risky.
3. This approach scales but most probably we will never need to handle amount of data that would prove to be problematic with earlier, flatted execution chain flow. Submitting a single async job with all submission level updates done synchronously would still delegate task to be executed on separate process but giving us better control over process. I don't think the goal is the infinite scalability here. Single celery process execution environment could be further scaled vertically if needed. 
4. We are losing capability to invoke jobs synchronously
5. we are forced to keep `@shared_task` wrappers in the same module as the other API, breaking `tasks.py` module convention


**Data cache** - Implemented approach is not the most efficient but it is a compromise to achieve scalability and the API clarity:
   `update_workflows_for_course(course_id, workflow_update_parameters=None)`
   For all scopes the `workflow_update_parameters` data cache dict structure is the same, with each scope consuming the data it requires
   To optimize it (lookup) we would need to either introduce more cache dict structures or duplicate data, or create separate DTO class, or introduce class level state, all of which introduce additional complexity, risks and other overheads, hence the compromise.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,595 @@
...
+        try:
+            # retrieve openassessment block
+            ora_block_key = UsageKey.from_string(peer_workflow.item_id)
+            ora_block = store.get_item(ora_block_key)
+
+            # retrieve course block
+            course_block_key = UsageKey.from_string(peer_workflow.course_id)
+            course_block = store.get_item(course_block_key)
+
+            if is_flexible_peer_grading_on(ora_block) or course_block.force_on_flexible_peer_openassessments:
```
nsprenkle, line 367: This can be simplified, we have a `flexible_peer_grading_enabled` check which takes into account both ORA and course settings.

You may also have to change `course_block.force_on_flexible_peer_openassessments` to a `course_block.get('force_on_flexible_peer_openassessments')` if there's ever a case where the course block doesn't have that setting yet.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,595 @@
...
+    except (UpdateWorkflowForSubmissionException, Exception) as e:  # pylint: disable=broad-except
+        logger.error(
+            "Batch workflow update for ORA block failed. "
+            "item_id=%s error_message:%s",
+            item_id,
+            str(e))
+        raise UpdateWorkflowsForOraBlockException(str(e)) from e
+
+
+def update_workflow_for_submission(submission_uuid, assessment_requirements, course_settings):
```
nsprenkle, line 262: Again, I think this could be helpful to have a version where `assessment_requirements` and `course_settings` are `kwargs`. That way, you could call this flexibly, without having to do those lookups in advance and having this function work better for that scope, but omitting those searches when the data is provided.
```suggestion
def update_workflow_for_submission(submission_uuid, assessment_requirements=None, course_settings=None):
```

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,595 @@
...
+    """
+    try:
+        start = time.time()
+
+        peer_workflows = get_blocked_peer_workflows()
+        workflow_update_parameters = get_workflow_update_parameters(peer_workflows)
+        if workflow_update_parameters.get("courses") is not None:
+            for course_object in workflow_update_parameters.get("courses"):
+                # execute asynchronously (submit Celery task)
+                update_workflows_for_course_task.apply_async([course_object["course_id"], workflow_update_parameters])
```
nsprenkle, line 94: This seems clunky, why are we passing the entire `workflow_update_parameters` when we only care about the parameters for a given course at this level?

This would also make unpacking of data at each subsequent level less complicated so you could get the data inline instead of having a bunch of additional utility functions to unpack that data for you.

nsprenkle (commented):
Moving in the right direction, but could still use some cleanup for better modularity and removing duplicate item lookups.

openassessment/workflow_batch_update_api.py
jszewczulak: Do we ever envision running asynchronous job for as single submission?

openassessment/workflow_batch_update_api.py
jszewczulak: `is_flexible_peer_grading_on(ora_block)` needs to extract some data from ora_block, that's why it makes sense to wrap it with separate method. course_block.force_on_flexible_peer_openassessments returns `boolean` no need to move it

openassessment/workflow_batch_update_api.py
nsprenkle: This is not the asynchronous job, this is an API util, no? There's a separate wrapper for the async job.

The challenge of keeping it like this (and including things like timing scripts, etc. in this scope) is you are limiting yourself to one way of using this API utility. If you ever **do** want to do something similar, but with slight variation, you're going to have to make new versions of this function.

In general, the tasks can be pretty rigid but it's useful to have the underlying APIs be pretty general-purpose.

openassessment/workflow_batch_update_api.py
nsprenkle: What I'm suggesting is to leverage the `flexible_peer_grading_enabled` check instead, which will already take both block and course level settings into account and lets us define that logic in fewer places.

openassessment/workflow_batch_update_api.py
nsprenkle: I have, but I think passing around the same, giant data blob is not the most efficient way to do it.

If I'm misunderstanding how the data gets shared to the async task, please correct me, but wouldn't this mean *every* one of those tasks has to be sent the **entire** "cached" lookup for every course, ORA, and submission we care about? Even if the scope only cares about a very limited subset of that data?

nsprenkle: That could make sense if you had, for example, a statically typed data object to pass around and didn't want to make variations for the different contexts. But in Python land, you're not getting anything by adding extra context unnecessarily to a dictionary. You are just creating extra work in unpacking the data later.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,595 @@
...
+    }
+    ```
+    """
+    workflow_update_parameters_dict = {}
+    store = modulestore()
+    for peer_workflow in peer_workflows:
+        try:
+            # retrieve openassessment block
+            ora_block_key = UsageKey.from_string(peer_workflow.item_id)
+            ora_block = store.get_item(ora_block_key)
```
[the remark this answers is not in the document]
jszewczulak: I think we need to look at the the execution context here - above code is being invoked only if data cache is not present, that additional cache lookup will never be executed if added.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,595 @@
...
+logger = logging.getLogger(__name__)
+
+
+@shared_task(bind=True,
+             acks_late=True,
+             autoretry_for=(Exception,),
+             max_retries=5,
+             retry_backoff=True,
+             retry_backoff_max=500,
+             retry_jitter=True)
```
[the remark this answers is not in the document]
jszewczulak: Interesting, this is the formatting PyCharm provides, plus doesn't break `pylint` 
2:1 ;-)

openassessment/workflow_batch_update_api.py
jszewczulak: I have made update for this

openassessment/workflow_batch_update_api.py
jszewczulak: To end the thread, I have made updates for this, all update_* signatures are consolidated for similar behavior.

openassessment/workflow_batch_update_api.py
jszewczulak: Again, this is a compromise after considering different approaches, it is a trade off but comes with advantages (easily scalable, etc). If it proves to be problematic for performance reasons we can update it.

openassessment/workflow_batch_update_api.py
jszewczulak: I have made changes, however, these are my comments:

Decorators (or annotations in Java) quite often (and in case of the single API boundaries often to very often) affect code readability. I would rather have more verbose imperatively written code and see everything that's going on, than additional complexity that is scattered all over the place and required additional hops to understand logic flow. 

E.g. in this case I had to introduce additional return object wrapper, with not clean (**kwargs antipattern) constructor implementation and the decorator function itself. As a result we have slightly less verbose code at the expense of complexity and readability. I would say the net result is 0

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,629 @@
...
+# pylint: disable=unused-argument
+def update_workflow_for_submission_task(self, submission_uuid, assessment_requirements=None, course_settings=None):
+    """
+    Async task wrapper
+    """
+    return update_workflow_for_submission(submission_uuid, assessment_requirements, course_settings)
+
+
+def logging_decorator(func):
```
nsprenkle, line 76: Nit: update for more descriptive title.

```suggestion
def log_result_and_time(func):
```

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,629 @@
...
+
+    def wrapper(*args, **kwargs):
+        start = time.time()
+        result = func(*args, **kwargs)
+        end = time.time()
+        log_message = ""
+        if result is not None:
+            for key, value in result.__dict__.items():
+                log_message = log_message + " " + str(key) + "=" + str(value)
+        log_message = log_message + " processing_time=" + str(round(end - start, 5))
```
nsprenkle, line 89: Nit: here is a good place for an f-string, also probably don't want to call `__dict__`directly.

```suggestion
            for key, value in dict(result).items():
                log_message = log_message + f" {str(key)}={str(value)}"
        log_message = log_message + f" processing_time={str(round(end - start, 5))}
```

Or even shorter:

```suggestion
        log_message = " ".join([f"{str(key)}={str(value)}" for key, value in dict(result).items()])
        log_message += f" processing_time={str(round(end-start, 5))}"
```

openassessment/workflow_batch_update_api.py
nsprenkle: I still disagree, on the record, but I don't see any further value arguing over it if you're willing to stand behind that decision.

openassessment/workflow_batch_update_api.py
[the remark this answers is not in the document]
nsprenkle: Can you clarify? I see this being called for `update_workflows_for_all_blocked_submissions`. When that is run you look up all peer workflows and, even if they are all under the same ORA block, you look up the block for each peer workflow.

What I'm suggesting is you check if the `peer_workflow.item_id` is already in `workflow_update_parameters_dict` and, if it is, don't do the ORA block lookup.

That seems like a pretty safe and obvious optimization that is not implemented currently, unless I'm misunderstanding the code.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,629 @@
...
+
+            # retrieve course block
+            course_block_key = UsageKey.from_string(peer_workflow.course_id)
+            course_block = store.get_item(course_block_key)
+
+            if is_flexible_peer_grading_on(ora_block, course_block):
+                # we are interested only in submissions for ORA with flexible peer grading configured
+                course_settings = {
+                    'force_on_flexible_peer_openassessments': course_block.force_on_flexible_peer_openassessments}
+                _add_workflow_update_parameters(workflow_update_parameters_dict,
```
nsprenkle, line 393: Nit: I personally dislike this style of function (pass in a dict which gets modified inside the function) as it hides and can tweak data in ways that are hard to inspect and test for. Would prefer:

```suggestion
                update_params = get_update_params_for_submission(<whole lotta args, for some reason>)
                workflow_update_parameters_dict.update(update_params)
```

This will also let you test that data transform more easily, if you needed to test it.

nsprenkle (commented):
Still one necessary optimization blocking approval, then a few nits which can be taken into consideration or ignored.

openassessment/workflow_batch_update_api.py
jszewczulak: Agreed on updating via reference, typically you would additionally return updated object (like in your example), however, for the internal/private/helper call this is acceptable IMO

openassessment/workflow_batch_update_api.py
jszewczulak: It is more than that so didn't want to define the purpose more finely.

openassessment/workflow_batch_update_api.py
jszewczulak: updated, however result object is not a dictionary so still hacky `__dict__` access

openassessment/workflow_batch_update_api.py
nsprenkle: Minimally should be something like `log_task_info`, even if you keep it that vague. Also having "decorator" after a decorator reads like "logging decorator decorator" and is redundant.

Or would require a docstring update, you can't say "does what it says" if it doesn't actually "say what it does".

nsprenkle: Also the logs should probably include the scope / function name. It does appear to have a description, but probably want something more machine-searchable.

openassessment/workflow_batch_update_api.py
```diff
@@ -374,99 +357,98 @@ def get_workflow_update_parameters(peer_workflows):
...
+            elif peer_workflow.item_id not in cache:
+                # retrieve openassessment block
+                ora_block_key = UsageKey.from_string(peer_workflow.item_id)
+                ora_block = store.get_item(ora_block_key)
+
+                workflow_update_data = _add_ora_data(workflow_update_data=workflow_update_data,
+                                                     course_id=peer_workflow.course_id,
+                                                     item_id=peer_workflow.item_id,
+                                                     submission_uuid=peer_workflow.submission_uuid,
+                                                     assessment_requirements=ora_block.workflow_requirements())
```
nsprenkle, line 397: This still has redundancy, you never add the `item_id` to the cache, so this will look up the ORA multiple times. This whole flow seems a bit convoluted and can be better thought out.

Proposed:
```
get peer_workflows
    collected_data = {}
    for workflow in peer_workflow:
        if peer_workflow.course_id not in collected_data:
            get course_data
            collected_data.coursed.push(course_data)
       if peer_workflow.item_id not in collected_data
            get ora_data
            collected_data.courses[peer_workflow.course_id].assessments = ora_data
       # for all workflows
       add submission data
    return collected_data
```

jszewczulak: item_id is added when we add entire course object, line #379, I don't see redundancy, it is actually quite optimized adding ora data without additional lookup.  
BTW, your logic does not include flexible peer grading verification.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,640 @@
...
```
jansenk, line 397: this reads to me like submission uuids that have a course and item that are already in the cache won't be included in the list, no?

jansenk (commented):
A few suggestions. it seems like you may be making this more complicated than this needs to be?

openassessment/tests/test_workflow_batch_update_api.py
```diff
@@ -0,0 +1,545 @@
...
+        pw_pat.completed_at = timezone.now() - datetime.timedelta(days=2)
+        pw_pat.save()
+        blocked = update_api.get_blocked_peer_workflows()
+        # we still expect 1 blocked submission (Tim)
+        self.assertEqual(len(blocked), 1)
+        self.assertEqual(blocked[0].student_id, 'Tim')
+
+    def test_get_blocked_peer_workflows_for_course(self):
+        # pylint: disable=unused-variable
+        tim_sub, tim = self._create_student_and_submission("Tim", "Tim's answer")
```
jansenk, line 60: this seems like a lot of duplicated setup. You could pull some of this into setUpTestData and then it would be shared between tests

jansenk, line 60: If you're not using these variables, just don't assign them.
```suggestion
        self._create_student_and_submission("Tim", "Tim's answer")
```

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,640 @@
...
+             autoretry_for=(Exception,),
+             max_retries=3,
+             retry_backoff=True,
+             retry_backoff_max=500,
+             retry_jitter=True)
+def update_workflows_for_all_blocked_submissions_task(self):  # pylint: disable=unused-argument
+    """
+    Async task wrapper
+    """
+    return update_workflows_for_all_blocked_submissions()
```
jansenk, line 30: This and the other tasks should be in tasks.py

openassessment/tests/test_workflow_batch_update_api.py
```diff
@@ -0,0 +1,545 @@
...
+        self.assertFalse(update_api.is_flexible_peer_grading_on(ora_block, course_block))
+
+        del (ora_block.requirements['peer'])
+        self.assertFalse(update_api.is_flexible_peer_grading_on(ora_block, course_block))
+
+        course_block.force_on_flexible_peer_openassessments = True
+        self.assertTrue(update_api.is_flexible_peer_grading_on(ora_block, course_block))
+
+    @patch('openassessment.workflow_batch_update_api.modulestore')
+    @patch('openassessment.workflow_batch_update_api.UsageKey.from_string')
```
jansenk, line 150: nit: you could avoid this by just using valid course keys. course-v1:org+course+1 course-v1:org+course+2

openassessment/tests/test_workflow_batch_update_api.py
```diff
@@ -0,0 +1,545 @@
...
+        self.assertEqual(wup["courses"][1]["course_id"], "course_id_2")
+        self.assertEqual(wup["courses"][1]["assessments"][0]["item_id"], "item_id_3")
+
+    # pylint: disable=unused-variable
+    def get_peer_workflows_for_test_get_workflow_update_data(self):
+        tim_sub, tim = self._create_student_and_submission("Tim", "Tim's answer")
+        miles_sub, miles = self._create_student_and_submission("Miles", "Miles's answer")
+        pat_sub, pat = self._create_student_and_submission("Pat", "Pat's answer")
+
+        pw_tim = PeerWorkflow.objects.get(student_id=tim["student_id"])
```
jansenk, line 172: you could also modify create_student_and_submission to take these as optional params

openassessment/tests/test_workflow_batch_update_api.py
```diff
@@ -0,0 +1,545 @@
...
+        peer_workflows = self.get_peer_workflows_for_test_get_workflow_update_data()
+        wup = update_api.get_workflow_update_data(peer_workflows)
+
+        #
+        self.assertEqual(len(wup["courses"]), 2)
+        self.assertEqual(wup["courses"][0]["course_id"], "course_id_1")
+        self.assertEqual(wup["courses"][0]["assessments"][0]["item_id"], "item_id_1")
+        self.assertEqual(wup["courses"][0]["assessments"][1]["item_id"], "item_id_2")
+        self.assertEqual(wup["courses"][1]["course_id"], "course_id_2")
+        self.assertEqual(wup["courses"][1]["assessments"][0]["item_id"], "item_id_3")
```
jansenk, line 164: I think if you had multiple learners per ora and checked for submission uuids here, this test would fail

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,640 @@
...
+
+                    workflow_update_data = _add_course_data(
+                        workflow_update_data=workflow_update_data,
+                        course_id=peer_workflow.course_id,
+                        item_id=peer_workflow.item_id,
+                        submission_uuid=peer_workflow.submission_uuid,
+                        assessment_requirements=ora_block.workflow_requirements(),
+                        course_settings=course_settings)
+
+                cache.add(peer_workflow.course_id)
```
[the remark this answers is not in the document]
jszewczulak: good catch, saves some db trips

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,640 @@
...
+    data = get_workflow_update_data(peer_workflows)
+    return _get_ora_data(data, item_id)
+
+
+def _add_course_data(workflow_update_data, course_id, item_id, submission_uuid, course_settings,
+                     assessment_requirements):
+    """
+    Adds course data structure to the provided data cache dictionary
+    """
+    if workflow_update_data is None or workflow_update_data.get("courses") is None:
```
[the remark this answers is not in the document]
jszewczulak: This is "private" function so in theory yes, we could assume input data quality; regardless it is a good habit to validate input.

openassessment/workflow_batch_update_api.py
jszewczulak: They should, they were but with nested/chained async call it is not possible (python imports cycle)

openassessment/tests/test_workflow_batch_update_api.py
jszewczulak: they are being used, line #70

jszewczulak: We differ in approach here, more generic (shared) data setup in tests, the less clarity.

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,640 @@
...
+    """
+    _get_course_data(workflow_update_data, course_id).get("assessments").append(
+        {"item_id": item_id, "assessment_requirements": assessment_requirements,
+         "submissions": [{"submission_uuid": submission_uuid}]})
+    return workflow_update_data
+
+
+def _get_course_data(workflow_update_data, course_id):
+    """
+    Helper function to return `update_workflow_data` dict object structure
```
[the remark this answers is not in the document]
jszewczulak: Yeah, there is a lot to consider here though. The problem starts with very initial data lookup (PeerWorkflow, submission level data ) and the necessity to manage cache for different scopes.  E.g. How do we pass data for ORA scope without passing cache exceeding ORA scope, or without changing method signatures to include `course_id`
  The "awkward" happens to compensate for the suboptimal data storage design and happens in a black box here allowing for cleaner API (course scope update - pass course_id, ORA scope update - pass item_id).

openassessment/workflow_batch_update_api.py
jszewczulak: This is a good finding, I guess late night large scope refactorings are not a good idea..., fixed.

openassessment/tests/test_workflow_batch_update_api.py
jszewczulak: I looked at it but not sure if I understand the problem

openassessment/workflow_batch_update_api.py
```diff
@@ -0,0 +1,269 @@
...
+            str(end - start))
+
+    except (OraWorkflowBatchUpdateErrorThresholdException, Exception) as e:  # pylint: disable=broad-except
+        logger.error(
+            "Batch workflow update. Error occurred while updating workflows for all blocked submissions.  Error:%s",
+            str(e))
+        raise OraWorkflowBatchUpdateException(str(e)) from e
+
+
+def update_workflows(assessment_requirements_dict, error_threshold=10):
```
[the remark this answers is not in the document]
jszewczulak: Outdated, code has been refactored

openassessment/workflow/workflow_batch_update_api.py
```diff
@@ -0,0 +1,600 @@
...
+                        assessment_requirements=ora_block.workflow_requirements(),
+                        course_settings=course_settings)
+
+                cache.add(peer_workflow.course_id)
+                cache.add(peer_workflow.item_id)
+
+            elif peer_workflow.item_id not in cache:
+                # retrieve openassessment block
+                ora_block_key = UsageKey.from_string(peer_workflow.item_id)
+                ora_block = store.get_item(ora_block_key)
```
jansenk, line 336: I think this is also not correct. If an ora is in a course that is already in the cache, we fall into this block, but we don't check for whether flex peer grading is enabled for this item. You need to add a check for that here, otherwise you'll be adding info for all second/third/fourth/nth oras regardless of whether they should actually have been included

openassessment/workflow/workflow_batch_update_api.py
```diff
@@ -0,0 +1,600 @@
...
+
+                workflow_update_data = _add_ora_data(workflow_update_data=workflow_update_data,
+                                                     course_id=peer_workflow.course_id,
+                                                     item_id=peer_workflow.item_id,
+                                                     submission_uuid=peer_workflow.submission_uuid,
+                                                     assessment_requirements=ora_block.workflow_requirements())
+            else:
+                workflow_update_data = _add_submission_data(workflow_update_data,
```
jansenk, line 344: Same here, we just check if the item and course are in the cache, not if they are flexible.

openassessment/workflow/workflow_batch_update_api.py
```diff
@@ -0,0 +1,600 @@
...
+    """
+    Retrieve ORA peer workflows not completed for >7 days
+
+    Returns:
+        list (PeerWorkflow): list of workfows not completed for > 7 days
+    """
+
+    filters = {
+        'created_at__lte': timezone.now() - datetime.timedelta(days=7),
+        'completed_at__isnull': True,
```
jansenk, line 262: Good thing I checked again. 
completed_at is the timestamp of when the learner completed the peer grades they needed to complete
grading_completed_at is the timestamp of when they recieved enough peer grades.
So, we want someone who HAS completed their assigned peer assessments and HASN'T recieved enough peer assessments.
```suggestion
        'grading_completed_at__isnull': True,
        'completed_at__isnull': False,
```

jansenk (changes requested):
Sorry, I want to merge this too, but I noticed a couple more issues

openassessment/workflow/workflow_batch_update_api.py
jszewczulak: Good findings, I have redesigned cache management approach a bit to make it more straightforward.

openassessment/workflow/workflow_batch_update_api.py
```diff
@@ -228,23 +228,26 @@ def update_workflow_for_submission(submission_uuid, assessment_requirements=None
         raise UpdateWorkflowForSubmissionException(str(e)) from e
 
 
-def is_flexible_peer_grading_on(ora_block, course_block):
+def is_flexible_peer_grading_on(assessment_requirements, course_settings):
```
jansenk, line 231: This and all tests for it can be removed, and you can use `openassessment.assessment.api.peer.flexible_peer_grading_enabled` It already has this logic, is tested, and has the identical signature

openassessment/workflow/workflow_batch_update_api.py
```diff
@@ -259,7 +262,7 @@ def get_blocked_peer_workflows(course_id=None, item_id=None, submission_uuid=Non
 
     filters = {
         'created_at__lte': timezone.now() - datetime.timedelta(days=7),
-        'completed_at__isnull': True,
+        'grading_completed_at__isnull': True,
```
jansenk, line 265: You can also add `completed_at__isnull: False` here. If a learner hasn't completed the number of peer reviews they need to complete, their workflow isn't going to do anything when we update it.

openassessment/workflow/workflow_batch_update_api.py
```diff
@@ -282,126 +285,163 @@ def get_workflow_update_data(peer_workflows):
...
+
+    submissions_cache = set([])
+
     for peer_workflow in peer_workflows:
+
         try:
-            if peer_workflow.course_id not in cache:
-                # retrieve course block
+            # pylint: disable=consider-iterating-dictionary
+            if peer_workflow.course_id not in course_settings_cache.keys():
```
jansenk, line 316: The call to keys() and the pylint are not necessary. `if <value> in <dict>:` checks if the value is in the key set implicitly.

openassessment/workflow/workflow_batch_update_api.py
```diff
@@ -282,126 +285,163 @@ def get_workflow_update_data(peer_workflows):
...
+                # retrieve course block from DB
                 course_block_key = UsageKey.from_string(peer_workflow.course_id)
                 course_block = store.get_item(course_block_key)
+                # add course settings to temp cache
+                course_settings_cache[peer_workflow.course_id] = {
+                    'force_on_flexible_peer_openassessments': course_block.force_on_flexible_peer_openassessments}
 
-                # retrieve openassessment block
+            # pylint: disable=consider-iterating-dictionary
+            if peer_workflow.item_id not in assessment_requirements_cache.keys():
```
jansenk, line 325: same here

jansenk (approved):
There are a few small nits I'd like if you'd address but other than that I believe this is complete

openassessment/workflow/workflow_batch_update_api.py
jszewczulak: The signature is the same but slightly different logic. if ORA peer grading or flexible peer grading is not defined existing api will return `None`.

PR #2046, opened 2023-09-07
make legacy template

leangseu-edx:
**Merging into `feat/2023-duai-ui` branch**

TLDR; move existing template to `legacy` folder

Ticket: https://2u-internal.atlassian.net/browse/AU-1411


muselesscreator (approved):
LGTM, but would like second pair of eyes

openassessment/templates/base.html
```diff
@@ -0,0 +1,3 @@
+<div>
+  <h1>Hello world</h1>
+</div>
```
nsprenkle, line 3: What is the purpose of this file?

muselesscreator: That is going to have an iframe to the new xblock UI

PR #2053, opened 2023-09-20
feat: ORA BFF Page Data

nsprenkle:
**TL;DR -** Adds top-level info for ORA page data (student/workflow specific data we will need to render any of the views in our MFE).

JIRA: [AU-1382](https://2u-internal.atlassian.net/browse/AU-1382)

**What changed?**

- Added 2 new XBlock handlers:
    - `get_block_learner_submission_data` - Our submission view in the MFE
    - `get_block_learner_assessment_data` - Our assessment view in the MFE
- Implemented serializers for top-level data.
- **NOTE** that at this time I have specifically left the `submission` field unimplemented. This is a large task (since we must gather data from different contexts for each step of the ORA workflow) that should be split up into a few tickets.
- Added beginning of a Grades API and extended the Peer Assessments API to get required data for our serializers.
- Added new operating mode for `get_submission_to_assess` from Peer API. Adding `peek=True` will allow you to check for peers to assess without beginning the assessment workflow. This helps us determine if we are currently blocked waiting for assessments, though in a sub-optimal way since this fetches a lot of data we won't end up using. Thought it was still a better approach than duplicating a lot of code to handle this slightly different use-case.

**Developer Checklist**

**Testing Instructions**

- Manual test of data via Postman (or similar API tool) at different parts of the workflow.
- **No unit tests, currently.**, has been ticketed separately in [AU-1464](https://2u-internal.atlassian.net/browse/AU-1464)

**Reviewer Checklist**

FYI: @openedx/content-aurora


jansenk (commented):
Thankfully it doesn't look like there's actually much duplication here, but there is a fair bit of adjacency / overlap. I suggest you merge this first and then I can rebase my changes on top of / into yours

openassessment/xblock/apis/grades_api.py
```diff
@@ -0,0 +1,79 @@
...
+            "points_earned": (Int) awarded points
+            "points_possible": (Int) max possible points
+        }
+        """
+        submission_uuid = self._get_submission_uuid()
+        assessment = self_api.get_assessment(submission_uuid)
+        return {
+            "points_earned": assessment["points_earned"],
+            "points_possible": assessment["points_possible"],
+        }
```
jansenk, line 33:
```suggestion
        assessment = self_api.get_assessment(submission_uuid)
        if assessment is not None:
            return {
                "points_earned": assessment["points_earned"],
                "points_possible": assessment["points_possible"],
            }
        return None
```

jansenk: Same suggestion for peer

openassessment/xblock/ui_mixins/mfe/ora_config_serializer.py
```diff
@@ -217,9 +217,7 @@ class OraBlockInfoSerializer(Serializer):
 
     def get_baseAssetUrl(self, block):
         # pylint: disable=protected-access
-        return block.get_base_url_path_for_course_assets(
```
jansenk, line 220: Why the filename change for this file?

jansenk (approved):
looks good. a few comments

openassessment/xblock/ui_mixins/mfe/ora_config_serializer.py
nsprenkle: I've been splitting the serializers by the response shape they are returning for. This is helping (me at least) keep things separated, since otherwise this would be a 1k+ line file by the time we're done with it which makes finding things harder.

openassessment/xblock/apis/grades_api.py
nsprenkle: Good catch, TY

PR #2059, opened 2023-09-29
feat: mfe file actions with legacy refactor

jansenk:
**TL;DR -** File actions refactors and BFF MFE endpoints

JIRA: [AU-1391](https://2u-internal.atlassian.net/browse/AU-1391)

**What changed?**

- Refactored legacy actions into shared API that raises exceptions for error states
- legacy handlers now call into shared api and handle request-level validations and error handling in handler method
- added mfe handlers to call into shared api and do request-level validations and error handling
- wrote tests for mfe handlers 

**Developer Checklist**

FYI: @openedx/content-aurora


openassessment/xblock/apis/submissions/submissions_actions.py
```diff
@@ -256,3 +261,97 @@ def save_submission_draft(student_submission_data, block_config_data, block_subm
...
+def get_upload_url(content_type, file_name, file_index, block_config, submission_info):
+    """
+    Request a URL to be used for uploading content for a given file
+
+    Returns:
+        A URL to be used to upload content associated with this submission.
+
+    """
+    if not block_config.allow_multiple_files:
+        if submission_info.files.has_any_file_in_upload_space():
```
leangseu-edx, line 340: Why is this checking with `not allow_multiple_files` then check max file in the function? What if the `allow_multiple_files = 2`? Unless that isn't an option.

jansenk: allow_multiple_files is just a boolean, it's only gonna be false (only one file) or true (as many as we allow)
we don't current;y allow control of # of submissions beyond that

jansenk: this is a really janky use case that gets really into the under the hood implementation details of file uploads. Rather than being able to trust the block-level fields, we basically hit every possible file key that could exist for this submission on the file backend and see if they exist

leangseu-edx: Sorry, I was trying to say `max_allowed_uploads`.

openassessment/xblock/ui_mixins/mfe/constants.py
```diff
@@ -13,4 +13,12 @@ class ErrorCodes:
...
     EMPTY_ANSWER = "ERR_EMPTY_ANSWER"
+    DELETE_NOT_ALLOWED = "ERR_DELETE_NOT_ALLOWED"
     UNKNOWN_ERROR = "ERR_UNKNOWN"
+
+
+class HandlerSuffixes:
+    SUBMISSION_DRAFT = 'draft'
+    SUBMISSION_SUBMIT = 'submit'
+    FILE_UPLOAD = 'upload'
+    FILE_DELETE = 'delete'
```
nsprenkle, line 24: Any reason to make this a `class` instead of an `Enum`? Because this feels more like an enum.

openassessment/xblock/ui_mixins/mfe/mixin.py
```diff
@@ -86,13 +89,51 @@ def _submission_create(self, data):
 
     @XBlock.json_handler
     def submission(self, data, suffix=""):
-        if suffix == 'draft':
+        if suffix == HandlerSuffixes.SUBMISSION_DRAFT:
             return self._submission_draft(data)
-        elif suffix == "create":
+        elif suffix == HandlerSuffixes.SUBMISSION_SUBMIT:
             return self._submission_create(data)
```
nsprenkle, line 95: I like these more but they technically differ from our contract. If we want to change, need to just communicate to frontend and update in contract.

Current paths:
- POST {xblock_id}/handler/submission/submit
- POST {xblock_id}/handler/submission/draft

Expected paths:
- POST {xblock_id}/handler/submit
- POST {xblock_id}/handler/submission/draft

nsprenkle (commented):
Looks good, only not approving because the paths differ from our contract. Make sure to update to align to contract or update the contract and communicate changes to @muselesscreator

openassessment/xblock/ui_mixins/legacy/handlers_mixin.py
```diff
@@ -111,25 +117,85 @@ def save_submission(self, data, suffix=""):  # pylint: disable=unused-argument
...
-
     @XBlock.json_handler
     def save_files_descriptions(self, data, suffix=""):  # pylint: disable=unused-argument
-        return save_files_descriptions(self.config_data, self.submission_data, data)
+        serializer = SaveFilesDescriptionRequestSerializer(data=data)
+        if not serializer.is_valid():
+            return {
+                'success': False,
+                'msg': self.config_data.translate('File descriptions were not submitted.'),
+            }
```
nsprenkle, line 127: Cool use of serializer validation 💯

openassessment/xblock/ui_mixins/mfe/ora_config_serializer.py
```diff
@@ -221,3 +226,20 @@ def get_baseAssetUrl(self, block):
...
+    hasCancelled = BooleanField(source="workflow.has_cancelled", default=False)
+    hasRecievedGrade = BooleanField(source="workflow.has_recieved_grade", default=False)
+    teamInfo = TeamInfoSerializer(source="team_info")
+    response = SerializerMethodField(source="*")
+
+    def get_response(self, data):
+        # The source data is different if we have an in-progress response vs a submitted response
+        if data['workflow']['has_submitted']:
+            return SubmissionSerializer(data['response']).data
+        return InProgressResponseSerializer(data).data
```
nsprenkle, line 245: This is a dupe, no? We shouldn't have this in ORA Config, since it is actually part of our submission.

https://github.com/openedx/edx-ora2/blob/0e06ac19a05a447f427595a22d78c60ddee2fd28/openassessment/xblock/ui_mixins/mfe/submission_serializers.py#L95-L109

nsprenkle (commented):
Looks good, but you have a serializer dupe that needs to be removed.

openassessment/xblock/ui_mixins/mfe/ora_config_serializer.py
jansenk: You're right, this is a mistake from my merge. Good catch.

PR #2060, opened 2023-09-29
test: add test serializer for page data

leangseu-edx:
Merge to feature branch **feat/2023-dual-ui**


mattcarter (approved):
Looks good, but I'm the approver of last resort, so having others look would be a good idea.

PR #2074, opened 2023-10-13
feat: refactor legacy assessment actions and add assessment mfe bff handler

jansenk:
**TL;DR -** - Add mfe BFF handler for submitting assessments

JIRA: [AU-1302](https://2u-internal.atlassian.net/browse/AU-1302)

**What changed?**

- Refactor legacy assessment actions to create shared assessment xblock api layer
- Implement assessment handler for mfe bff

FYI: @openedx/content-aurora


openassessment/xblock/ui_mixins/mfe/mixin.py
```diff
@@ -226,3 +232,70 @@ def get_learner_submission_data(self):
             'response': response,
             'file_data': file_data,
         }
+
+    def _assessment_submit_handler(self, data):
+        serializer = AssessmentSubmitRequestSerializer(data=data)
+        if not serializer.is_valid():
+            raise OraApiException(400, error_codes.INCORRECT_PARAMETERS, serializer.errors)
```
nsprenkle, line 239: 💯

openassessment/xblock/ui_mixins/mfe/page_context_serializer.py
```diff
@@ -294,7 +294,7 @@ def get_submission(self, instance):
             if active_step == "training":
                 response = instance.student_training_data.example
             elif active_step == "peer":
-                response = instance.peer_assessment_data().get_peer_submission()
+                response = instance.peer_assessment_data().get_active_assessment_submission()
```
nsprenkle, line 297: This looks to me like we will have to separately trigger a `get_peer_submission` when we load the page if we don't have a submission yet... I would think we want to `get_new_or_active_assessment_submission`, no?

@muselesscreator, correct me if I'm wrong.

openassessment/xblock/ui_mixins/legacy/views_mixin.py
```diff
@@ -4,11 +4,11 @@
...
-from .peer_assessments.views import render_peer_assessment, peer_path_and_context
-from .self_assessments.views import render_self_assessment, self_path_and_context
-from .staff_assessments.views import render_staff_assessment, staff_path_and_context
-from .student_training.views import render_student_training, training_path_and_context
-from .submissions.views import render_submission, get_submission_path, get_submission_context
+from .views.peer import render_peer_assessment, peer_path_and_context
+from .views.self import render_self_assessment, self_path_and_context
+from .views.staff import render_staff_assessment, staff_path_and_context
+from .views.training import render_student_training, training_path_and_context
+from .views.submission import render_submission, get_submission_path, get_submission_context
```
nsprenkle, line 11: Is there a reason for this refactor? Outside of preference?

nsprenkle (commented):
Mostly looks good, with the following caveats:
1. I initial page context load should also fetch a submission to assess. Proposing refactor to `get_peer_submission` to check if you have an active one and return that, else get a new one. 
2. Trying to understand why refactors include moving files around.

openassessment/xblock/ui_mixins/legacy/views_mixin.py
jansenk: After I removed the "actions" files, the only thing remaining in the individual step folders was "views.py", and I thought It would be cleaner to have those in a folder rather than many folders all with one single file.

openassessment/xblock/ui_mixins/mfe/page_context_serializer.py
jansenk: This is what I was told yesterday but I have better clarity on the desired behavior now.

nsprenkle (approved):
LGTM

PR #2078, opened 2023-10-17
chore: fix studio base view

leangseu-edx:
Merge to **feat/2023-dual-ui**

TLDR; `base.html` exist in both `edx-platform` and `edx-ora2`. It confuse for which to use for `cms`.

What change?
- move `base.html` to subdirection `openassessmentblock`
- I would assume we will need subdirectory for all html in xblock because unique prefix seems important.


openassessment/xblock/openassessmentblock.py
```diff
@@ -611,7 +611,8 @@ def student_view(self, context=None):  # pylint: disable=unused-argument
             "title": self.title,
             "xblock_id": self.get_xblock_id(),
         }
-        template = get_template("base.html")
+        # template = get_template("legacy/oa_base.html")
```
mattcarter, line 614: Is there a reason to leave in the commented out code?

leangseu-edx: Nope, it was a mistake.

mattcarter (approved):
approved

jansenk (approved):

PR #2081, opened 2023-10-18
test: fix flaky test by freezing time

jansenk:
Fix flaky test.

The URL was different because an expiry timestamp was generated on the URL
sometimes the expiration would be a microsecond apart. This should fix that failure

FYI: @openedx/content-aurora


mattcarter (approved):
approved

PR #2088, opened 2023-10-23
test: fix failing tests by mocking xmodule runtime

jansenk:
Some test failures were merged into the feature branch. This PR is a quick fix. More work should be done in the future to remove all references to the xmodule runtime

### DIFF
--- a/openassessment/xblock/test/base.py
+++ b/openassessment/xblock/test/base.py
@@ -310,6 +310,30 @@ class XBlockHandlerTestCaseMixin:
         with open(os.path.join(base_dir, path)) as file_handle:
             return file_handle.read()
 
+    @staticmethod
+    def _create_mock_runtime(
+            item_id,
+            is_staff,
+            is_admin,
+            anonymous_user_id,
+            user_is_beta=False,
+    ):
+        """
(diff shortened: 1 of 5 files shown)


openassessment/xblock/test/base.py
```diff
@@ -310,6 +310,30 @@ def load_fixture_str(path):
...
             return file_handle.read()
 
+    @staticmethod
+    def _create_mock_runtime(
+            item_id,
+            is_staff,
+            is_admin,
+            anonymous_user_id,
+            user_is_beta=False,
+    ):
```
nsprenkle, line 320: Nit: I would always vote for more kwargs so we don't have to pass the same args for most of our use cases.
```suggestion
    def _create_mock_runtime(
            item_id,
            anonymous_user_id,
            is_staff=False,
            is_admin=False,
            user_is_beta=False,
    ):
```

nsprenkle (approved):
LGTM

openassessment/xblock/test/base.py
jansenk: I agree but in the spirit of smallest change, this was just cut-pasted from another file

PR #2099, opened 2023-10-31
chore: Update to the new version of paragon in the new scope.

feanil:
Part of https://github.com/openedx/axim-engineering/issues/23

This replaces the `@edx/paragon` packag to point to the `paragon` package at
the `openedx` scope(`@openedx/paragon`). Imports have been updated to use the
same locations in the new package.


brian-smith-tcril:
looks like this line needs to be changed https://github.com/openedx/edx-ora2/blob/63ee4fa62d4ab82c5ccd7bed6d910a5b0655ad68/openassessment/xblock/static/sass/_openassessment.scss#L99

PR #2107, opened 2023-11-03
fix: fix effective assessment type

nsprenkle:
**TL;DR -** Correctly return the effective assessment type

JIRA: [AU-1529](https://2u-internal.atlassian.net/browse/AU-1529)

**Reviewer Checklist**

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/xblock/apis/grades_api.py
+++ b/openassessment/xblock/apis/grades_api.py
@@ -15,21 +15,44 @@ class GradesAPI:
         return self._block.submission_uuid
 
     @property
+    def score_overridden(self):
+        """
+        Determine if score was overridden by staff.
+        Adapted from grade_mixin._get_assessment_type.
+
+        Returns: True if score was overridden by staff, False otherwise.
+        """
+        workflow = self._block.get_workflow_info()
+        score = workflow['score']
+
+        complete = score is not None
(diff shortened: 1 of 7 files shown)


openassessment/xblock/apis/grades_api.py
```diff
@@ -14,6 +14,46 @@ def __init__(self, block):
...
+        Determine which assessment step we will use as our "graded" step.
+
+        This follows the order:
+        1) Staff (if assessment received / overridden)
+        2) Peer (if assessment step configured)
+        3) Self (if assessment step configured)
+
+        NOTE: The logic in a few places differs, but this combines the best I've found.
+        """
+        if self.staff_score is not None or self.score_overridden:
```
leangseu-edx, line 47: So if `staff_score` is part of `assessment_steps` but the staff didn't do the grading, it will fall automatically to peer?

Btw, should we add 
```python
if not has_received_final_grade:
  return None
```

nsprenkle: There are actually 2 places where we check this in the previous implementation, and they disagree on this logic. Looking in more, I think I will switch it to `if self.staff_score is in self._block.assessment_steps` plus override logic.

I don't think we need the `if not has_received_final_grade` since we can tell what the assessment type should be (with the exception of override) even if we don't have a grade yet, no?

leangseu-edx (approved):

PR #2112, opened 2023-11-14
chore: respond step name for waiting

leangseu-edx:
**TL;DR -** Handle waiting state

JIRA: https://2u-internal.atlassian.net/browse/AU-1542

**What changed?**

- On waiting with `staff` required, return staff
- On waiting with `peer`, return peer
- (Unit test) If staff not require, it should be done

**Developer Checklist**

**Testing Instructions**

**Reviewer Checklist**

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/xblock/ui_mixins/mfe/page_context_serializer.py
+++ b/openassessment/xblock/ui_mixins/mfe/page_context_serializer.py
@@ -252,17 +252,20 @@ class ProgressSerializer(Serializer):
         """Return the active step name"""
         if not instance.workflow_data.has_workflow or instance.workflow_data.is_cancelled:
             return "submission"
+        elif instance.workflow_data.is_waiting:
+            # If we are waiting, we are either waiting on a peer or a staff grade.
+            # Staff takes precidence, so if a required (not automatically inserted) staff
+            # step exists, we are considered to be in "staff". If there is a peer, we are
+            # considered to be in "peer"
+            workflow_requirements = instance.workflow_data.workflow_requirements
+            if "staff" in workflow_requirements and workflow_requirements["staff"]["required"]:
+                return "staff"
+            elif "peer" in workflow_requirements:
+                return "peer"
+            else:
+                raise UnknownActiveStepException("Workflow is in waiting but no staff or peer step is required.")
         else:
-            step_name = STEP_NAME_MAPPINGS[instance.workflow_data.status]
-            if step_name == "waiting":
-                workflow_requirements = instance.workflow_data.workflow_requirements
-                if 'staff' in workflow_requirements and workflow_requirements['staff']['required']:
-                    return "staff"
-                elif 'peer' in workflow_requirements:
-                    return "peer"
-                else:
-                    raise UnknownActiveStepException("Workflow is in waiting but no staff or peer step is required.")
-            return step_name
+            return STEP_NAME_MAPPINGS[instance.workflow_data.status]
 
 
 class PageDataSerializer(Serializer):
--- a/openassessment/xblock/ui_mixins/mfe/page_context_serializer.py
+++ b/openassessment/xblock/ui_mixins/mfe/page_context_serializer.py
@@ -22,6 +22,10 @@ from openassessment.xblock.ui_mixins.mfe.submission_serializers import DraftResp
 from openassessment.xblock.ui_mixins.mfe.serializer_utils import STEP_NAME_MAPPINGS, CharListField
 
 
+class UnknownActiveStepException(Exception):
+    """Raised when we can't determine the active step"""
+
+
 class AssessmentScoreSerializer(Serializer):
     """
     Returns:
@@ -249,7 +253,16 @@ class ProgressSerializer(Serializer):
         if not instance.workflow_data.has_workflow or instance.workflow_data.is_cancelled:
             return "submission"
         else:
-            return STEP_NAME_MAPPINGS[instance.workflow_data.status]
+            step_name = STEP_NAME_MAPPINGS[instance.workflow_data.status]
+            if step_name == "waiting":
+                workflow_requirements = instance.workflow_data.workflow_requirements
+                if 'staff' in workflow_requirements and workflow_requirements['staff']['required']:
+                    return "staff"
+                elif 'peer' in workflow_requirements:
+                    return "peer"
+                else:
+                    raise UnknownActiveStepException("Workflow is in waiting but no staff or peer step is required.")
+            return step_name
 
 
 class PageDataSerializer(Serializer):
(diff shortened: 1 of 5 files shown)


openassessment/xblock/ui_mixins/mfe/page_context_serializer.py
```diff
@@ -248,7 +252,16 @@ def get_activeStepName(self, instance):
         if not instance.workflow_data.has_workflow:
             return "submission"
         else:
-            return STEP_NAME_MAPPINGS[instance.workflow_data.status]
+            step_name = STEP_NAME_MAPPINGS[instance.workflow_data.status]
```
jansenk, line 255: How about we remove waiting from step_name_mappings and then do

```
elif instance.workflow_data.is_waiting:
    <what you have here>
else:
     return STEP_NAME_MAPPINGS[instance.workflow_data.status]

```

openassessment/xblock/ui_mixins/mfe/test_page_context_serializer.py
```diff
@@ -606,6 +619,29 @@ def test_self_assessment_not_available(self, xblock):
 
         self.assertNestedDictEquals(expected_data, progress_data)
 
+    @scenario('data/grade_scenario_self_staff.xml', user_id='Alan')
+    def test_self_staff_assessment__waiting(self, xblock):
+        # Given I am on the staff step and waiting for submissions
+        self.create_submission_and_assessments(xblock, 'submission_text', [], [], SELF_ASSESSMENT)
+        xblock.workflow_data.workflow['status'] = 'waiting'
```
jansenk, line 626: I don't think this line is doing anything. Did you mean for this to be
```suggestion
        assert xblock.workflow_data.workflow['status'] == 'waiting'
```

openassessment/xblock/ui_mixins/mfe/test_page_context_serializer.py
```diff
@@ -606,6 +619,29 @@ def test_self_assessment_not_available(self, xblock):
...
+
+    @scenario('data/grade_scenario_self_staff_not_required.xml', user_id='Alan')
+    def test_self_staff_assessment__not_required(self, xblock):
+        # Given I am on the staff step and waiting for submissions
+        self.create_submission_and_assessments(xblock, 'submission_text', [], [], SELF_ASSESSMENT)
+
+        # When I ask for progress
+        progress_data = ProgressSerializer(xblock).data
+
+        # Expect active step to be staff instead of waiting
```
jansenk, line 642: This comment doesn't reflect what's happening. I think this test is actually not necessary because as soon as we submit the self assessment in this case, since the staff is not required (and therefore auto-generated) it's just skipped and we go straight to done rather than waiting

openassessment/xblock/ui_mixins/mfe/test_page_context_serializer.py
```diff
@@ -475,6 +475,19 @@ def test_peer_assessment(self, xblock):
 
         self.assertNestedDictEquals(expected_data, progress_data)
 
+    @scenario("data/peer_only_scenario.xml", user_id="Alan")
+    def test_peer_assessment__waiting(self, xblock):
+        # Given I am on the peer step and waiting for submissions
+        self.create_test_submission(xblock)
+        # xblock.workflow_data.workflow.set("status", "waiting")
+        xblock.workflow_data.workflow["status"] = "waiting"
```
jansenk, line 483:
```suggestion
        # The block requires us to grade 5, and be graded by three, so have us grade 5, and be graded by none
        peers = ['p1', 'p2', 'p3', 'p4', 'p5']
        self.create_submission_and_assessments(xblock, 'submission_text', peers, [self.PEER_ASSESSMENTS[0]] * 5, None, waiting_for_peer=True)
        assert xblock.workflow_data.workflow["status"] = "waiting"
```

openassessment/xblock/ui_mixins/mfe/test_page_context_serializer.py
```diff
@@ -606,6 +619,29 @@ def test_self_assessment_not_available(self, xblock):
...
+    @scenario('data/grade_scenario_self_staff_not_required.xml', user_id='Alan')
+    def test_self_staff_assessment__not_required(self, xblock):
+        # Given I am on the staff step and waiting for submissions
+        self.create_submission_and_assessments(xblock, 'submission_text', [], [], SELF_ASSESSMENT)
+
+        # When I ask for progress
+        progress_data = ProgressSerializer(xblock).data
+
+        # Expect active step to be staff instead of waiting
+        self.assertEqual('done', progress_data['activeStepName'])
```
nsprenkle, line 643: Your comment contradicts the actual test. Which is correct?

openassessment/xblock/ui_mixins/mfe/page_context_serializer.py
```diff
@@ -248,7 +252,16 @@ def get_activeStepName(self, instance):
...
+            if step_name == "waiting":
+                workflow_requirements = instance.workflow_data.workflow_requirements
+                if 'staff' in workflow_requirements and workflow_requirements['staff']['required']:
+                    return "staff"
+                elif 'peer' in workflow_requirements:
+                    return "peer"
+                else:
+                    raise UnknownActiveStepException("Workflow is in waiting but no staff or peer step is required.")
+            return step_name
```
nsprenkle, line 264: Nit: Can remove the need to early assign step name, and use more consistent quotation marks.

```suggestion
            if instance.workflow_data.status == "waiting":
                workflow_requirements = instance.workflow_data.workflow_requirements
                if "staff" in workflow_requirements and workflow_requirements["staff"]["required"]:
                    return "staff"
                elif "peer" in workflow_requirements:
                    return "peer"
                else:
                    raise UnknownActiveStepException("Workflow is in waiting but no staff or peer step is required.")
            return STEP_NAME_MAPPINGS[instance.workflow_data.status]
```

nsprenkle (commented):
Open question on one of the tests

openassessment/xblock/ui_mixins/mfe/test_page_context_serializer.py
```diff
@@ -606,6 +619,29 @@ def test_self_assessment_not_available(self, xblock):
...
+
```
jansenk, line 644: Add an xml test scenario for peer then staff, then write a test with the same setup as my recommended setup for your peer step above, where we're waiting on peer AND staff, and assert that the active step will be "staff"

jansenk, line 644: Add a test case where we are in a self-only ORA (no staff or peer), then patch the workflow API object's is_waiting to return True. Assert the expected error is raised.

jansenk (changes requested):
Some test changes

jansenk (approved):

PR #2116, opened 2023-11-17
feat: Hotjar tracking integration to trigger survey within ORA iframe

jszewczulak:
**TL;DR -** Manual Hotjar tracking integration to trigger survey within ORA xblock iframe. 

JIRA: [AU-1471](https://2u-internal.atlassian.net/browse/AU-1471)

**Developer Checklist**

**Testing Instructions**

**Reviewer Checklist**

FYI: @openedx/content-aurora


openassessment/templates/openassessmentblock/base.html
```diff
@@ -5,6 +5,17 @@
   If the new views are not enabled, `.ora-view` will be hidden, otherwise, `.ora-legacy-view`
   will be hidden.
 -->
+<!-- Hotjar Tracking Code for https://www.edx.org/ -->
+<script>
+    (function(h,o,t,j,a,r){
```
leangseu-edx, line 10: Question, are we tracking all of the open assessment?

jszewczulak: Yes, hotjar integration is for all ORAs, then we have logic in views to dispatch event under certain conditions to render survey widget.

leangseu-edx: Wouldn't that make sense to conditionally having hotjar include? If we only need it for survey, I don't see the reason to have it connect to all of the open assessment.

jszewczulak: Again, this is just hotjar platform integration for ORA iframe (we already have full integration in LMS). While we do want to dispatch survey event conditionally, we don't want to limit ourselves to survey only. Hotjar offers other services and we don't want to refer to code change and deployment each time we want to configure on Hotjar. Tat would be anti-pattern.

leangseu-edx: I don't think it is anti-pattern to just do what needed to be done. If there isn't a feature require hotjar in the roadmap or near future, it seems over kill to connect hotjar to all open assessment.

mattcarter (approved):
Approved

kdmccormick:
I'm not happy to see more edX.org-specific tracking code added directly to an Open edX repository like this.

Since you've already added it, though, @jszewczulak will this load on sites that do not set the hotjar site id?

PR #2119, opened 2023-11-20
fix: remove rather than hide unused ORA view

jansenk:
Remove unused ORA view rather than hide.

All of the staff area logic got really confiused when there were more elements than expected. Fixible, but this is a much much simpler fix


mattcarter (approved):
approved

nsprenkle (approved):
Have tested and verified fix :shipit:

PR #2129, opened 2023-12-01
refactor: reduce duplicate workflow update calls

nsprenkle:
**TL;DR -** Reduce the number of duplicate workflow updates from common ORA data load patterns.

**What changed?**

Asking for a `workflow` from the `WorkflowAPI` called `get_workflow_info` which would update the workflow before fetching workflow information. Since many of the other fields on `WorkflowAPI` referenced that `workflow` property, asking for **any** of this information would cause additional refreshes of the workflow, needlessly.

Loading the ORA MFE, I was averaging over 20 workflow refreshes and a local load time of ~500ms.

This PR begins to add a few refactors aimed at reducing how frequently we call a workflow update to (ideally) once per user action, calling cached workflow info any time between those actions.

- Memoize `workflow` property on `WorkflowAPI` to return last-fetched workflow data instead of re-updating/fetching.
- Load `workflow` on `WorkflowAPI` init (e.g. XBlock / MFE load)
- Reduce the number of times `WorkflowAPI` is initialized by memoizing `workflow_data` on `OpenassessmentBlock` and `ORADataAccessor` on fetch / update.
- Attempt to re-fetch `workflow` when a workflow is not present (this is not ideal from a performance standpoint but appears to fix a large number of tests).

**Other changes:**
- Remove unused `grades` accessor from `WorkflowAPI`.
- Updated some mocking to fix tests which relied on old call structure.
- Fixed a handful of typos I encountered along the way.

**NOTE** that this does mean it is more possible to get stale workflows so it is up to each endpoint / caller to determine if they need to refresh the workflow state on a per-action basis.

**Developer Checklist**

**Testing Instructions**

- Unit testing
- Smoke test suites

**Reviewer Checklist**

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/__init__.py
+++ b/openassessment/__init__.py
@@ -2,4 +2,4 @@
 Initialization Information for Open Assessment Module
 """
 
-__version__ = '6.0.27'
+__version__ = '6.0.28'
--- a/openassessment/workflow/api.py
+++ b/openassessment/workflow/api.py
@@ -195,7 +195,7 @@ def update_from_assessments(
     This is a little wonky from a REST, get-doesn't-change-state point of view,
     except that what's stored in the `AssessmentWorkflow` isn't the canonical
     true value -- it's just the most recently known state of it based on the
-    last known requirments. For now, we have to query for truth.
+    last known requirements. For now, we have to query for truth.
 
     Args:
         submission_uuid (str): Identifier for the submission the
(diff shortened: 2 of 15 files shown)


mattcarter (approved):
approved

openassessment/__init__.py
```diff
@@ -2,4 +2,4 @@
 Initialization Information for Open Assessment Module
 """
 
-__version__ = '6.0.24'
+__version__ = '6.025'
```
leangseu-edx, line 5:
```suggestion
__version__ = '6.0.25'
```

leangseu-edx (approved):
Update version bump. The rest look good.

jansenk (approved):
great job

openassessment/__init__.py
nsprenkle: Lol, I fixed this and forgot to push 🤦

PR #2135, opened 2023-12-07
fix: add step param

jansenk:
Add a "step" param to the "create assessment" MFE endpoint. This controls which type of assessment we create.


mattcarter (approved):
approved

PR #2138, opened 2023-12-11
feat: mfe uat fixes

jansenk:
Two changes:

 - Don't return None if there are no files, just return an empty list
 - Include the max file size in the file upload config


mattcarter (approved):
approved

PR #2149, opened 2023-12-20
feat: serial workflow advance fix

nsprenkle:
**TL;DR -** Check both `graded` and `complete` workflow status indicators to confirm we've completed a workflow step instead of just `complete`.

**What changed?**

In the new ORA UI we serially progress through steps, blocking movement forward to a subsequent step until all requirements are met for the current step. This blocks us from determining workflow status directly (as previously, we would jump forward as far as you are able to view, skipping peer if it was followed by an additional step) and requires us to manually determine how far we have progressed.

Previous checks incorrectly determined this by only checking `complete` (whether *you* have finished your required work on a step) instead of also checking `graded` (whether you have received required external grades e.g. from peers).

This adds the additional check that both `complete` and `graded` must be `True` for a step to be considered finished.

**Developer Checklist**

**Reviewer Checklist**

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/__init__.py
+++ b/openassessment/__init__.py
@@ -2,4 +2,4 @@
 Initialization Information for Open Assessment Module
 """
 
-__version__ = '6.0.20'
+__version__ = '6.0.21'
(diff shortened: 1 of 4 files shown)


mattcarter (approved):
Approved.

leangseu-edx (approved):
LGTM 👍

PR #2157, opened 2024-01-05
feat: add handling for peer submission timeout / cancel

nsprenkle:
**TL;DR -** Send the `ERR_SUBMISSION_NO_LONGER_AVAILABLE` error code and a `404` error when trying to submit a grade in the following circumstances:
1. Lease on a peer assessment has expired.
2. Submission that a student was grading has been cancelled.

JIRA: [AU-1647](https://2u-internal.atlassian.net/browse/AU-1647)

**Developer Checklist**

**Testing Instructions**

**Reviewer Checklist**

FYI: @openedx/content-aurora


jansenk:
This actually won't fix the issue, and I think the issue is different than what we think it will be anyways:

Old code:
 - Pass the current submission uuid along with the assessment info
 - Backend calls get_peer_whatever which discards "old" PeerWorkflowItem and grabs a new one
 - passed old uuid doesn't match new uuid and assess fails
 
 We don't pass a submission uuid back to the mfe view. What I think will happen now is:
- Call peer_assess with assessment for old submission
- Old peer workflow item is discarded and we grab a new one
- apply assessment for old submission to new submission

No bueno!! We should probably make a new ticket to address this

nsprenkle:
Okay, will close this PR, awaiting a different fix.

> This actually won't fix the issue, and I think the issue is different than what we think it will be anyways:
> 
> Old code:
> 
> * Pass the current submission uuid along with the assessment info
> * Backend calls get_peer_whatever which discards "old" PeerWorkflowItem and grabs a new one
> * passed old uuid doesn't match new uuid and assess fails
> 
> We don't pass a submission uuid back to the mfe view. What I think will happen now is:
> 
> * Call peer_assess with assessment for old submission
> * Old peer workflow item is discarded and we grab a new one
> * apply assessment for old submission to new submission
> 
> No bueno!! We should probably make a new ticket to address this

PR #2162, opened 2024-01-16
Jkantor/submission check

jansenk:
Don't check for equality in the client and server uuids when the client doesn't provide one


mattcarter (approved):
approved

PR #2176, opened 2024-02-12
fix: date config for mfe

jansenk:
**TL;DR -**  Date config should change the step start / end dates returned by the MFE
 
JIRA: [AU-1677](https://2u-internal.atlassian.net/browse/AU-1677)

**What changed?**

- Refactor step settings serializer to make a call to is_closed to resolve dates with respect to date config type
- Some additional refactors to make the code (imo) a little cleaner
- Tests

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/xblock/ui_mixins/mfe/ora_config_serializer.py
+++ b/openassessment/xblock/ui_mixins/mfe/ora_config_serializer.py
@@ -4,7 +4,6 @@ Serializers for ORA's BFF.
 These are the response shapes that power the MFE implementation of the ORA UI.
 """
 # pylint: disable=abstract-method
-
 from rest_framework.serializers import (
     BooleanField,
     DateTimeField,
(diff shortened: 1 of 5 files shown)


openassessment/xblock/ui_mixins/mfe/ora_config_serializer.py
```diff
@@ -156,42 +145,60 @@ def _get_step(self, rubric_assessments, step_name):
...
+    enableFlexibleGrading = BooleanField(
+        source="enable_flexible_grading", required=False
     )
+
+
+class StaffSettingsSerializer(AssessmentStepSettingsSerializer):
+    STEP_NAME = 'staff-assessment'
+
+
+class AssessmentStepsSettingsSerializer(Serializer):
```
jansenk, line 196: This diff is messy. Here's what I did.

- Changed source to "*" because we needed the fields on the block
- Removed the step name var and just baked that into the class itself, and therefore called each specific settings step class.
- Made all of the settings classes inherit from the abstract assessment step settings class, which just gives them the to_representation and the is_required field, and rather than calling the serializers from the to_representation method, we just get that behavior because they are inheriting.
- Check for the two steps with the due dates we care about, and resolve and insert the correct dates
-

leangseu-edx (approved):
LGTM 👍

PR #2182, opened 2024-02-27
docs: ADR for lightweight extension points

mariajgrimaldi:
**TL;DR -** This PR adds an architectural decision record (ADR) justifying adding lightweight extension points in ORA during the submission lifecycle to solve this feature request: https://openedx.atlassian.net/wiki/spaces/COMM/pages/3891855361/Assessment+step+with+external+tools+-+Turnitin

The ADR talks about sharing students' submission information with external services; in the feature request, that external service is Turnitin, a plagiarism tool. 

**What changed?**

- A new architectural decision record (ADR) was added by the name of:  0003-lightweight-extension-points.rst

**Developer Checklist**

**Testing Instructions**

You could follow these instructions if you want to test what's on the ADR. They are tailored to be used with the Turnitin tool, so you'll need credentials to integrate with it:

1. Install the repositories with the extension points implementations, which allow ORA to share the student's submission data with external services. As mentioned in the feature request, the external service, in this case, is Turnitin. 
```
pip install git+https://github.com/edunext/openedx-events@ora-extension-points
pip install git+https://github.com/edunext/openedx-filters@ora-extension-points
pip install git+https://github.com/edunext/edx-ora2@ora-extension-points
```
2. Configure the Open edX Filter in your environment. The event receiver will work out of the box after the next step.
```
    OPEN_EDX_FILTERS_CONFIG = {
      "org.openedx.learning.ora.submission_view.render.started.v1": {
        "fail_silently": False,
        "pipeline": [
          "platform_plugin_turnitin.extensions.filters.ORASubmissionViewTurnitinWarning",
        ]
      },
    }
```
3. Since we're sharing data with Turnitin, we must install the necessary tools to communicate with the service. Our team implemented a set of APIs in an Open edX plugin to do so, so installing the plugin will make those APIs available:
```
pip install git+https://github.com/eduNEXT/platform-plugin-turnitin.git
```
4. Now, configure your credentials:
```
    TURNITIN_TII_API_URL = "https://xyz-sandbox.com"
    TURNITIN_TCA_INTEGRATION_FAMILY = "xyz"
    TURNITIN_TCA_INTEGRATION_VERSION = "X.Y.Z"
    TURNITIN_TCA_API_KEY = "xyzw"
    TURNITIN_API_TIMEOUT = 30
```
Then, as a student, upload an assignment. Then, with the assignment ID, as an instructor, access the Turnitin API. Please use a bearer token as the authentication method as usual with Open edX APIs.

Generate similarity report for a submission given its submission ID:  
PUT https://{{lms_domain}}/platform-plugin-turnitin/{{course_id}}/api/v1/similarity-report/{{submission_id}}/

Get similarity report status for a submission given its submission ID:
GET https://{{lms_domain}}/platform-plugin-turnitin/{{course_id}}/api/v1/similarity-report/{{submission_id}}/

Get viewer URL for the report given the submission ID:
GET https://{{lms_domain}}/platform-plugin-turnitin/{{course_id}}/api/v1/viewer-url/{{submission_id}}/

**Reviewer Checklist**

FYI: @openedx/content-aurora

### DIFF
--- a/docs/decisions/0003-lightweight-extension-points.rst
+++ b/docs/decisions/0003-lightweight-extension-points.rst
@@ -18,11 +18,12 @@ Decisions
 
 We'll enable external interactions with students' responses by using the `Hooks Extensions Framework`_. Using this approach, we'll implement extension points with minimal modifications in the edx-ora2 repository, making interventions with the students' response lifecycle possible. This is an interim solution for the problem stated above while the community develops a more sophisticated process that covers more use cases. 
 
-This approach includes implementing two extension points:
+This approach includes implementing two extension points, which definitions will live in `openedx-filters`_ and `openedx-events`_ accordingly:
 
-- For the first extension point we will use an `Open edX Filter`_ with the following definition:
+For the first extension point we will use an `Open edX Filter`_ with the following definition:
 
-.. code::
+
+..  code-block:: python
   
   class ORASubmissionViewRenderStarted(OpenEdxPublicFilter):
       """
@@ -44,7 +45,8 @@ This approach includes implementing two extension points:
 
 That will be triggered before `rendering the submission HTML section of the block for the legacy view`_:
 
-.. code::
+
+..  code-block:: python
 
     if path == "legacy/response/oa_response.html":
         try:
@@ -56,9 +58,10 @@ That will be triggered before `rendering the submission HTML section of the bloc
 
 This implementation will allow us to modify what's rendered to the student, via the view context and template, for cases when needed. For example, some third-party services need acknowledgment before receiving users' information.
 
-- The second extension point will be an Open edX Event. The event payload should contain enough information for later processing; in this case, we'll the following event definition:
+The second extension point will be an Open edX Event. The event payload should contain enough information for later processing; in this case, we'll the following event definition:
 
-.. code::
+
+..  code-block:: python
 
     attr.s(frozen=True)
     class ORASubmissionData:
@@ -90,7 +93,8 @@ This implementation will allow us to modify what's rendered to the student, via 
 
 The event will be sent `after a student submits a response to the assessment`_ so it has access to the student's submission key data:
 
-.. code::
+
+..  code-block:: python
 
     @staticmethod
     def send_ora_submission_created_event(submission: dict) -> None:
@@ -111,8 +115,7 @@ The event will be sent `after a student submits a response to the assessment`_ s
             )
         )
 
-     ...
-
+     # Sent after the submission
      self.send_ora_submission_created_event(submission)
 
 This event will allow us to act after a submission is made based on the data sent.
@@ -128,7 +131,8 @@ Extension developers commonly use those extension points in Open edX plugins to 
 
 Let's say you want to add an acknowledgment notice to your submission template so students know their information is being shared with third-party services when submitting a response. The extension developer could implement a `pipeline step`_ for the filter that changes the ``oa_response.html`` template for an ``oa_response_ack_modified.html`` template with its context:
 
-.. code::
+
+..  code-block:: python
 
     from openedx_filters import PipelineStep
     
@@ -157,7 +161,8 @@ Let's say you want to add an acknowledgment notice to your submission template s
 
 See `how to implement pipeline steps`_ for more information. Now, by listening to the `Open edX Event`_, the developer could act on the submission-created notification. Since the event payload has enough information to get the student's submissions, including files, the event receiver can obtain the submission to send it to another service for analysis:
 
-.. code::
+
+..  code-block:: python
 
     from some_plugin.tasks import ora_submission_created_processing_task
 
@@ -192,4 +197,6 @@ customization and addition to the workflow step, it was concluded that the more 
 .. _how to implement pipeline steps: https://docs.openedx.org/projects/openedx-filters/en/latest/how-tos/using-filters.html#implement-pipeline-steps
 .. _how to listen for Open edX Events: https://docs.openedx.org/projects/openedx-events/en/latest/how-tos/using-events.html#receiving-events
 .. _after a student submits a response to the assessment: https://github.com/openedx/edx-ora2/blob/master/openassessment/xblock/ui_mixins/legacy/handlers_mixin.py#L67
-.. _`platform roadmap GH ticket`: https://github.com/openedx/platform-roadmap/issues/253
+.. _platform roadmap GH ticket: https://github.com/openedx/platform-roadmap/issues/253
+.. _openedx-events: https://github.com/openedx/openedx-events
+.. _openedx-filters: https://github.com/openedx/openedx-filters
--- a/docs/decisions/0003-lightweight-extension-points.rst
+++ b/docs/decisions/0003-lightweight-extension-points.rst
@@ -9,16 +9,16 @@ Status
 Context
 *******
 
-Open-ended questions are commonly used in education for assessment purposes, but their flexible nature can facilitate cheating. Therefore, instructors need a way of assessing when cheating happens when reviewing responses. One widely-used solution to this problem in the academic field is online plagiarism tools. These tools receive the students' responses, analyze them, and provide reports that assist instructors in scoring.
+Open-ended questions are commonly used in education for assessment purposes, but their flexible nature can facilitate cheating. Therefore, instructors need a way of assessing when cheating happens when reviewing responses. One widely-used solution to this problem in the academic field is the use online plagiarism tools by instructors. These tools receive the students' responses, analyze them, and provide reports that assist instructors in scoring.
 
-No standard exists for ORA to facilitate integration with tools that process students' answers for plagiarism or other useful reports. Therefore, ORA must allow external interactions with students' responses so they can be sent for analysis.
+No standard exists for ORA to facilitate integration with tools that process students' answers for plagiarism or other useful reports for the course staff. Therefore, ORA must allow external interactions with students' responses so they can be sent for analysis.
 
 Decision
 ********
 
 We'll enable external interactions with students' responses by using the `Hooks Extensions Framework`_. Using this approach, we'll implement extension points with minimal modifications in the edx-ora2 repository, making interventions with the students' response lifecycle possible. This is an interim solution for the problem stated above while the community develops a more sophisticated process that covers more use cases. This approach includes implementing two extension points:
 
-- The first extension point will implemented before rendering the submission HTML section of the block for the legacy view. We'll use an `Open edX Filter`_ with the following definition:
+- The first extension point will implemented before `rendering the submission HTML section of the block for the legacy view`_. We'll use an `Open edX Filter`_ with the following definition:
 
 .. code::
   
@@ -40,9 +40,9 @@ We'll enable external interactions with students' responses by using the `Hooks 
           data = super().run_pipeline(context=context, template_name=template_name, )
           return data.get("context"), data.get("template_name")
 
-This implementation will allow us to modify what's rendered to the student for cases when needed. For example, some third-party services need acknowledgment before receiving users' information.
+This implementation will allow us to modify what's rendered to the student, via the view context and template, for cases when needed. For example, some third-party services need acknowledgment before receiving users' information.
 
-- The second extension point will be an Open edX Event sent after a student submits a response to the assessment. The event payload should contain enough information for later processing; in this case, we'll the following event definition:
+- The second extension point will be an Open edX Event sent `after a student submits a response to the assessment`_. The event payload should contain enough information for later processing; in this case, we'll the following event definition:
 
 .. code::
 
@@ -83,7 +83,7 @@ Extension developers commonly use those extension points in Open edX plugins to 
 - Change the template that is rendered to the student
 - Send students' submission data to another service
 
-Let's say you want to add an acknowledgment notice to your submission template so students know their information is being shared with third-party services when submitting a response. The extension developer could implement a pipeline step for the filter that changes the ``oa_response.html`` template for an ``oa_response_ack_modified.html`` template with its context:
+Let's say you want to add an acknowledgment notice to your submission template so students know their information is being shared with third-party services when submitting a response. The extension developer could implement a `pipeline step`_ for the filter that changes the ``oa_response.html`` template for an ``oa_response_ack_modified.html`` template with its context:
 
 .. code::
 
@@ -109,10 +109,10 @@ Let's say you want to add an acknowledgment notice to your submission template s
             """
             return {
                 "context": context,
-                "template_name": "some_plugin/oa_response_ack_modified.html",
+                "template_name": "some_plugin/oa_response_with_acknowledgement.html",
             }
 
-By listening to the Open edX Event, the developer could act on the submission-created notification. Since the event payload has enough information to get the student's submissions, including files, the event receiver can obtain the submission to send it to another service for analysis:
+See `how to implement pipeline steps`_ for more information. Now, by listening to the `Open edX Event`_, the developer could act on the submission-created notification. Since the event payload has enough information to get the student's submissions, including files, the event receiver can obtain the submission to send it to another service for analysis:
 
 .. code::
 
@@ -131,7 +131,7 @@ By listening to the Open edX Event, the developer could act on the submission-cr
             submission.file_downloads,
         )
 
-Extension developers could interact with an essential part of the student's assessment lifecycle with these changes. But when none of these extension points are configured for use, then ORA assessments will behave as usual.
+See `how to listen for Open edX Events`_ for more information. Extension developers could interact with an essential part of the student's assessment lifecycle with these changes. But when none of these extension points are configured for use, then ORA assessments will behave as usual.
 
 Rejected Alternatives
 *********************
@@ -140,4 +140,11 @@ As suggested in the `platform roadmap GH ticket`_ for this feature, the team res
 customization and addition to the workflow step, it was concluded that the more straightforward solution was implementing a lightweight extension mechanism. 
 
 .. _Hooks Extensions Framework: https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0050-hooks-extension-framework.html
+.. _rendering the submission HTML section of the block for the legacy view: https://github.com/openedx/edx-ora2/blob/master/openassessment/xblock/ui_mixins/legacy/views/submission.py#L19
+.. _Open edX Filter: https://docs.openedx.org/projects/openedx-filters/en/latest/
+.. _Open edX Event: https://docs.openedx.org/projects/openedx-filters/en/latest/
+.. _pipeline step: https://docs.openedx.org/projects/openedx-filters/en/latest/concepts/glossary.html#pipeline-steps
+.. _how to implement pipeline steps: https://docs.openedx.org/projects/openedx-filters/en/latest/how-tos/using-filters.html#implement-pipeline-steps
+.. _how to listen for Open edX Events: https://docs.openedx.org/projects/openedx-events/en/latest/how-tos/using-events.html#receiving-events
+.. _after a student submits a response to the assessment: https://github.com/openedx/edx-ora2/blob/master/openassessment/xblock/ui_mixins/legacy/handlers_mixin.py#L67
 .. _`platform roadmap GH ticket`: https://github.com/openedx/platform-roadmap/issues/253
--- a/docs/decisions/0003-lightweight-extension-points.rst
+++ b/docs/decisions/0003-lightweight-extension-points.rst
@@ -1,64 +1,99 @@
 Lightweight extension points
 ############################
 
-.. The title should be a short noun phrase. For example, "1. Django IDA" or "9. LDAP for Multitenant Integration"; filename should be lowercase with punctuation removed and spaces replaced by dash
-
 Status
-
 ******
 
 **Draft** *2024-02-27*
 
 Context
-
 *******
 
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @mariajgrimaldi! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation

This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,202 @@
...
+            )
+        )
+
+     # Sent after the submission
+     self.send_ora_submission_created_event(submission)
+
+This event will allow us to act after a submission is made based on the data sent.
+
+Consequences
+************
```
mariajgrimaldi, line 124: Note to author: we might need adding a little more specifications about Turnitin? Or the _external services_ generic references are enough?

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,202 @@
...
+      Custom class used to create ORA submission view filters and its custom methods.
+      """
+  
+      filter_type = "org.openedx.learning.ora.submission_view.render.started.v1"
+  
+      @classmethod
+      def run_filter(cls, context: dict, template_name: str):
+          """
+          Execute a filter with the signature specified.
+          Arguments:
```
BryanttV, line 39:
```suggestion
          Execute a filter with the signature specified.

          Arguments:
```

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,202 @@
...
+The event will be sent `after a student submits a response to the assessment`_ so it has access to the student's submission key data:
+
+
+..  code-block:: python
+
+    @staticmethod
+    def send_ora_submission_created_event(submission: dict) -> None:
+        """
+        Send an event when a submission is created
+        Args:
```
BryanttV, line 103:
```suggestion
        Send an event when a submission is created

        Args:
```

docs/decisions/0003-lightweight-extension-points.rst
BryanttV: I think that keeping it generic would be best since we could use any other third-party tool through the extension points.

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,202 @@
+Lightweight extension points
+############################
+
+Status
+******
+
+**Draft** *2024-02-27*
+
+Context
```
felipemontoya, line 9: I think the context described here is focusing too much on the particular integration that we are looking to build on top of an extension mechanism. Given the experience we have had trying to modify ORA2 so far the issue is that the needs of different users like edx.org, the spanish universitites, other on campus intiatives and other members of the community don't align.

I propose we focus the ADR towards the issues created by not having any extension mechanism that does not require forking the code. By this I mean, ORA is in itself designed to be extended, it is possible to create new steps and configure them in, however this requires essencially modifiying the core of the ORA repo a.ka. forking it.

We have at least two examples one being the code we wrote for the spanish campus uses and the other being RG's ai-ora2 work.

I propose something like this for the context:


>   Open Responses are commonly used in education for assessment purposes and the flexibility of ORA has made it a key feature of the Open edX platform. However as developers it is very hard to accomodate the needs of educators and other stakeholders in the teaching process when they want to enhance the learners experience with other tools such as plagiarism detection, AI grading, coding graders and others.
> 
>   As the code is open it is always possible to modify it incurring in a technical mainteinance debt that for the most part either prevents educators from actually experimenting or prevents the instances to upgrade to newer versions once the changes to ORA become important in the teaching process.

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,202 @@
...
+*********
+
+We'll enable external interactions with students' responses by using the `Hooks Extensions Framework`_. Using this approach, we'll implement extension points with minimal modifications in the edx-ora2 repository, making interventions with the students' response lifecycle possible. This is an interim solution for the problem stated above while the community develops a more sophisticated process that covers more use cases. 
+
+This approach includes implementing two extension points, which definitions will live in `openedx-filters`_ and `openedx-events`_ accordingly:
+
+For the first extension point we will use an `Open edX Filter`_ with the following definition:
+
+
+..  code-block:: python
```
felipemontoya, line 26: To make the ADR more readable I suggest we move the code to a second file with code examples that we can reference from this text, but keep the conceptual part as the center of this file.

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,202 @@
...
+        ora_submission_created_processing_task.delay(
+            submission.id,
+            submission.file_downloads,
+        )
+
+See `how to listen for Open edX Events`_ for more information. 
+
+Extension developers could interact with an essential part of the student's assessment lifecycle with these changes. But when none of these extension points are configured for use, then ORA assessments will behave as usual.
+
+Rejected Alternatives
```
felipemontoya, line 186: Given that there is currently no other option for extending ORA without a fork we are not really rejecting any other alternative at this point. It could be argued that we are rejecting the construction of a larger (or more comprehensive) framework for extension but I don't see it like that. It is more like this is the first step towards a larger framework. If we were to propose a project to extend ORA with a mechanism for dependency injection we would still propose it to be build on top of the hooks framework.

At this ADR we are only commiting to the first few hooks because we understand very well the effort that it requires. However there is no technical limit for this proposal to grow into more hooks and to eventually support a broad array of extension use cases. Maybe this reflection is more for the consequences part. We are opening the door to more extension points in the future that are also built using events and filters and we can make it more explicit.

felipemontoya (changes requested):
Thanks a lot for starting this @mariajgrimaldi. I left comments on the main sections but since those can disapear once we change the code I'll leave a short summary here.

The current form of the ADR is focused on the use case we have at hand and I believe that is less important than the goal of moving the needle from the world we leave now where extending ORA requires modifying the core code, which is difficult and creates maintenance issues. I propse we the ADR on the advocacy for a more extensible ORA using the existing Hooks Extensions Framework, allowing for external interactions with minimal code changes so that the level of effort is something we can commit to in the near future.

mariajgrimaldi:
Hi @felipemontoya, thanks for the review. I'll be including your suggestions as soon as possible.

docs/decisions/0003-lightweight-extension-points.rst
mariajgrimaldi: I was researching a bit about where the code for the proposal should live and found this: 

https://docs.openedx.org/projects/openedx-proposals/en/latest/best-practices/oep-0019-bp-developer-documentation.html#adrs

> Architecture Decision Records (ADRs) are local lightweight documents of technical decisions co-located with their corresponding code...

So that's why I'm trying to illustrate what this will look like, but in reality, we should use the implementation as examples. I propose removing the code altogether without adding a new file, and then, if needed, we add references to the actual code implemented. What do you think?

docs/decisions/0003-lightweight-extension-points.rst
mariajgrimaldi: I wasn't sure about what to put here, so I'll use this comment instead. Thanks.

docs/decisions/0003-lightweight-extension-points.rst
mariajgrimaldi: Agree. Thanks!

mariajgrimaldi:
I believe I addressed all your comments @felipemontoya, could you check again? Thanks.

docs/decisions/0003-lightweight-extension-points.rst
mariajgrimaldi: I see. I already included the suggestions. Thanks for the feedback!

mariajgrimaldi:
Hello folks, @pomegranited, @GlugovGrGlib, @e0d. I'm tagging you here as candidates to review this extension proposal if you can do so. The PR cover letter has all the info you might need; feel free to ask for any clarification. Thanks!

pomegranited:
Very interesting thank you @mariajgrimaldi ! I'll take a closer look next week.

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,51 @@
...
+
+The first extension point we will implement is an `Open edX Filter`_ that will be executed before `rendering the submission HTML section of the block for the legacy view`_, with input arguments ``context`` and ``template path``. This implementation will allow us to modify what's rendered to the student, via the view ``context`` and ``template``, for cases when needed. 
+
+The second extension point to be implemented is an `Open edX Event`_. It will be sent `after a student submits a response to the assessment`_ with the student's submission key data, like the ORA submission ID and files uploaded in the submission, as the event's payload. This event (that works like a notification) will allow us to take action after a submission is made based on the data sent.
```
e0d, line 23: Are copies of the files included in the event or URLs?

docs/decisions/0003-lightweight-extension-points.rst
e0d: I suppose an alternative approach would be to make adding grading steps much more robust, using inversion of control and dependency injection to add steps rather than forks.  That would likely be significantly more work and would add another divergent pattern to events which we are using elsewhere.  A couple of general questions.  Could we imagine a solution where an event could be added to a function without changing the code?  In other languages than python you might do this with aspects, AOP point cuts or similar.  Not sure this is a good idea, so not advocating for it currently, but it would be extremely flexible.

docs/decisions/0003-lightweight-extension-points.rst
mariajgrimaldi: In [this](https://github.com/openedx/edx-ora2/pull/2182#discussion_r1521546894) comment, I mentioned what we plan on using as the events payload. Is it worth adding to the ADR the reference implementation I mentioned in the comment?

pomegranited: I can see how the first pre-submission filter gives the opportunity to inject a EULA statement, which we assume was seen by the student before they submit. And I can see how the second "submission created" event gives the chance to pass that submission on to an external assessment tool like Turnitin. Ok.

But shouldn't the results of this assessment should be recorded as part of the ORA assessment pipeline? How do instructors see which submissions were flagged by the external assessment tool? Do they have to view the similarity report for each submission and use that when making their Staff assessment? If so, I assume this is done by extending the ORA2 Grading MFE to support showing links to the [Turnitin plugin's API views](https://github.com/eduNEXT/platform-plugin-turnitin/blob/main/platform_plugin_turnitin/api/v1/views.py), so that's a third extension point.

Each of these integration points seem OK to add -- they are indeed lightweight, and won't affect other users of ORA2. But I'm having trouble agreeing that all together, they are the best way to implement such a broadly useful feature.

Is it very difficult to define a new "External Tool" assessment step in the ORA2 pipeline that could work for Turnitin and other external tools?

felipemontoya: Thanks for reviewing this.

> But shouldn't the results of this assessment should be recorded as part of the ORA assessment pipeline?

There is some nuance to that. Every tool is different and the tool we have implemented (turnitin) does not give you an assesment, what it does is create a report saying "from our analysis this much is original and this much is not" which requires a human evaluator to look at this. The similarity report would for instance flag as "not original" a piece of text you are making a citation from, which is a completely valid thing to do.

> How do instructors see which submissions were flagged by the external assessment tool? Do they have to view the similarity report for each submission and use that when making their Staff assessment? If so, I assume this is done by extending the ORA2 Grading MFE to support showing links to the [Turnitin plugin's API views](https://github.com/eduNEXT/platform-plugin-turnitin/blob/main/platform_plugin_turnitin/api/v1/views.py), so that's a third extension point.

What we are working towards is that this report is available for authors during the grading step in the MFE. That would indeed be another extension point. This time in the ora-grading-MFE. We are working towards having frontend-pluggability in there, as part of the effort that Adolfo is leading together with 2U in the `frontend-plugin-framework` repo.

> But I'm having trouble agreeing that all together, they are the best way to implement such a broadly useful feature.
Is it very difficult to define a new "External Tool" assessment step in the ORA2 pipeline that could work for Turnitin and other external tools?

It is quite a lot of work yes. Specially trying to make that step configurable and flexible for different and varied use cases. For instance an "external tool assessment step" would not be helpful for our usecase as turnitin does not give an assessment, it only presents the staff user extra info to make an informed choice about the grade they will grant.

We initially started down that path by doing several POCs [PRs](https://github.com/eduNEXT/edx-ora2/pulls) about different things we wanted to extend ora with. We also studied an [implementation](https://github.com/raccoongang/ai-ora2/pull/1/files) RG made for an AI grader. Sadly we could not take on that much scope in the context we are at and that is when we decided to propose something very lightweight to get started. Naturally this 2 hooks won't be enough for all cases, but we can build more such hooks incrementally instead of leaving ora more time without an extensibility mechanism.

pomegranited: @felipemontoya Thank you for talking me through your use case and the various other paths attempted here.

> It is quite a lot of work yes. Specially trying to make that step configurable and flexible for different and varied use cases.

I'd like that to be a goal, but I understand that resources are limited. Maybe once we have some more use cases, then we can look at  bringing some of your plugin implementation into the ORA2 core?

pomegranited: @mariajgrimaldi Do you mind to mention "contributing an External Tool step to ORA2" as a rejected alternative here due to time constraints? Otherwise this looks good to me :)

felipemontoya: Absolutely. I would like to pursue the goal of having more flexibility in ora, but it was important to acknowledge now that we can't pursue it at the moment in the context of the turnitin extension we are writing for the spanish universities. However we wanted to make the ADR to discuss the posibility of bringing some extensibility now, even as we call it, a lightweight extension.

pomegranited:
Hi @mariajgrimaldi  :) FYI I'm only waiting on [this comment](https://github.com/openedx/edx-ora2/pull/2182#discussion_r1531248440) to be addressed, and then this is good to merge.

mariajgrimaldi:
@pomegranited @e0d: Thanks for the patience. I'll address your comments later today or early tomorrow. Thanks again!

docs/decisions/0003-lightweight-extension-points.rst
mariajgrimaldi: > Could we imagine a solution where an event could be added to a function without changing the code?

That's an approach that we haven't explored yet. The closest example I've seen is how people have implemented filters as decorators when used in multiple places (here's an example: https://github.com/openedx/event-routing-backends/blob/master/event_routing_backends/processors/openedx_filters/decorators.py#L9). This approach looks cleaner and is easier to maintain than duplicating code. However, I think we might be missing the explicitness of what's been processed and where for more complex functions. I imagine there's a way to solve that. It would be interesting to look further into it!

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,52 @@
+Lightweight extension points
+############################
+
+Status
+******
+
+**Draft** *2024-02-27*
```
pomegranited, line 7: Before merge: mark this as Accepted.

pomegranited (approved):
:+1: @mariajgrimaldi Thank you for sharing the conclusions of your research work in this ADR, and for coming up with such a nice, lightweight way to provide this functionality. 

Just need to mark this ADR as "Approved", and I'll merge.

- [x] <strike>I tested this</strike> N/A
- [x] I read through the ADR and checked the links.

mariajgrimaldi:
I marked the ADR as accepted, @pomegranited; thank you for the review and all the help!

docs/decisions/0003-lightweight-extension-points.rst
```diff
@@ -0,0 +1,51 @@
...
+
+As the code is open, it is always possible to modify it. Still, it might result in a technical maintenance debt, which usually prevents educators from experimenting or installations from upgrading to newer versions once the changes to ORA become important in the teaching process.
+
+Decisions
+*********
+
+As the first step towards making implementing new use cases during the ORA user's lifecycle more flexible, we will introduce two extension points in the ORA codebase. These extension points will be built on top of the `Hooks Extensions Framework`_. The definitions for these extension points will reside in `openedx-filters`_ and `openedx-events`_. These definitions will be imported into the edx-ora2 repository and triggered in two places during the learners' lifecycle, resulting in minimal modifications to the ORA implementation.
```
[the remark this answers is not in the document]
e0d: This is very helpful, but I'm not sure I understand what purpose the filter plays here.  Is it to display the result from TurnItIn if it exists to the instructor grading the work?

docs/decisions/0003-lightweight-extension-points.rst
e0d: > There is some nuance to that. Every tool is different and the tool we have implemented (turnitin) does not give you an assesment, what it does is create a report saying "from our analysis this much is original and this much is not" which requires a human evaluator to look at this. The similarity report would for instance flag as "not original" a piece of text you are making a citation from, which is a completely valid thing to do.

So I'm clear, this means that this approach, events and filters, cannot be used for external tools that do return a grade and might need to block the ORA workflow if the result warranted that?

I understand the scope and complexity concerns here.  This solution is nice in that it is plugable.  However, if it can't support grading or interacting with the overall workflow I want to be explicit about that.

docs/decisions/0003-lightweight-extension-points.rst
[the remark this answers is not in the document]
mariajgrimaldi: A warning must be displayed before sharing information with external tools so the user knows what's being submitted will be shared with another service. In this case, this is done by implicitly accepting an EULA agreement before sending the information. Now, if the user disagrees with it, it'd be enough not to submit their assessment. We're adding this warning by modifying the ORA template in a plugin:
image

But this filter could be used for another type of use case where the submission view needs modification or extra functionality; we went for implicit compliance for time constraints, but it could be something even more complex.

docs/decisions/0003-lightweight-extension-points.rst
mariajgrimaldi: With these two hooks, feedback for the user cannot be done. I'll include it in the ADR. 

However, way back I did a rough POC that considered giving feedback to the user:

```
   @togglable_mobile_support
   def student_view(self, context=None):  # pylint: 
       try:
           self.update_workflow_status()
       except AssessmentWorkflowError:
           # Log the exception, but continue loading the page
           logger.exception('An error occurred while updating the workflow on page load.')

       ui_models = self._create_ui_models()
       # All data we intend to pass to the front end.
       context_dict = {
           "title": self.title,
           "prompts": self.prompts,
           "prompts_type": self.prompts_type,
           "rubric_assessments": ui_models,
           "show_staff_area": self.is_course_staff and not self.in_studio_preview,
       }
       template_name = "openassessmentblock/oa_base.html"

       try:
           # .. filter_implemented_name: ORAStudentViewRenderStarted
           # .. filter_type: org.openedx.learning.dashboard.render.started.v1
           template_name, context_dict = ORAStudentViewRenderStarted.run_filter(
             	template_name=template_name,
             	context=context_dict
           )
       except ORAStudentViewRenderStarted.RenderInvalidTemplate as exc:
           template_name, context_dict = (exc.template, exc.template_context)
       except ORAStudentViewRenderStarted.RenderCustomFragment as exc:
           fragment = exc.fragment
       else:
           template = get_template(template_name)
           fragment = self._create_fragment(template, context_dict, initialize_js_func='OpenAssessmentBlock')

       template = get_template(template_name)
       return self._create_fragment(template, context_dict, initialize_js_func='OpenAssessmentBlock')

class ORAStudentViewRenderStarted(OpenEdxPublicFilter):

   filter_type = "org.openedx.learning.dashboard.render.started.v1"

   class RenderInvalidTemplate(OpenEdxFilterException):

       def __init__(self, message, dashboard_template="", template_context=None):
           super().__init__(
               message,
               dashboard_template=dashboard_template,
               template_context=template_context,
           )

   class RenderCustomFragment(OpenEdxFilterException):

       def __init__(self, message, fragment=None):
           super().__init__(
               message,
               fragment=fragment,
           )

   @classmethod
   def run_filter(cls, context, template_name):
       data = super().run_pipeline(context=context, template_name=template_name)
       return data.get("context"), data.get("template_name")
```

With that filter, we change the context and the rendered template (openassessmentblock/oa_base.html). Thus, we can add a button with a handler that calls Turnitin with the response value.

The template path we will add will be a template that lives in a platform-plugin-turnitín. This plugin will have the same content as the original template (openassessmentblock/oa_base.html), but we will add the `Check with Turnitin` button, which, when pressing it, will send the information to turnitín. Here you can see the button:

image

When clicked, an HTTP request will be made to Turnitín with the content of my answer. Then, a mechanism should be able to get the info to show it to the user. I didn't get to that, though.

openedx-webhooks:
@mariajgrimaldi 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/edx-ora2&entry.867055334=https://github.com/openedx/edx-ora2/pull/2182&entry.1484655318=https://github.com/mariajgrimaldi&entry.752974735=2024-02-27+14:11&entry.1917517419=2024-04-04+23:05&entry.2133058324=Yes) so we can improve your experience in the future.

PR #2191, opened 2024-03-13
test: add test for flexible edge case

jansenk:
when a learner recieves the neccessary number of peer grades, but takes more than seven days to complete thir own peer reviews, we still reduce the number of grades used to calculate the final score, even though more are available

JIRA: https://2u-internal.atlassian.net/browse/AU-1871

This is just a test to capture the behavior we're interested in. A fix will be addressed in [AU-1904
](https://2u-internal.atlassian.net/browse/AU-1904)
FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/assessment/test/test_peer.py
+++ b/openassessment/assessment/test/test_peer.py
@@ -9,6 +9,7 @@ import pytz
 
 from django.db import DatabaseError, IntegrityError
 from django.utils import timezone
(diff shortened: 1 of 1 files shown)


nsprenkle (approved):

nsprenkle:
@jansenk , do you have the fix ticket groomed yet? Once that is linked, I think this is good to go.

PR #2195, opened 2024-03-19
feat: use available peer reviews even if flexible grading is used

jansenk:
**TL;DR -** When we are using flexible grading and reducing the number of required peer grades, we should still use as many grades as possible, up to the full required amount. If we have more peer grades than the absolute minimum required, still use them, but don't use more than "must be graded by"

JIRA: [AU-1904](https://2u-internal.atlassian.net/browse/AU-1904)

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/assessment/api/peer.py
+++ b/openassessment/assessment/api/peer.py
@@ -37,6 +37,20 @@ def flexible_peer_grading_enabled(peer_requirements, course_settings):
     return peer_requirements.get("enable_flexible_grading")
 
 
+def flexible_peer_grading_active(submission_uuid, peer_requirements, course_settings):
+    """
+    Is flexible peer grading on, and has enough time elapsed since submission to enable it?
+    """
+    if not flexible_peer_grading_enabled(peer_requirements, course_settings):
+        return False
+
+    submission = sub_api.get_submission(submission_uuid)
+    # find how many days elapsed since subimitted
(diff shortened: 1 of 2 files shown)


BrandonHBodine (approved):
lgtm. I don't have the full context for the scope of the project though.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "edx-ora2",
-  "version": "6.5.0",
+  "version": "6.5.1",
```
nsprenkle, line 3: Note to also update this in `package-lock.json`.

openassessment/assessment/api/peer.py
```diff
@@ -244,9 +249,19 @@ def get_score(submission_uuid, peer_requirements, course_settings):
 
     # Check if enough peers have graded this submission
     num_required_peer_grades = required_peer_grades(submission_uuid, peer_requirements, course_settings)
-    if items.count() < num_required_peer_grades:
+    num_recieved_peer_grades = items.count()
+    if num_recieved_peer_grades < num_required_peer_grades:
```
nsprenkle, line 253: Nit: the order / comments of these is confusing. It isn't clear to me on first read that `num_required_peer_grades` is not the actual configured amount but can morph if flex grading is active. Maybe putting the flex grading check above this or extending the comment to that effect would be helpful.

nsprenkle (approved):
Looks good, though I have nits on how to make this read better.

PR #2196, opened 2024-03-21
feat: implement grading strategies for peer evaluations

mariajgrimaldi:
**TL;DR -** This PR adds a new grading calculation strategy that can be configured in the assessment steps configurations for peer steps instead of the median calculation.

Currently, for steps that calculate a final score based on multiple scores (e.g. peer steps) do so using the median by default; the calculation is done [here](https://github.com/openedx/edx-ora2/blob/a2f652ac6aa1a0758bacc4832a1a52445229059e/openassessment/assessment/models/base.py#L491-L520). Using the median for this use case has its advantages. However, using it by default without reconfigurability options removes other use cases that can be done using the mean calculation for the final score instead. This also opens the door for other grading strategies mentioned in [this thread](https://openedx.slack.com/archives/C014VAUJUJ0/p1695241319300679).

That's why [this product proposal](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3891462181/Configurable+grading+strategy+for+peer+assessments) was created and later approved. 

**What changed?**

- The students' score is obtained before setting the score [here](https://github.com/openedx/edx-ora2/blob/a2f652ac6aa1a0758bacc4832a1a52445229059e/openassessment/workflow/models.py#L439-L444)
- Each step API has its own `get_score` function
- The peer step `get_score` gets the median score of all scores the student has at the moment of generation by using [get_assessments_median_scores](https://github.com/openedx/edx-ora2/blob/a2f652ac6aa1a0758bacc4832a1a52445229059e/openassessment/assessment/api/peer.py#L264-L267)
- The `get_assessments_median_scores` calls `get_median_scores_dict`, which calculates the final score for the student using the median.

So we can make the grading strategy configurable we:

Changed each reference to `get_assessments_median_scores` for a more generic `get_assessments_peer_scores_with_grading_strategy` which discriminates based on the workflow configuration which calculation to use. If the feature flag `ENABLE_ORA_PEER_CONFIGURABLE_GRADING` is off, then the default `get_assessments_median_scores` is used.

By default, the median calculations are used.

**Developer Checklist**

**Testing Instructions**

1. Clone this branch into your file system
2. Install edx-ora2 as an editable requirement (-e optional)
3. Now, add an ORA component to your course. 
4. Turn on the feature by adding: `FEATURES["ENABLE_ORA_PEER_CONFIGURABLE_GRADING"] = True`
5. Configure the steps as follows so these tests instructions make sense:
image
6. Use just one rubric configured with these option points:
image
7. The median and the mean look alike when the dataset is not skewed. So, so we can check we're using the mean instead of the median, let's use the following scoring sequence:
Student 1 submits the ORA activity with a distinctive answer so we can distinguish it later
Student 2 scores Student 1 with 0
Student 3 scores Student 1 with 1
Student 4 scores Student 1 with 5
Now, the median is: 1 but the mean is: 2. So check that grading matches the mean value and the explanation says mean instead of median.
image

If you don't configure the mean grading then the default behavior is kept:
image

**Reviewer Checklist**


openedx-webhooks:
Thanks for the pull request, @mariajgrimaldi! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation

This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

openassessment/assessment/api/peer.py
```diff
@@ -480,6 +492,44 @@ def get_rubric_max_scores(submission_uuid):
         logger.exception(error_message)
         raise PeerAssessmentInternalError(error_message) from ex
 
+def get_peer_assessment_scores(submission_uuid, score_type="median"):
```
mariajgrimaldi, line 495: note to author: can we change this name?

openassessment/assessment/models/base.py
```diff
@@ -488,6 +488,22 @@ def create(cls, rubric, scorer_id, submission_uuid, score_type, feedback=None, s
...
+        Args:
+            scores_dict (dict): A dictionary of lists of int values. These int values
+                are reduced to a single value that represents the median.
+            grading_strategy (str): The type of score to calculate.  Defaults to "median".
+
+        Returns:
+            (dict): A dictionary with criterion name keys and median score
+                values.
+        """
+        return getattr(cls, f"get_{grading_strategy}_score_dict")(scores_dict)
```
mariajgrimaldi, line 505: note to author: should we manage unknown grading strategy names?

mariajgrimaldi:
I'm still missing generating all the translations so tests don't fail and adding support for the ORA data reports. Meanwhile, I'll open the PR for review. Thank you!

openassessment/assessment/models/base.py
```diff
@@ -518,6 +534,36 @@ def get_median_score_dict(cls, scores_dict):
...
+    @classmethod
+    def get_mean_score_dict(cls, scores_dict):
+        """Determine the mean score in a dictionary of lists of scores
+
+        For a dictionary of lists, where each list contains a set of scores,
+        determine the mean value in each list.
+
+        Args:
+            scores_dict (dict): A dictionary of lists of int values. These int
+                values are reduced to a single value that represents the median.
```
BryanttV, line 546:
```suggestion
                values are reduced to a single value that represents the mean.
```

openassessment/assessment/models/base.py
```diff
@@ -518,6 +534,36 @@ def get_median_score_dict(cls, scores_dict):
...
+        Examples:
+            >>> scores = {
+            >>>     "foo": [1, 2, 3, 4, 5],
+            >>>     "bar": [6, 7, 8, 9, 10]
+            >>> }
+            >>> Assessment.get_mean_score_dict(scores)
+            {"foo": 3, "bar": 8}
+
+        """
+        median_scores = {}
```
BryanttV, line 561:
```suggestion
        mean_scores = {}
```

openassessment/assessment/models/base.py
```diff
@@ -518,6 +534,36 @@ def get_median_score_dict(cls, scores_dict):
...
+            (dict): A dictionary with criterion name keys and mean score
+                values.
+
+        Examples:
+            >>> scores = {
+            >>>     "foo": [1, 2, 3, 4, 5],
+            >>>     "bar": [6, 7, 8, 9, 10]
+            >>> }
+            >>> Assessment.get_mean_score_dict(scores)
+            {"foo": 3, "bar": 8}
```
BryanttV, line 558: Could we use another example to differentiate the mean from the median?

openassessment/assessment/models/base.py
```diff
@@ -552,6 +598,30 @@ def get_median_score(scores):
...
+        Examples:
+            >>> scores = 1, 2, 3, 4, 5]
+            >>> Assessment.get_median_score(scores)
+            3
+
+        """
+        total_criterion_scores = len(scores)
+        if total_criterion_scores == 0:
+            return 0
+        return int(math.ceil(sum(scores) / float(total_criterion_scores)))
```
BryanttV, line 623: Should we always round down? or would it be better to round up the nearest integer?

openassessment/assessment/models/base.py
```diff
@@ -552,6 +598,30 @@ def get_median_score(scores):
             )
         return median_score
 
+    @staticmethod
+    def get_mean_score(scores):
```
BryanttV, line 602: Please, update the docstring

openassessment/xblock/grade_mixin.py
```diff
@@ -301,7 +302,10 @@ def has_feedback(assessments):
         if staff_assessment:
             median_scores = staff_api.get_assessment_scores_by_criteria(submission_uuid)
         elif "peer-assessment" in assessment_steps:
-            median_scores = peer_api.get_assessment_median_scores(submission_uuid)
+            median_scores = peer_api.get_assessment_scores_with_grading_strategy(
```
BryanttV, line 305: The name of this variable would no longer be quite correct, but I understand that if we update it it would later affect the code.

openassessment/templates/legacy/edit/oa_edit_peer_assessment.html
```diff
@@ -54,6 +54,18 @@
                         </p>
                     {% endif %}
                 </li>
+                {% if enable_peer_configurable_grading %}
```
BryanttV, line 57: Should we allow the grading strategy to be changed after the grading has already been calculated? I don’t know if this could generate some kind of inconsistency.

openassessment/assessment/models/base.py
mariajgrimaldi: Yes, that was what was determined in the product proposal: https://openedx.atlassian.net/wiki/spaces/COMM/pages/3891462181/Configurable+grading+strategy+for+peer+assessments

openassessment/templates/legacy/edit/oa_edit_peer_assessment.html
mariajgrimaldi: Thank you. You raised a valid concern. From an implementation point of view, the calculations happen automatically for each criterion but not for the final grade, so that would be confusing for end-users. What I think we can do is avoiding changing the grading strategy after the problem is released.

openassessment/xblock/grade_mixin.py
mariajgrimaldi: Yes, I'll need to refactor so that more generic variables can be used (the median_scores repeat quite a few times).

mariajgrimaldi (commented):
@BryanttV: I think I addressed all your comments. Could you review again? 

I still need to fix the JS tests, so I'll work on that. Thanks!

openassessment/templates/legacy/edit/oa_edit_peer_assessment.html
mariajgrimaldi: Done! Thank you again

BryanttV:
Hi @mariajgrimaldi, I have tested again and everything seems to work correctly. Only a minor change would be needed. Thank you!

openassessment/assessment/models/base.py
```diff
@@ -552,6 +598,28 @@ def get_median_score(scores):
...
+        Examples:
+            >>> scores = [5, 6, 12, 16, 22, 53]
+            >>> Assessment.get_mean_score(scores)
+            19
+
+        """
+        total_criterion_scores = len(scores)
+        if total_criterion_scores == 0:
+            return 0
+        return int(math.ceil(sum(scores) / float(total_criterion_scores)))
```
BryanttV, line 621: `math.ceil` already returns an integer.

```suggestion
        return math.ceil(sum(scores) / float(total_criterion_scores))
```

mariajgrimaldi: Done. Thank you!

mariajgrimaldi:
Hello, @pomegranited! I hope you're doing well. Could you give us a hand reviewing this PR? We'd appreciate it.

Regarding the test failures, I'll be working on increasing the PR coverage this week. I hope to update the PR soon. Thanks!

openassessment/assessment/models/base.py
```diff
@@ -552,6 +598,28 @@ def get_median_score(scores):
...
+        Examples:
+            >>> scores = [5, 6, 12, 16, 22, 53]
+            >>> Assessment.get_mean_score(scores)
+            19
+
+        """
+        total_criterion_scores = len(scores)
+        if total_criterion_scores == 0:
+            return 0
+        return math.ceil(sum(scores) / float(total_criterion_scores))
```
pomegranited, line 621: I got an unexpected mean score:

image

(it's the second default rubric, so poor=0, fair=1, good=3, excellent=5):
1 + 0 + 3 + 1 + 5 = 10, 10/5 =2

Is it because there's no `2` in the rubric, so it rounds up to the next one `3`?

pomegranited (approved):
Hi @mariajgrimaldi , thank you for this feature! I think it'll be useful for content authors.

I noted a few minor nits, and had a question about rounding, but nothing that's blocking merge. If you agree with my suggested changes, could you apply them and do a version bump? I can merge tomorrow.

:+1:

- [x] I tested this on my tutor dev stack with:
  * the flag disabled (by default) -- no option provided, defaulted to median
  * the flag enabled -- options provided, changed to "mean", was not able to change after block was published
- [x] I read through the code
- [x] I checked for accessibility issues by using my keyboard to navigate the ORA editor.
- [x] Includes documentation
- [x] User-facing strings are extracted for translation

openassessment/assessment/models/base.py
mariajgrimaldi: This threw me for a loop. So I tried replicating the same scenario. Here's the criteria I used (defaults): 
Ideas: poor=0, fair=3, good=5
Content: poor=0, fair=1, good=3, excellent=5
Graded by 5 configuration
And got:
image
 
While researching, I found that the assessments are [retrieved, ordered, and then truncated by the number of peer grades needed](https://github.com/openedx/edx-ora2/blob/master/openassessment/assessment/api/peer.py#L243-L248). Could you confirm that the number of peer grades configured was 5?

mariajgrimaldi: I also added an extra test here with the exact values: https://github.com/openedx/edx-ora2/pull/2196/files#diff-7fe9b2681acaf897f2dd62a13553badf288fd6144832ce0562c4a58bd2640e7fR2381

openassessment/assessment/models/base.py
mariajgrimaldi: I added an assert so it's clear. Thanks for the suggestion!

openassessment/assessment/models/base.py
mariajgrimaldi: I also reproduced the same scenario with the graded-by-3 configurations and got 2 since the 1st three peers graded 1, 0, and 3 in that order. Please let me know if the submission order also affected your tests.

pomegranited: Oh wow, submission order does affect the final score! That's annoying.

> Could you confirm that the number of peer grades configured while testing was 5?

Yes, 5 peer grades were required, and only 5 were received.

I used the default graded-by-5 configuration, with
Ideas: poor=0, fair=3, good=5
Content: poor=0, fair=1, good=3, excellent=5

Audit user received the same "Content" peer assessments as Honor, but in a different order, and so they got different scores. (Apologies about the "Ideas" assessments, I thought I did them all the same too, but apparently not.)

Audit received:
image

Honor received:
image

pomegranited:
@mariajgrimaldi I've re-tested with your changes and, with the exception of the [ordering issue discussed above](https://github.com/openedx/edx-ora2/pull/2196#discussion_r1594831503), I found no other issues.

Redwood gets cut soon, so I'm happy to merge this now, and then you can do a follow-up PR to fix the ordering bug? We'll also need a version bump to get this into edx-platform:

> Bumped version number in [openassessment/__init__.py](https://github.com/openedx/edx-ora2/blob/master/openassessment/__init__.py#L4) and [package.json](https://github.com/openedx/edx-ora2/blob/master/package.json#L3)

mariajgrimaldi:
@pomegranited: thank you again for such a detailed review! We appreciate it. I just bumped version number in [openassessment/init.py](https://github.com/openedx/edx-ora2/blob/master/openassessment/__init__.py#L4) and [package.json](https://github.com/openedx/edx-ora2/blob/master/package.json#L3), and I'll be researching further the issue with the grading ordering in the next few days. I'll let you know what I find!

openedx-webhooks:
@mariajgrimaldi 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/edx-ora2&entry.867055334=https://github.com/openedx/edx-ora2/pull/2196&entry.1484655318=https://github.com/mariajgrimaldi&entry.752974735=2024-03-21+16:46&entry.1917517419=2024-05-09+01:09&entry.2133058324=Yes) so we can improve your experience in the future.

pomegranited:
Awesome, thank you for your rapid reply @mariajgrimaldi ! [ora2==6.10.0](https://pypi.org/project/ora2/6.10.0/) is released and ready for you to do a version bump PR in edx-platform if you want.

efortish:
Hello @pomegranited , hope you are doing well !
I was researching about that bug you reported about the mean calculation, I tried to reproduce a lot of scenarios to find out what was happening.
I'd like to ask you if you set the `graded by 5` before start to record the peer assessments or if you did it after that,
because I tried both scenarios and the second one was the only that showed me the mean=3, and it this because if you want to change the `graded by ` the change will only affect those new users who are going to be graded, the users who were already `graded by 3` will remain with that rubric and it wont change.
Additionally I found out why you are receiving the mean=3 in your scenario 
image
And it is because for some reason if you settled the `graded by 3 ` the UI doesn't display the PEERS in the right order, I tried a few test cases and I realize that if it is `graded by 3` and you received 5 peer assessments  the 1st will be PEER 5, the 2nd will be PEER 4, the 3rd will be PEER 3, etc. but when the mean is calculated the took only the 1st 3 peer assessments (peer 5,4,3) that's why you are receiving mean=3 (It is taking excellent fair and good in that order), to prove what i'm saying I added some prints in my test cases with the same rubric as you, and this was the result:
image
So, even if you received 5 peer assessments the mean will be only calculated with the 1st 3 assesstments.
I know this can be confusing, but I tried every single scenario based on your bug report and I didn't found any bug related with the mean function and the print logs were clear enough.

Please let me know if you have any questions about this and I'll be glad to answer asap, for now I'd like to say that there is no bug related to the mean function

pomegranited:
Hi @efortish , thank you for looking into this issue!

> I'd like to ask you if you set the graded by 5 before start to record the peer assessments or if you did it after that,

I'm afraid it's been too long since I did this testing, and I can't remember exactly what I did. But your explanation sounds very plausible. Ok to declare this not an issue :)

PR #2206, opened 2024-04-16
build: Update codecov and use the new token.

feanil:
Newer versions of codecov require a repo specific token to publish
coverage reports.  The older versions don't but are failing more often
as they prepare to drop support for them. As a part of this change a
the token was added as a repository actions secret.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,8 @@ jobs:
 
     - name: Run Coverage
       if: matrix.python-version == '3.8' && matrix.toxenv=='django42'
-      uses: codecov/codecov-action@v1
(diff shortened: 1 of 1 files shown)


mariajgrimaldi:
@pomegranited: this solves the error we mentioned here https://github.com/openedx/edx-ora2/pull/2203#issuecomment-2059177087

pomegranited (approved):
:+1: Thank you for fixing this @feanil !

- [x] I didn't test this, CI did.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] <strike>Includes documentation</strike> N/A
- [x] <strike>User-facing strings are extracted for translation</strike> N/A

PR #2213, opened 2024-05-23
fix: use get to avoid key error

jansenk:
** If the environment doesn't have ENABLE_ORA_PEER_CONFIGURABLE_GRADING defined this line will throw a keyerror

- Note: I don't know why  it isn't defined? It looks like it should be? Nevertheless the error is thrown. Investigation for another time.

JIRA: [CR-6800](https://2u-internal.atlassian.net/browse/CR-6800)

FYI: @openedx/content-aurora

### DIFF
--- a/openassessment/__init__.py
+++ b/openassessment/__init__.py
@@ -2,4 +2,4 @@
 Initialization Information for Open Assessment Module
 """
 
-__version__ = '6.11.0'
+__version__ = '6.11.1'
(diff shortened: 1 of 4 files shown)


jansenk:
The only failure is codecov. I'm trying to fix an active issue. Not gonna bother with codecov right now

BrandonHBodine (approved):
lgtm

PR #2236, opened 2024-09-25
build: Manually pull some RTD Context.

feanil:
See https://about.readthedocs.com/blog/2024/07/addons-by-default/ for details but
essentially RTD is changing how it's building docs and this will let us handle the
change gracefully.


pomegranited (approved):
:+1: Thanks for this @feanil !

- [x] I couldn't test this directly (requires deployment to RTD), but I checked that the docs still build ok locally with this change.
- [x] I read through the code and the notes about the update on the RTD site.
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] <strike>Includes documentation</strike> N/A
- [x] <strike>User-facing strings are extracted for translation</strike> N/A

PR #2386, opened 2026-04-08
build: Remove outdated sphinx-contentui package

sarina:
See openedx/docs.openedx.org#1417

This is an outdated package that was included in the docs cookie cutter. It can cause build errors and doesn't seem to be used.


sarina:
@UsamaSadiq tagging you here because the PR you merged yesterday appears to have broken tests on master

sarina:
Rebased on master to pick up https://github.com/openedx/edx-ora2/pull/2391 and tests are passing now, thanks for the quick response!

feanil (approved):

PR #2389, opened 2026-04-09
Ahtisham/add ora reminder notification

AhtishamShahid:
This pull request introduces the ORA Reminder feature, which automatically sends periodic notifications to learners who have submitted an Open Response Assessment (ORA) but have not yet completed required peer or self review steps. The implementation involves a new database model, a Celery sweeper task, and integration with the submission flow to create and manage reminders. Documentation and configuration options are also provided.

Key changes include:

**Feature Implementation: ORA Reminders**
* Added a new `ORAReminder` model to persist reminder schedules for learners, including fields for user, course, submission, deadlines, and reminder state. This enables efficient tracking and querying of pending reminders without frequent modulestore lookups. 
* Integrated reminder creation into the submission process by adding `_handle_post_submission_notifications` in `submissions_actions.py`, which creates a reminder entry and ensures the sweeper task is running after a qualifying submission.
* Registered and imported the `sweep_ora_reminders` Celery task so it is autodiscovered and can process due reminders on a platform-wide schedule.

**Documentation and Configuration**
* Added comprehensive documentation in `docs/ora_reminders.rst`, describing how the reminder system works, configuration options, and operational details for durability and error handling.

**Supporting Changes**
* Bumped the package version to `7.1.0` to reflect the addition of this new feature.


docs/ora_reminders.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Set ``ENABLE_ORA_REMINDERS`` to ``True`` in your Django settings:
+
+.. code-block:: python
+
+    ENABLE_ORA_REMINDERS = True
+
+The notification type ``ora_reminder`` must also be registered in edx-platform's
+``openedx.core.djangoapps.notifications.base_notification`` (already included
+in edx-platform ≥ the version that ships this feature).
```
feanil, line 22: Is there already a corresponding PR for this or will you be making one, once this is released?

docs/ora_reminders.rst
```diff
@@ -0,0 +1,120 @@
...
+     - Maximum number of reminders sent per learner per ORA submission.
+       Once this limit is reached the reminder row is deactivated.
+   * - ``ORA_REMINDER_SWEEP_INTERVAL_SECONDS``
+     - ``1800``
+     - How often (in seconds) the sweeper re-schedules itself. Each run
+       processes all rows whose ``next_reminder_at`` has passed.
+   * - ``ORA_REMINDER_SWEEP_BATCH_SIZE``
+     - ``1000``
+     - Maximum rows processed per sweep cycle. If more rows are due they
+       will be picked up on the next sweep.
```
feanil, line 55: Does this mean that if there are more than 48,000 reminders that need to be sent, that the ones above 48,000 won't be sent in time?

openassessment/workflow/models.py
```diff
@@ -1091,3 +1091,56 @@ def get_latest_workflow_cancellation(cls, submission_uuid):
...
+    to hit the modulestore or CourseOverview.
+
+    Termination uses a time-based condition (spec §3): the sweeper stops sending
+    when ``now >= submission_time + INITIAL_DELAY_HOURS + MAX_COUNT * INTERVAL_HOURS``.
+
+    .. no_pii:
+    """
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, related_name="ora_reminders", on_delete=models.CASCADE)
+    course_id = models.CharField(max_length=255, db_index=True)
+    ora_usage_key = models.CharField(max_length=255)
```
feanil, line 1116: Use a UsageKey model field instead: Use a usage key model field instead: https://github.com/openedx/opaque-keys/blob/master/opaque_keys/edx/django/models.py#L263

openassessment/workflow/models.py
```diff
@@ -1091,3 +1091,56 @@ def get_latest_workflow_cancellation(cls, submission_uuid):
...
+    # Cached deadline fields — populated at submission time to avoid
+    # expensive modulestore / CourseOverview lookups during sweep.
+    ora_due_date = models.DateTimeField(null=True, blank=True)
+    course_end_date = models.DateTimeField(null=True, blank=True)
+    peer_assessment_due = models.DateTimeField(null=True, blank=True)
+    self_assessment_due = models.DateTimeField(null=True, blank=True)
+
+    # Reminder state — next_reminder_at is NULL when the row is inactive.
+    next_reminder_at = models.DateTimeField(db_index=True, null=True, blank=True)
+    is_active = models.BooleanField(default=True, db_index=True)
```
feanil, line 1131: When you instantiate this object, it will be null by default, and it will have `is_active` set to `True` those seem like two contradictory settings.  Does it make sense to set is_active to `False` by default? or set `auto_now_add` to `True` on the reminder date?

openassessment/workflow/tasks.py
```diff
@@ -1,11 +1,32 @@
...
+
+    Clears the sweep lock first because pending countdown tasks from a
+    previous worker process are lost on restart.  In a multi-worker setup
+    ``cache.add`` inside ``ensure_sweep_chain_running`` still prevents
+    duplicate chains — only the first worker to re-acquire the lock wins.
+    """
+    from django.core.cache import cache
+    from openassessment.xblock.utils.ora_reminders import ensure_sweep_chain_running, SWEEP_LOCK_KEY
+    cache.delete(SWEEP_LOCK_KEY)
+    ensure_sweep_chain_running()
```
feanil, line 28: Won't this break if more than one worker is started?  If Worker A is started up and starts running the sweep.  Then later an operator needs more workers and starts up worker B, won't worker B clobber the cache key and start running the sweep?

openassessment/xblock/apis/submissions/submissions_actions.py
```diff
@@ -35,6 +39,127 @@
...
+
+        # Construct the content URL to the ORA block
+        if block_config_data.course:
+            course_id_str = str(block_config_data.course.id)
+        else:
+            course_id_str = student_item_dict.get("course_id")
+
+        item_id = student_item_dict.get("item_id")
+        lms_root_url = getattr(settings, 'LMS_ROOT_URL', '')
+        content_url = f"{lms_root_url}/courses/{course_id_str}/jump_to/{item_id}"
```
feanil, line 124: Why use an LMS root URL at all, why not use the relative URL instead?

openassessment/xblock/utils/ora_reminders.py
```diff
@@ -0,0 +1,426 @@
...
+        try:
+            _process_single_reminder(reminder, now)
+            processed += 1
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(
+                'ora_reminders: Error processing reminder id=%s (user=%s, ora=%s).',
+                reminder.id, reminder.user_id, reminder.ora_usage_key,
+            )
+
+    if processed:
```
feanil, line 252: Should the sweep lock be cleared at this point?

openassessment/xblock/utils/ora_reminders.py
```diff
@@ -0,0 +1,376 @@
...
+    Returns:
+        bool: True if peer submissions are available to review, False if waiting for peers
+    """
+    try:
+        peer_submission = peer_api.get_submission_to_assess(
+            submission_uuid,
+            graded_by=1,
+            peek=True
+        )
+        return peer_submission is not None
```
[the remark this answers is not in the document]
AhtishamShahid: fixed - now storing peer_must_be_graded_by on ORAReminder at submission time, read from the ORA's peer config in submissions_actions.py and passed through to get_submission_to_assess

docs/ora_reminders.rst
AhtishamShahid: Here is PR for edx-platfrom . https://github.com/openedx/openedx-platform/pull/38298

docs/ora_reminders.rst
AhtishamShahid: batch is ordered oldest-first so nothing gets permanently skipped. Also, added a note that ORA_REMINDER_SWEEP_BATCH_SIZE can be increased for high-volume deployments

openassessment/workflow/models.py
AhtishamShahid: Updated.

openassessment/workflow/models.py
AhtishamShahid: changed is_active default to False. create_ora_reminder always sets it explicitly to True so runtime behavior is unchanged, but the schema is consistent now

openassessment/workflow/tasks.py
AhtishamShahid: on worker restart the old chain's countdown tasks are lost but the lock might still be alive. deleting first + using cache.add (atomic) means only one worker wins the lock and starts a single chain. added a comment explaining this in the code

openassessment/xblock/apis/submissions/submissions_actions.py
AhtishamShahid: when user clicks on the notification user is redirected to that item, and the notification system is built in a way that would not be able to construct url dynamically

docs/ora_reminders.rst
```diff
@@ -0,0 +1,122 @@
+ORA Reminder Notifications
```
saraburns1, line 1: can you please put the documentation [here](https://github.com/openedx/docs.openedx.org/blob/main/source/site_ops/how-tos/enable_notifications.rst) or create an issue in that repo for someone else to do

AhtishamShahid: Hi , I have created PR to add docs . https://github.com/openedx/docs.openedx.org/pull/1446/changes

AhtishamShahid: Docs PR is Approved & merged.

openassessment/workflow/models.py
```diff
@@ -1091,3 +1093,73 @@ def get_latest_workflow_cancellation(cls, submission_uuid):
...
+    All deadline fields are cached at creation time so the sweeper never needs
+    to hit the modulestore or CourseOverview.
+
+    Termination uses a count-based condition: the sweeper stops sending when
+    ``reminder_sent_count`` reaches ``ORA_REMINDER_MAX_COUNT``.
+
+    .. no_pii:
+    """
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, related_name="ora_reminders", on_delete=models.CASCADE)
+    course_id = models.CharField(max_length=255, db_index=True)
```
feanil: This should also be a usage key.

openassessment/xblock/utils/ora_reminders.py
```diff
@@ -0,0 +1,517 @@
...
+        initial_delay = getattr(settings, 'ORA_REMINDER_INITIAL_DELAY_HOURS', 0)
+        ORAReminder.objects.filter(submission_uuid=submission_uuid).update(
+            is_active=True,
+            last_known_step=new_step,
+            step_start_time=now_utc,
+            next_reminder_at=now_utc + timedelta(hours=initial_delay),
+            reminder_sent_count=0,
+        )
+        ensure_sweep_chain_running()
+        logger.info(
```
feanil, line 249: `ensure_sweep_chain_running()` here runs synchronously inside the AssessmentWorkflow save's transaction. The call path on a typical peer→self (or self→peer) transition is:

`student_view` → `update_workflow_status` → `update_from_assessments` → `self.save()` (workflow/models.py:465) → this signal handler → `reset_ora_reminder_for_step` → `ensure_sweep_chain_running`.

`ensure_sweep_chain_running` does `cache.get` × 2, `cache.delete`, `cache.add`, and `sweep_ora_reminders.apply_async()` (broker enqueue, usually Redis). All of that runs inside the request-handling transaction before `student_view` can return HTML. One concerns:

* **Error propagation:** `reset_ora_reminder_for_step` only catches `ORAReminder.DoesNotExist`. A `redis.ConnectionError` (or any other cache/broker exception) would bubble out through `update_from_assessments`. The xblock catches `AssessmentWorkflowError` at `openassessmentblock.py:572` but not generic `Exception` — so a broker hiccup during a status transition could turn into a 500 on ORA page load.

Suggest deferring with `transaction.on_commit` so the cache/broker work fires after the workflow transaction commits, off the critical path:

```python
from django.db import transaction
...
transaction.on_commit(ensure_sweep_chain_running)
```

The `ORAReminder.objects.filter(...).update(...)` right above this should stay in the transaction (it correctly commits/rolls back with the workflow change) — only the cache/broker side-effects need to defer.

openassessment/xblock/apis/submissions/submissions_actions.py
```diff
@@ -179,6 +311,14 @@ def create_submission(
             group_by_id
         )
 
+    # Handle post-submission notifications and reminders
+    _handle_post_submission_notifications(
```
feanil, line 328: Is there a reason this can't be async? Why does it need to be run synchronously in the submit transaction?

openassessment/workflow/tasks.py
```diff
@@ -1,11 +1,37 @@
...
+    """
+    from django.core.cache import cache
+    from openassessment.xblock.utils.ora_reminders import ensure_sweep_chain_running, SWEEP_LOCK_KEY
+    # Delete the lock before attempting to acquire it.  On worker restart the
+    # previous chain's countdown tasks are lost, so the lock may still be set
+    # even though no chain is running.  Clearing it lets ensure_sweep_chain_running
+    # start a fresh chain via cache.add (which is atomic, so in a multi-worker
+    # restart only one worker wins the lock and starts a single chain).
+    cache.delete(SWEEP_LOCK_KEY)
+    ensure_sweep_chain_running()
```
feanil, line 35: Following up on the earlier thread. `cache.add` atomicity does make *simultaneous* startup safe, but the unconditional `cache.delete` here makes rolling deploys worse, not better:

- LMS in production runs many celery workers (default queue, high, high_mem). The `sweep_ora_reminders` task isn't routed to a specific queue, so every default-queue worker can run it.
- During a rolling deploy workers restart one at a time over several minutes. Each restart hits this line and yanks the lock out from under whatever chain is currently in flight.
- The just-restarted worker then re-acquires via `cache.add` and starts a parallel chain. The old chain's pending countdown task is still queued in the broker and will run too. Both chains sweep, both read the same due rows, both send notifications, both advance `next_reminder_at`.
- Chains converge within ~1 sweep cycle (whoever wins `cache.add` at end-of-sweep keeps going), but for that window duplicates are real.

`ensure_sweep_chain_running` already has the heartbeat-staleness check (lines 169-182 of `ora_reminders.py`) which handles the legit "previous chain died" case. The explicit `cache.delete` here is both redundant *and* harmful in the multi-worker case. Suggest removing it entirely and trusting the heartbeat check.

openassessment/xblock/utils/ora_reminders.py
```diff
@@ -0,0 +1,517 @@
...
+    """
+    sweep_interval = getattr(settings, 'ORA_REMINDER_SWEEP_INTERVAL_SECONDS', 1800)
+    lock_timeout = sweep_interval * SWEEP_LOCK_TIMEOUT_MULTIPLIER
+    # Never self-chain when tasks run eagerly (CELERY_ALWAYS_EAGER=True) — it causes infinite recursion.
+    should_rechain = not getattr(settings, 'CELERY_ALWAYS_EAGER', False)
+
+    try:
+        if not getattr(settings, 'ENABLE_ORA_REMINDERS', False):
+            logger.info('ora_reminders: ENABLE_ORA_REMINDERS is disabled. Sweeper stopping.')
+            should_rechain = False
```
feanil, line 309: When the feature is disabled here, the lock isn't cleared (the `finally` block skips `cache.delete` because `should_rechain=False`), so it persists for the full 90-min TTL. Disable-then-re-enable within that window means no new chain starts.

The existing heartbeat-staleness check at `ensure_sweep_chain_running` (lines 169-182) does eventually recover: when the feature is off the heartbeat stops being written (line 276 is after this return), so after `2 × sweep_interval` (~60 min) the next `ensure_sweep_chain_running` call clears the stale lock. But that's a long wait.

Suggest invalidating the heartbeat here so recovery is immediate on re-enable, without clearing the lock (which would cause mixed-state churn across enabled/disabled workers):

```python
should_rechain = False
cache.delete(SWEEP_HEARTBEAT_KEY)
return
```

Next `ensure_sweep_chain_running` call sees the missing heartbeat → clears the lock → starts a fresh chain.

openassessment/xblock/utils/ora_reminders.py
```diff
@@ -0,0 +1,517 @@
...
+    Detect workflow step transitions and reset the ORA reminder window.
+
+    Fires on every AssessmentWorkflow save.  The status guard and the
+    last_known_step comparison inside reset_ora_reminder_for_step make this
+    a no-op for saves that don't represent a transition to a new peer/self step.
+    """
+    if instance.status not in PENDING_REMINDER_STEPS:
+        return
+    reset_ora_reminder_for_step(instance.submission_uuid, instance.status)
+
```
feanil, line 283: Potential bug — the receiver and the sweeper use *different* sources of truth for "what step the learner is on":

- This receiver uses `instance.status`.
- `_get_workflow_step` was just rewritten (commit `7083a1dce`) to *not* trust `workflow.status` because `peer.can_be_skipped()` flips the cached status to `'self'` prematurely. The sweeper instead queries `AssessmentWorkflowStep` directly for the first incomplete step by `order_num`.

Scenario:
1. User submits, status=`peer`, signal fires, reminder created with `last_known_step='peer'`.
2. Sweeper runs, `_get_workflow_step` returns `'peer'` (real pending), sends a peer reminder, increments count, sets `last_known_step='peer'`.
3. User loads ORA page → `update_from_assessments` → `workflow.save()` with `status='self'` (peer skipped via `can_be_skipped`). Signal fires here with `instance.status='self'`.
4. `reset_ora_reminder_for_step` compares `last_known_step='peer'` ≠ `new_step='self'` → resets `reminder_sent_count=0`, sets `last_known_step='self'`, reactivates.
5. Sweeper runs again, `_get_workflow_step` still returns `'peer'` (because `submitter_completed_at` is `None`), sends another peer reminder, resets `last_known_step='peer'`.
6. Workflow saves again with `status='self'` (still skipped). Signal fires. `'peer'` ≠ `'self'` → reset count to 0 again. Loop.

The reminder budget never accumulates → user can receive an unbounded number of peer reminders. The receiver should use the same step-resolution as the sweeper (call `_get_workflow_step(instance.submission_uuid)` instead of using `instance.status`), or store an explicit "real pending step" rather than mixing two different definitions.

openassessment/xblock/test/test_ora_reminders.py
```diff
@@ -0,0 +1,1388 @@
...
+# ---------------------------------------------------------------------------
+# reset_ora_reminder_for_step
+# ---------------------------------------------------------------------------
+@override_settings(
+    ENABLE_ORA_REMINDERS=True,
+    ORA_REMINDER_INITIAL_DELAY_HOURS=INITIAL_DELAY,
+)
+class TestResetOraReminderForStep(CacheResetTest):
+    """Tests for reset_ora_reminder_for_step."""
+
```
feanil, line 1395: Suggestion: add a test for the `on_workflow_status_change` receiver itself, not just for `reset_ora_reminder_for_step`. The existing tests cover the helper in isolation, but the signal wiring (connecting receiver to `AssessmentWorkflow.post_save`) is untested. A simple test that saves an `AssessmentWorkflow` with `status='peer'` and asserts the reminder was reset would catch regressions like an accidental `@receiver` decorator removal.

openassessment/workflow/models.py
```diff
@@ -1091,3 +1093,73 @@ def get_latest_workflow_cancellation(cls, submission_uuid):
...
+    # Default False so that an accidentally-created row without a next_reminder_at
+    # is not treated as active by the sweeper.  create_ora_reminder always sets
+    # this to True explicitly.
+    is_active = models.BooleanField(default=False, db_index=True)
+
+    # Step tracking — used to detect when the workflow transitions to a new step
+    # (e.g. self → peer) so the reminder window can be reset per-step.
+    last_known_step = models.CharField(max_length=32, null=True, blank=True)
+    step_start_time = models.DateTimeField(null=True, blank=True)
+
```
feanil, line 1141: After the count-based termination refactor (commit `7083a1dce`), `step_start_time` is written in two places (`create_ora_reminder` and `reset_ora_reminder_for_step`) but never read in any guard — the previous time-window guard that anchored on it is gone. Either remove the field (with a migration) or document what it's still for (ops diagnostics?). Carrying schema you don't use is a small but real cost.

openassessment/xblock/apis/submissions/submissions_actions.py
```diff
@@ -35,6 +39,134 @@
...
+        _, _, _, due = block.is_closed(step=step_name)
+        if due is not None and due.tzinfo is None:
+            due = due.replace(tzinfo=timezone.utc)
+        return due
+    except Exception:  # pylint: disable=broad-except
+        return None
+
+
+def _handle_post_submission_notifications(submission, student_item_dict, block_config_data, block_workflow_data):
+    """
```
feanil, line 62: This function is the only entry point that creates `ORAReminder` rows, and `create_team_submission` (in this same file) doesn't call it. So team-based ORAs never get reminder rows. If team ORAs can have peer/self steps (the `ENABLE_ORA_TEAM_SUBMISSIONS` flag exists), team learners silently miss out on reminders.

This may be intentional, but right now nothing in the code or docs says so. Either:
- Add an early-return + comment here that says "team submissions intentionally excluded because <reason>", and document the limitation in `docs/ora_reminders.rst`, or
- Wire the team flow through the same handler.

docs/ora_reminders.rst
```diff
@@ -0,0 +1,122 @@
...
+   ``next_reminder_at = submission_time + ORA_REMINDER_INITIAL_DELAY_HOURS``.
+
+2. The ``sweep_ora_reminders`` Celery task runs every
+   ``ORA_REMINDER_SWEEP_INTERVAL_SECONDS`` seconds (self-chaining pattern).
+   On each run it queries all active rows whose ``next_reminder_at ≤ now``.
+
+3. For each due row, guards are checked in order:
+
+   - **Time window elapsed** — hours since submission exceed
+     ``INITIAL_DELAY + MAX_COUNT × INTERVAL`` → row deactivated.
```
feanil: Out of date relative to the code. Commit `7083a1dce` replaced the time-window guard with a count-based termination (`reminder_sent_count >= ORA_REMINDER_MAX_COUNT`). This bullet should now say something like:

> **Max reminders sent** — `reminder_sent_count >= ORA_REMINDER_MAX_COUNT` → row deactivated. Only actual notification sends increment the counter; peer-unavailable deferrals do not consume the budget.

Same applies to other "time window" / "without tracking a count" references in this file — make sure the prose matches the new behavior throughout.

settings/base.py
```diff
@@ -175,7 +175,38 @@
...
+# Maximum number of reminders sent per learner per ORA submission.
+ORA_REMINDER_MAX_COUNT = 3
+
+# Hours between consecutive reminder notifications.
+ORA_REMINDER_INTERVAL_HOURS = 48
+
+# Hours after submission before the first reminder is sent.
+# Allows learners time to complete reviews before being nudged.
+ORA_REMINDER_INITIAL_DELAY_HOURS = 0
+
```
feanil, line 200: Heads up: this default changed from `24` to `0` in commit `7083a1dce`. With `0`, a learner gets a "reminder" notification immediately after submission — which isn't really a reminder, it's an acknowledgment. Is that intentional? If so, the docs probably want to call it out; if not, restoring the `24`-hour default seems safer.

openedx-webhooks:
Thanks for the pull request, @AhtishamShahid!


This repository is currently maintained by `@openedx/openedx-unmaintained`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

openassessment/workflow/models.py
AhtishamShahid: Done — `course_id` is now a `CourseKeyField`.

openassessment/xblock/utils/ora_reminders.py
AhtishamShahid: Done — `ensure_sweep_chain_running` is now deferred via `transaction.on_commit` inside `reset_ora_reminder_for_step`, so the cache/broker work runs after the workflow transaction commits and a broker hiccup can no longer surface as a 500 on ORA page load. The `ORAReminder.objects.filter(...).update(...)` stays inside the transaction, as you suggested.

openassessment/xblock/apis/submissions/submissions_actions.py
AhtishamShahid: The reminder DB write and chain start are now deferred via `transaction.on_commit`, so they run after the submission commits (off the critical path, and skipped entirely if the submission rolls back). The context-gathering above has to stay in-request because it needs the xblock (deadlines, content URL, peer config).

openassessment/workflow/tasks.py
AhtishamShahid: Done — removed the unconditional `cache.delete` in `on_worker_ready`; it now just calls `ensure_sweep_chain_running`, which clears the lock only when the heartbeat is stale/missing. Rolling restarts no longer spawn parallel chains. Verified with a multi-worker chaos test (8 rounds of restart / stop-start / full recreate → each learner received exactly `MAX_COUNT` notifications, zero duplicates).

openassessment/xblock/utils/ora_reminders.py
AhtishamShahid: Done — the disabled branch now deletes the heartbeat, and `ensure_sweep_chain_running` treats a missing heartbeat as a dead chain, so re-enabling restarts the sweeper on the next call instead of waiting out the lock TTL.

openassessment/xblock/utils/ora_reminders.py
AhtishamShahid: Fixed — `on_workflow_status_change` now resolves the pending step via `_get_workflow_step` (the same source the sweeper uses) instead of `instance.status`, so peer's `can_be_skipped()` status flip no longer resets the count on every page load. Added receiver-wiring + transition tests and verified the peer↔self transitions end-to-end.

openassessment/xblock/test/test_ora_reminders.py
AhtishamShahid: Done — added `TestOnWorkflowStatusChange`, which saves an `AssessmentWorkflow` and asserts the reminder is reset through the signal wiring (plus the status-vs-real-step case and a no-op same-step save).

openassessment/xblock/apis/submissions/submissions_actions.py
AhtishamShahid: Documented — added a comment in `create_team_submission` plus a "Limitations" section in `docs/ora_reminders.rst` explaining that team submissions are intentionally excluded (the row is keyed on an individual `submission_uuid`).

docs/ora_reminders.rst
AhtishamShahid: Done — rewrote the guards section for count-based termination (`reminder_sent_count >= ORA_REMINDER_MAX_COUNT`) and updated the durability/heartbeat prose throughout.

settings/base.py
AhtishamShahid: Intentional — `0` is a product decision: the first notification fires immediately after submission as an acknowledgement, and subsequent nudges follow `ORA_REMINDER_INTERVAL_HOURS`. I've added a comment marking it intentional.

openassessment/workflow/models.py
AhtishamShahid: Done — removed `step_start_time`. Since the reminder migrations are unreleased, the change is folded into `0007_orareminder` rather than a new alter-migration.

docs/ora_reminders.rst
AhtishamShahid: Done — rewrote the guards section for count-based termination (`reminder_sent_count >= ORA_REMINDER_MAX_COUNT`) and updated the durability/heartbeat prose throughout.

settings/base.py
AhtishamShahid: Intentional — `0` is a product decision: the first notification fires immediately after submission as an acknowledgement, and subsequent nudges follow `ORA_REMINDER_INTERVAL_HOURS`. I've added a comment marking it intentional.

feanil (approved):
Thanks for addressing all the issues and talking through this with me!

